PR #10, opened 2018-02-07
updated context target 

bw2:
This is to work around this kubectl error:
 'The connection to the server localhost:8080 was refused - did you specify the right host or port?'

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,4 @@
-.DS_Store
+'.DS_Store
 npm-debug.log
 __pycache__/
 *.pyc
@@ -8,6 +8,7 @@ __pycache__/
 env
 build
 dist
+.idea
 node_modules
 resources/*
 *lib
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,4 @@
-'.DS_Store
+.DS_Store
 npm-debug.log
 __pycache__/
 *.pyc
--- a/projects/gnomad/deploy/Makefile
+++ b/projects/gnomad/deploy/Makefile
@@ -23,12 +23,12 @@ SERVICE_FORWARDING_RULE_ID=$(shell gcloud compute forwarding-rules list --format
 start-gnomad: build-javascript build-legacy 	create-gnomad-legacy gnomad-legacy-service
 
 context:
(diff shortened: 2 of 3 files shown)


projects/gnomad/deploy/Makefile
```diff
@@ -23,12 +23,12 @@ SERVICE_FORWARDING_RULE_ID=$(shell gcloud compute forwarding-rules list --format
...
-	gcloud container clusters get-credentials $(CLUSTER_NAME) --zone=$(GCLOUD_ZONE)
+	gcloud container clusters get-credentials $(CLUSTER_NAME) --zone=$(GCLOUD_ZONE) --project=$(GCLOUD_PROJECT)
 	kubectl config set-context $(CLUSTER_NAME) \
-	--cluster $(GKE_CLUSTER_NAME) \
-	--user $(GKE_CLUSTER_NAME) \
+	--cluster $(CLUSTER_NAME) \
+	--user $(CLUSTER_NAME) \
 	--namespace $(CLUSTER_NAMESPACE)
-	kubectl config use-context $(CLUSTER_NAME)
+	kubectl config use-context gke_$(GCLOUD_PROJECT)_$(GCLOUD_ZONE)_$(CLUSTER_NAME)
```
nawatts: Was it important that the arguments to `set-context` and `use-context` matched?

Previously, this ran
```
set-context $(CLUSTER_NAME)
use-context $(CLUSTER_NAME)
```

Now it runs
```
set-context $(CLUSTER_NAME)
use-context gke_$(GCLOUD_PROJECT)_$(GCLOUD_ZONE)_$(CLUSTER_NAME)
```

nawatts:
There's a similar `context` block in cluster/Makefile... does it need to be changed as well?

https://github.com/macarthur-lab/gnomadjs/blob/2169e0b5af52403fa77aa05dbd0266626090078c/cluster/Makefile#L31-L37

.gitignore
```diff
@@ -1,4 +1,4 @@
-.DS_Store
+'.DS_Store
```
nawatts: I think this should stay `.DS_Store`

bw2: oops, good point

PR #18, opened 2018-04-27
Specify mapping for regional missense constraint data

nawatts:
Fixes https://github.com/macarthur-lab/gnomad_browser/issues/108

Currently, when uploading regional missense constraint data, ES automatically sets the mapping for `regional_missense/region`. Since the records in the data file are ordered by chromosome (1-22, X), ES maps the `chr` field to a numeric type (`long`).

However, not all records have a numeric value for `chr`. Some genes (such as CDKL5 which was reported in the issue) have a value of 'X' for the `chr` field. With `chr` mapped to a `long` type, records for chromosome 'X' do not appear in query results.

This change explicitly sets the mapping for `regional_missense/region` instead of allowing ES to auto-detect it. With `chr` set to a text field, query results include records for chromosome 'X'.

### DIFF
--- a/packages/api/src/schema/types/regionalConstraint.js
+++ b/packages/api/src/schema/types/regionalConstraint.js
@@ -13,7 +13,7 @@ export const regionalConstraintRegion = new GraphQLObjectType({
   fields: () => ({
     transcript: { type: GraphQLString },
     gene: { type: GraphQLString },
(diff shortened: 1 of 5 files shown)


nawatts:
This will require reloading the regional missense constraint data:
```
cd projects/gnomad/data
make regional-constraint-regional-missense
```

PR #24, opened 2018-05-07
Update styles of regional constraint track

nawatts:
* Color regions based on their significance and observed/expected missense.
* Show observed/expected missense in region.
* Add help button.
* When hovering the mouse over a region, show a tooltip with observed/expected missense and significance. 
* For consistency with other Track components, move RegionalConstraintTrack into its own package instead of defining it inline in projects/gnomad/RegionViewer.

## Before
screen shot 2018-06-20 at 10 12 52 am

## After
screen shot 2018-06-20 at 10 13 34 am
screen shot 2018-06-20 at 10 12 24 am

### DIFF
--- a/packages/api/package.json
+++ b/packages/api/package.json
@@ -25,6 +25,7 @@
   "license": "ISC",
   "dependencies": {
     "compression": "^1.7.1",
+    "core-js": "^2.5.0",
     "cors": "^2.8.1",
     "elasticsearch": "^13.3.1",
     "express": "^4.14.0",
--- a/packages/track-regional-constraint/src/RegionalConstraintTrack.js
+++ b/packages/track-regional-constraint/src/RegionalConstraintTrack.js
@@ -1,0 +1,82 @@
+import PropTypes from 'prop-types'
+import React from 'react'
+import styled from 'styled-components'
+
+
+const Container = styled.div`
+  display: flex;
+  flex-direction: row;
+  margin-top: 5px;
+`
+
+
+const LeftPanel = styled.div`
+  align-items: center;
+  display: flex;
+  width: ${props => props.width}px;
+`
+
+
+const CenterPanel = styled.div`
+  display: flex;
+  align-items: center;
+  width: ${props => props.width}px;
+`
+
+
+export default function RegionalConstraintTrack({
+  height,
+  leftPanelWidth,
+  positionOffset,
+  regionalConstraintData,
+  strand,
+  width,
+  xScale,
+}) {
+  return (
+    <Container>
+      <LeftPanel width={leftPanelWidth}>
+        <p>Regional constraint</p>
+      </LeftPanel>
+      <CenterPanel width={width}>
+        <svg height={height} width={width}>
+          {regionalConstraintData.map((region) => {
+            const regionStart = strand === '+'
+              ? region.genomic_start
+              : region.genomic_end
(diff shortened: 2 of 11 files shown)


anneodonnell:
I think this looks pretty nice. We probably need a brief description of it in the FAQ as I'm sure we will get questions about what the numbers mean. 

I wonder if the regional constraint table should have some kind of O/E mention also to give people a clue about these numbers.

nawatts:
We have some documentation for regional constraint here: https://github.com/macarthur-lab/gnomad-docs/blob/master/docs/regional-constraint.md

We can add a help icon/popup to the track to show this.

ksamocha:
To @anneodonnell's point, you might want to write "O/E" instead of gamma. We only used gamma in the paper for ease.

ksamocha:
I'll note here the point I brought up at the beginning of the call today: there are two separate pieces of information for all gene/region -- obs/exp and signficance (chisq_diff_null). Currently, gnomAD is coloring by obs/exp to highlight extremely depleted regions. DECIPHER decided to only color regions by their obs/exp value, but only if those regions were significantly different from expectation.

I don't think there is a correct way to do this. There are 143 regions that have obs/exp < 0.6 and a chi-squared value that we wouldn't consider significant (< 10.8).
```
> awk '{ if($9<0.6 && $10<10.8) print }' genes_regional_data_cleaned_chisq_0.0_metrics_bpadded2_2016_10_05.txt | wc -l
     143
```

Something to think about before fully approving this. Otherwise, I think it looks pretty good.

nawatts:
Changed gamma to O/E and added a help button.

screen shot 2018-06-07 at 11 26 06 am

nawatts:
Based on feedback from today's meeting, the remaining todos are:

* [ ] Change label to "Regional missense constraint" (this also affects help text)
* [ ] Remove O/E labels from regions and instead show them in a tooltip on hover
* [ ] Color regions based on significance instead of observed/expected variation

nawatts:
Updated the PR description with current screenshots.

packages/api/package.json
```diff
@@ -25,6 +25,7 @@
   "license": "ISC",
   "dependencies": {
     "compression": "^1.7.1",
+    "core-js": "^2.5.0",
```
nawatts: This was necessary to solve some version mismatch problem in `babel-polyfill` after updating dependencies.

packages/ui/src/tooltip/withTooltip.js
```diff
@@ -0,0 +1,94 @@
...
+        onMouseEnter: this.renderTooltipInPortal,
+        onMouseLeave: this.removeTooltip,
+        ref: this.referenceElementRef,
+      }
+    )
+  }
+}
+
+
+export function withTooltip(Component) {
```
mattsolo1: Love this!

packages/track-regional-constraint/src/RegionalConstraintTrack.js
```diff
@@ -0,0 +1,149 @@
...
+    return '#ffeda0'
+  }
+  if (region.obs_exp > 0.2) {
+    return '#feb24c'
+  }
+  return '#f03b20'
+}
+
+
+const WithRegionTooltip = withTooltip(({ region }) => (
```
mattsolo1: This tooltip component is beautiful

PR #66, opened 2018-05-17
Fix keyboard shortcuts for search

### LINKED ISSUE #65: Keyboard shortcut for variant search does not work
The variant search input is intended to take keyboard focus when either "Cmd+F" or "Meta+S" is pressed. However, pressing either of those key combinations throws an error.

https://github.com/macarthur-lab/gnomadjs/blob/e1e2bf9090f4756ed34f702bb0b7485f0ed99295/projects/gnomad/src/Settings.js#L34-L39

### What you did:

* Go to gene page
* Press "Cmd+F"

### What happened:

"Uncaught TypeError: Cannot read property 'focus' of undefined" is thrown.

This also prevents the browser's built in search from working on other pages such as Terms or FAQ.

nawatts:
Resolves #65.

* Moves handling of global keyboard shortcuts for searching into the search component.
* Remove `mousetrap` dependency from other packages/projects.
* Change imports from `@broad/ui/src/search/simpleSearch` to import from `@broad/ui`.

### DIFF
--- a/packages/ui/src/search/simpleSearch.js
+++ b/packages/ui/src/search/simpleSearch.js
@@ -1,5 +1,6 @@
-import React from 'react'
+import Mousetrap from 'mousetrap'
 import PropTypes from 'prop-types'
+import React, { Component } from 'react'
 import styled from 'styled-components'
 
 const SearchWrapper = styled.div`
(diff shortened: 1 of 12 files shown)


packages/ui/src/search/simpleSearch.js
```diff
@@ -9,69 +10,88 @@ const SearchWrapper = styled.div`
...
-          }}
+          onChange={this.onChange}
+          placeholder={this.props.placeholder}
+          innerRef={this.inputRef}
+          type="text"
+          value={this.state.value}
         />
-        {/* <datalist id={listName}>
-          {options.map(item => <option value={item} />)}
-        </datalist> */}
```
nawatts: Since the `datalist` element wasn't rendered, the `options` and `listName` props for this component were unused.

PR #70, opened 2018-06-07
Tweaks to SCHEMA browser's variant table

nawatts:
Based on feedback from @tarjindersingh

* Reorder columns
* ~Add column for N de novos~
* Remove hard coded case AC/AN values
* Handle null values for HGVSc/p in API


nawatts:
> Let's get rid of that ac denovo column for now

👍Dropped that commit.

PR #71, opened 2018-06-07
When sorting variants by a field, sort variants with a null value for that field to the end of the list

nawatts:
Currently, when sorting the variants table by a field in ascending order, variants with a null value (no data) for that field appear at the top of the list.

With this change, variants with a null value for the sort column are _always_ sorted to the bottom of the list regardless of the sort order.


nawatts:
> Do you mind also making this correction on the schizophrenia results page?

Done in #74.

PR #73, opened 2018-06-07
Correct dependencies of redux-variants package

nawatts:


nawatts:
It looks like it may have been copied from `help`'s package.json.

PR #74, opened 2018-06-07
Make gene results table sorting consistent with variant table

nawatts:
Follow up from https://github.com/macarthur-lab/gnomadjs/pull/71#pullrequestreview-126883192.

Currently, when sorting the gene results table by a field in ascending order, results with a null value (no data) for that field appear at the top of the list.

With this change, gene results with a null value for the sort column are always sorted to the bottom of the list regardless of the sort order.

Some of the sorting code is duplicated between here and the redux-variants package (see #71). It could possibly be extracted to the utilities package.

### DIFF
--- a/projects/schizophrenia/src/GeneResults.js
+++ b/projects/schizophrenia/src/GeneResults.js
@@ -19,6 +19,15 @@ import {
   Search,
 } from '@broad/ui'
 
+
+function isEmpty(val) {
+  return (
+    val === undefined
+    || val === null
+    || val === ''
+  )
+}
+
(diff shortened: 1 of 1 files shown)


projects/schizophrenia/src/GeneResults.js
```diff
@@ -83,7 +92,7 @@ class SchizophreniaGeneResults extends PureComponent {
 
   setSearchText = searchText => this.setState({
     searchText: String(searchText).toUpperCase(),
-   })
+  })
```
nawatts: Unrelated change to make indentation consistent.

PR #76, opened 2018-06-07
Refactor Manhattan plot

nawatts:
Some performance related changes to the manhattan package:
* Remove component definitions from render method
  Redefining components inside render causes those components to be unmounted/remounted on each render.
* Optimize x axis calculation to iterate over the data array fewer times

Other minor changes
* Remove unused CSS file
* Extract random color function from render method
  No need to redefine it every time Manhattan renders
* Add prop types and default props
* Remove dependency on Ramda
  There was only one use of one Ramda function (`R.pluck`)
* Fix height of x axis background
  It was set to the height of the graph instead of the height of the axis


packages/manhattan/src/index.js
```diff
@@ -1,42 +1,46 @@
...
   data,
-  title = '',
-  width = 900,
-  height = 500,
-  sexChromosomes = false,
-  showAxisBounds = false,
+  title,
+  width,
+  height,
+  includeSexChromosomes,
```
nawatts: This is entirely subjective/aesthetic, but I think code with boolean variables is easier to understand when the variable names clearly convey true/false (ex. `isFoo`, `hasBar`, etc.)

packages/manhattan/src/index.js
```diff
@@ -1,42 +1,46 @@
...
+    ({ ...acc, [chr.replace('chr', '')]: randomColor() }), {})
 
   const xScale = scaleLinear()
     .domain([0, data.length])
     .range([0 + padding, width])
 
+  const yExtent = extent(data, d => d['-log10p'])
   const yScale = scaleLinear()
-    .domain([min(yData), max(yData) + (max(yData) * 0.1)])
+    .domain([yExtent[0], yExtent[1] * 1.1])
```
nawatts: Previously, this iterated over `data` 4 times (once with `R.pluck`, once with `d3.min`, and twice with `d3.max`). Using `d3.extent` with an accessor reduces that to one iteration.

packages/manhattan/src/index.js
```diff
@@ -91,83 +93,76 @@ const ManhattanPlot = ({
...
     >
       Chromosome
     </text>
   )
 
-  const Xticks = () => {
-    const snpsSplitByChrom = plotChromosomes.map(chr =>
-      ({
-        name: chr,
-        data: data.filter(snp => chr === `chr${snp.chromosome}`),
```
nawatts: Previously, this `filter` inside of `map` required iterating over the `data` array once per chromosome (22-24 times). This change reduces it to one iteration.

packages/manhattan/src/index.js
mattsolo1: Agreed

packages/manhattan/src/index.js
mattsolo1: Nice, didn't know about extent

PR #79, opened 2018-06-08
Add case/control allele frequency columns to SCHEMA analysis groups table

nawatts:


projects/schizophrenia/src/VariantDetails/AnalysisGroupsTable.js
```diff
@@ -12,24 +12,47 @@ import {
 } from '@broad/ui'
 
 
+function formatExponential(number) {
+  return Number(number.toPrecision(4)).toExponential()
+}
```
nawatts: This is duplicated so many times... need to extract it to the ui package at some point.

PR #82, opened 2018-06-15
Prevent transcript fan out from moving region viewer

nawatts:
Currently, when the list of alternate transcripts is closed, the entire region viewer with all of the tracks shifts to the right. This is confusing because it appears as though more content is changing than just the transcript track.

## Before
screen shot 2018-06-15 at 4 52 14 pm
screen shot 2018-06-15 at 4 52 17 pm

## After
screen shot 2018-06-15 at 7 16 25 pm
screen shot 2018-06-15 at 7 16 29 pm

### DIFF
--- a/packages/track-transcript/src/TranscriptConnected.js
+++ b/packages/track-transcript/src/TranscriptConnected.js
@@ -1,7 +1,5 @@
-import React from 'react'
-import PropTypes from 'prop-types'
 import { connect } from 'react-redux'
-import { TranscriptTrack } from './index'
+import { bindActionCreators } from 'redux'
 
 import {
   currentGene,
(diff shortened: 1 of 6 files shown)


packages/track-transcript/src/TranscriptTrack.js
```diff
@@ -93,37 +93,25 @@ const TranscriptName = styled.div`
 `
 
 const GtexTitleWrapper = styled.div`
+  box-sizing: border-box;
```
nawatts: This element has `width: 100%` and `padding-left: 10px`, so with the default `box-sizing: content-box`, it extended 10px outside of the right panel.

packages/track-transcript/src/TranscriptConnected.js
```diff
@@ -92,4 +31,11 @@ const mapStateToProps = (state, ownProps) => ({
   strand: strand(state),
 })
 
-export default connect(mapStateToProps, geneActions)(TranscriptConnected)
+const mapDispatchToProps = dispatch => ({
+  onExonClick: bindActionCreators(geneActions.setCurrentExon, dispatch),
+  onTissueChange: bindActionCreators(geneActions.setCurrentTissue, dispatch),
+  onTranscriptNameClick: bindActionCreators(geneActions.setCurrentTranscript, dispatch),
+  transcriptButtonOnClick: bindActionCreators(geneActions.toggleTranscriptFanOut, dispatch),
+})
```
nawatts: Instead of using a separate component (`TranscriptConnected`) to map props from Redux to `TranscriptTrack`, we can do that directly in `mapDispatchToProps`.

mattsolo1: Yeah, this is much cleaner

packages/track-transcript/src/TranscriptConnected.js
```diff
@@ -17,69 +15,10 @@ import {
   actions as geneActions,
 } from '@broad/redux-genes'
 
-const TranscriptConnected = ({
```
mattsolo1: Nice

PR #83, opened 2018-06-18
Use position for x axis in Manhattan plot

nawatts:
Previously, all data points on the Manhattan plot were evenly spaced along the x axis despite some data points being at the same position in the genome.

For example, when plotting `resources/schizophrenia-manhattan.json`:

Before

before

After

after

### DIFF
--- a/packages/manhattan/src/index.js
+++ b/packages/manhattan/src/index.js
@@ -37,8 +37,9 @@ const ManhattanPlot = ({
 
   const yExtent = extent(data, d => d['-log10p'])
   const yScale = scaleLinear()
-    .domain([yExtent[0], yExtent[1] * 1.1])
+    .domain(yExtent)
     .range([height - padding, 10])
+    .nice()
 
   const titleText = (
     <text
--- a/packages/manhattan/src/index.js
+++ b/packages/manhattan/src/index.js
@@ -29,7 +29,7 @@ const ManhattanPlot = ({
 
   const plotChromosomes = includeSexChromosomes ? HUMAN_CHROMOSOMES : HUMAN_AUTOSOMES
   const chromosomeColors = plotChromosomes.reduce((acc, chr) =>
-    ({ ...acc, [chr.replace('chr', '')]: randomColor() }), {})
+    ({ ...acc, [chr]: randomColor() }), {})
 
   const xScale = scaleLinear()
     .domain([0, data.length])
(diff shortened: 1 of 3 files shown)


packages/manhattan/src/index.js
```diff
@@ -196,21 +168,21 @@ const ManhattanPlot = ({
...
   data: PropTypes.arrayOf(PropTypes.shape({
     snp: PropTypes.string.isRequired,
-    chromosome: PropTypes.oneOf([1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 'X', 'Y']).isRequired,
+    chromosome: PropTypes.oneOf(HUMAN_CHROMOSOMES).isRequired,
     pos: PropTypes.number.isRequired,
     '-log10p': PropTypes.number.isRequired,
   })).isRequired,
   height: PropTypes.number,
   includeSexChromosomes: PropTypes.bool,
-  showAxisBounds: PropTypes.bool,
```
nawatts: This drew boxes around the axis `g` elements and didn't seem useful.

packages/manhattan/src/index.js
```diff
@@ -196,21 +168,21 @@ const ManhattanPlot = ({
 ManhattanPlot.propTypes = {
   data: PropTypes.arrayOf(PropTypes.shape({
     snp: PropTypes.string.isRequired,
-    chromosome: PropTypes.oneOf([1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 'X', 'Y']).isRequired,
+    chromosome: PropTypes.oneOf(HUMAN_CHROMOSOMES).isRequired,
```
nawatts: We've had problems before (particularly with Elasticsearch mappings) from mixing numeric and string types for chromosomes.

packages/utilities/src/constants/chromosomes.js
```diff
@@ -1,5 +1,5 @@
-export const HUMAN_AUTOSOMES = Array.from(new Array(22), (x, i) => `chr${i + 1}`)
+export const HUMAN_AUTOSOMES = Array.from(new Array(22), (x, i) => `${i + 1}`)
```
nawatts: The ManhattanPlot component was the only place these constants were used, so I removed the `chr` to match the example data.

packages/manhattan/src/index.js
```diff
@@ -196,21 +168,21 @@ const ManhattanPlot = ({
 ManhattanPlot.propTypes = {
   data: PropTypes.arrayOf(PropTypes.shape({
     snp: PropTypes.string.isRequired,
```
mattsolo1: Eventually we'll want to change snp to something that plots the more general case of a locus (chrom-pos or xpos), so that plot can support genes as well.

PR #87, opened 2018-06-20
Fix table row rendering for rows that have not yet loaded

nawatts:
Currently, in the example code using Apollo for populating the variant table, scrolling fast enough will result in an error:

```
Uncaught TypeError: Cannot read property 'variant_id' of undefined
    at getDataCell (Table.js:312)
    at Table.js:455
    at Array.map (<anonymous>)
    at getDataRow (Table.js:454)
    at Table.js:551
    at List._this._cellRenderer (List.js:82)
    at defaultCellRangeRenderer (defaultCellRangeRenderer.js:99)
    at Grid._calculateChildrenToRender (Grid.js:871)
    at Grid.componentWillUpdate (Grid.js:669)
    at ReactCompositeComponent.js:708
```

This is because the row rendering function does not correctly account for rows where data has not yet been loaded.

### DIFF
--- a/packages/table/src/Table.js
+++ b/packages/table/src/Table.js
@@ -529,39 +529,44 @@ const headers = tableConfig => tableConfig.fields
   .filter(field => !field.disappear)
   .map(field => getHeaderCell(field))
 
-const isRowLoaded = (tableData, loadLookAhead) => ({ index }) => {
-  if (Array.isArray(tableData)) return !!tableData[index + loadLookAhead]
-  else if (Immutable.List.isList(tableData)) {
-    // Console.log(index, loadLookAhead, tableData.get(index + loadLookAhead))
-    return tableData.get(index + loadLookAhead)
+
+function getRowData(tableData, rowIndex) {
+  if (Array.isArray(tableData)) {
+    return tableData[rowIndex]
   }
-  return false
+
+  if (Immutable.List.isList(tableData)) {
+    return tableData.get(rowIndex)
+  }
+
+  return undefined
 }
 
-const tableRowRenderer = (tableConfig, tableData, searchText, showIndex, onRowClick, onRowHover) =>
-  ({ key, index, style }) => {
-    let tData
-    if (Array.isArray(tableData)) tData = tableData[index]
-    else if (Immutable.List.isList(tableData)) {
-      tData = tableData.get(index)
-    }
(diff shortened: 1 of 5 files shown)


packages/table/src/Table.js
```diff
@@ -529,39 +529,44 @@ const headers = tableConfig => tableConfig.fields
...
 }
 
-const tableRowRenderer = (tableConfig, tableData, searchText, showIndex, onRowClick, onRowHover) =>
-  ({ key, index, style }) => {
-    let tData
-    if (Array.isArray(tableData)) tData = tableData[index]
-    else if (Immutable.List.isList(tableData)) {
-      tData = tableData.get(index)
-    }
-    if (!isRowLoaded(tableData, 25)) {
```
nawatts: `isRowLoaded` returns a function, so this was treating every row as if it was loaded.

packages/table/src/Table.js
```diff
@@ -529,39 +529,44 @@ const headers = tableConfig => tableConfig.fields
   .filter(field => !field.disappear)
   .map(field => getHeaderCell(field))
 
-const isRowLoaded = (tableData, loadLookAhead) => ({ index }) => {
-  if (Array.isArray(tableData)) return !!tableData[index + loadLookAhead]
```
nawatts: Checking a different index to see if a row has been loaded could cause some rows to be loaded multiple times.

packages/table/src/Table.js
```diff
@@ -529,39 +529,44 @@ const headers = tableConfig => tableConfig.fields
...
-    else if (Immutable.List.isList(tableData)) {
-      tData = tableData.get(index)
-    }
-    if (!isRowLoaded(tableData, 25)) {
-      return <div>Loading</div>
-    }
-    const row = getDataRow(tableConfig, tData, searchText, index, showIndex, onRowClick, onRowHover)
+
+function tableRowRenderer (tableConfig, tableData, searchText, showIndex, onRowClick, onRowHover) {
+  return function({ key, index, style }) {
```
nawatts: A stylistic change, but I find `return function() { ... }` easier to read than arrow functions returning arrow functions.

packages/table/src/Table.js
mattsolo1: doh

PR #88, opened 2018-06-20
Remove unused navigator reducer

nawatts:
The navigator track's state is determined based on the variant and table reducers. The navigator reducer was unused.


packages/table/src/VariantTable.js
```diff
@@ -92,7 +92,6 @@ const mapStateToProps = (state) => {
     variantCount: finalFilteredVariantsCount(state),
     tablePosition: currentTableIndex(state),
     searchText: variantSearchText(state),
-    currentNavigatorPosition: state.navigator.currentNavigatorPosition,
```
nawatts: This `currentNavigatorPosition` was never used.

PR #89, opened 2018-06-25
Remove unused or unnecessary test files

nawatts:
* Remove unused test configuration files (mocha.opts)
* Remove tests that contain no assertions
* Remove tests that only cover third party libraries


packages/api/src/tests/csv-export.spec.js
```diff
@@ -1,83 +0,0 @@
...
-    })
-  }).then((response) => {
-    if (response.ok) { return response.json() }
-    return response.text().then((body) => {
-      throw Error(`${response.status} ${response.statusText}\n${body}`)
-    })
-  })
-}
-
-test('Given list of variant ids and arguments, return data for export.', (assert) => {
```
nawatts: This test had no assertions. It's effectively asserting that the API is up.

packages/api/src/tests/elastic-spec.js
```diff
@@ -1,483 +0,0 @@
-import test from 'tape-promise/tape'
```
nawatts: This test was only running Elasticsearch queries and didn't cover any gnomadjs code.

PR #90, opened 2018-06-26
Add track for ClinVar variants

nawatts:
screen shot 2018-07-17 at 4 56 09 pm
screen shot 2018-07-17 at 4 56 16 pm
screen shot 2018-08-09 at 1 57 41 pm

### DIFF
--- a/packages/track-variant/package.json
+++ b/packages/track-variant/package.json
@@ -7,6 +7,7 @@
     "lib"
   ],
   "dependencies": {
+    "@broad/ui": "*",
     "@broad/utilities": "*",
     "d3-scale": "^1.0.6",
     "d3-shape": "^1.2.0",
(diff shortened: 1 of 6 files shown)


mattsolo1:
What you think abou hovering over individual variants to see more information? We may also want to find out from users if it would be useful to link to the Clinvar page. From the api data it looks like you can find the relevant page by searching for the allele id, for example variant id `X-18525252-AT-A` has allele id `153551` which links to this page https://www.ncbi.nlm.nih.gov/clinvar/variation/143819/

It's confusing because the url actually uses the Variation ID, 143819, which I couldn't find in the api data.

```
{
  gene(gene_name: "CDKL5") {

    clinvar_variants {
      variant_id
      pos
      clinical_significance
      molecular_consequence
      allele_id
      
    }
  }
}
```

nawatts:
Sounds good. Looks like that ID used for the URL may be "MEASURESET_ID" in Elasticsearch.

bw2:
Another detail - for the clinvar vcf data, you want to use the `allele_id` column to generate the link to the NCBI clinvar page:
`"http://www.ncbi.nlm.nih.gov/clinvar/?term="+allele_id+"[alleleid]"`

PR #100, opened 2018-07-02
Move variant information to modal in SCHEMA browser

nawatts:
Previously, the selected variant's information was displayed below the table. When first viewing the browser, it wasn't obvious there was any content below the table. With it there, switching between variants required scrolling up and down the page, which was awkward because the table is itself scrollable.

screen shot 2018-07-02 at 3 32 34 pm


projects/schizophrenia/src/ExomePage/RegionViewer.js
```diff
@@ -175,11 +157,9 @@ export default connect(
     visibleVariants: finalFilteredVariants(state),
     screenSize: screenSize(state),
     variantFilter: variantFilter(state),
-    focusedVariant: focusedVariant(state),
```
nawatts: Removing this line makes the click handler on variant IDs in the table 10x faster (from ~300ms to ~30ms). With this line, changing the focused variant causes this SchizophreniaGeneViewer component to render, and its render method maps/filters over the variants list 5 times.

Issue #97, opened 2018-07-02
Add selected transcripts to browser history


nawatts:
After selecting a transcript on the gene page, pressing the browser's button should switch to the previously selected transcript.

nawatts:
Done in #106.

Issue #30, opened 2018-05-10
Not all exons represented in transcript models


bw2:
(from https://github.com/macarthur-lab/gnomad_browser/issues/119)
```
Hi Matt,

I'm not sure why not all exons are visible in the transcript cartoon.

e.g. http://gnomad-beta.broadinstitute.org/gene/LAMP2 has 3 main isoforms (ENSTs ending 600,639 & 335). These have 3 different terminal exons (exon 9 in each case), but otherwise are identical.
The terminal exon is visible for the first of these, but not the other 2.

http://ftp.ebi.ac.uk/pub/databases/lrgex/LRG_749.xml has nice image to show what I mean.

Is this because the exons are not in the canonical transcript perhaps?
```

PR #106, opened 2018-07-09
Set current transcript based on a URL parameter

### LINKED ISSUE #46: "Gene not found" searches should have informative page rather than javascript error
> Moved from https://github.com/macarthur-lab/gnomad_browser/issues/74
> Originally created by @mattsolo1 on *Wed, 06 Dec 2017 16:16:10 GMT*

nawatts:
Add a second route (`/gene/:gene/transcript/:transcriptId`) for the gene page and make the transcript IDs in the transcript track links to that route. This allows sharing links to specific transcripts and switching to a previously selected transcript using the browser's back button (#97).

Also:
* Separate the "Export variants" button's Redux connection from GenePage. This keeps GenePage focused on loading data for a gene.
* Move loading state to GenePage and add error state. This saves us from having to reimplement loading spinners and error messages in each project that uses gene page. The error state also resolves #46.
* Previously, GenePage would push to history when the current gene in Redux changed. Having routing depend on a particular component being mounted when dispatching an action is error prone (it only works when viewing the gene page). TopBar previously relied on this behavior. This changes TopBar to push the URL itself instead of going through Redux.


nawatts:
This will also require a change to exac.py to register the new route for the gene page.

https://github.com/macarthur-lab/gnomad_browser/blob/a3014ed0a15e5d325ea6b67a69704fcfd15e524d/exac.py#L863-L864

nawatts:
We may want to consider expanding the transcript track by default if a user reaches gnomAD from a link to a non-canonical transcript. Or when the track is closed, provide some indication that the visible variants do not match the canonical transcript (which is the only transcript displayed when the track is closed).

packages/redux-genes/src/GenePageHoc.js
```diff
@@ -1,85 +1,139 @@
...
-      this.props.resetFilter()
+      this.loadGeneData()
     }
   }
 
+  componentWillUnmount() {
+    this.mounted = false
+  }
+
+  loadGeneData() {
```
mattsolo1: Slightly confusing to have two functions named this in the same component

nawatts: Agreed. A more appropriate name for the function passed in would be `fetchGeneData`, though I wavered on that since it has a side effect of changing the store. But I suppose the GenePage shouldn't care or be aware of that. Will change.

projects/gnomad/src/GenePage/index.js
```diff
@@ -53,49 +51,28 @@ const BottomButtonSection = styled.div`
   margin-top: 20px;
 `
 
-// const loading_gifs = [
```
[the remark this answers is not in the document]
nawatts: I'm not opposed to an animated loading indicator. However, loading images from a site we don't control is a security risk and it's unwelcome load on their servers.
https://en.wikipedia.org/wiki/Inline_linking#Controversial_uses_of_inline_linking

I'd suggest we either use one of the Font Awesome [spinner icons](https://fontawesome.com/v4.7.0/icons/#spinner) or something like [SpinKit](https://github.com/tobiasahlin/SpinKit).

nawatts:
> * Like you touched on, we need to better way to display which transcript is active and which exons are relevant. Especially when a noncanonical transcript is selected and the transcript track is collapsed, we should devise a way to display the selected noncanonical transcript. Perhaps by highlighting exons on the canonical transcript itself or by showing the selected noncanonical transcript in a separate lane below.

I like showing the selected transcript as well as the canonical transcript when the transcript track is closed. I don't think all transcripts are subsets of the canonical transcript, so we would still have to account for regions in the selected transcript that don't appear in the canonical transcript (#30).

> * We will want to have a way for handling transcripts not found if a user types the wrong thing in the url. Currently, the page loads but there are no variants. Ideally we would have a message saying transcript not found for gene.

Yes, good catch. Will add this.

> * What you think about using query strings instead of parameters? Maybe the url would look something like http://localhost:8008/gene/BRCA1/transcript?id=ENST00000461574. We may want additional things in the url such as consequence category, dataset (exac vs gnomAD), etc. More info on query strings w/ react router: https://tylermcginnis.com/react-router-query-strings/

I would argue that ideally, dataset and transcript ID would be URL parameters and consequence category, include filtered variants, search string, etc would be query parameters. The difference being that dataset, gene name, and transcript identify a set of variants, while consequence category etc. filter the set of variants.

> * Eventually we will want to avoid having to reload the page for gene and transcript changes.

Agreed. I think that would be taken care of as we move to Apollo. As we push responsibility for loading various pieces of data to components lower in the tree, eventually the GenePage HOC would disappear entirely.

packages/redux-genes/src/GenePageHoc.js
```diff
@@ -1,85 +1,148 @@
...
+
+    this.props.fetchGeneData(this.props.geneName, this.props.transcriptId)
+      .then((geneData) => {
+        let loadError = null
+        if (!geneData) {
+          loadError = 'Gene not found'
+        }
+        if (
+          this.props.transcriptId
+          && !geneData.transcripts.map(t => t.transcript_id).includes(this.props.transcriptId)
```
nawatts: This is a quick fix for handling "transcript not found" but isn't great since it relies on `fetchGeneData` requesting transcripts for the gene.

PR #110, opened 2018-07-10
Fetch ClinVar data from latest index

nawatts:
Get ClinVar variants from an index populated by https://github.com/macarthur-lab/hail-elasticsearch-pipelines/blob/master/hail_scripts/load_clinvar_to_es_pipeline.py.

Address @bw2's comment about querying by gene ID instead of symbol.

### DIFF
--- a/packages/api/src/schema/types/clinvar.js
+++ b/packages/api/src/schema/types/clinvar.js
@@ -65,7 +65,6 @@ export function lookupClinvarVariantsByGeneId(client, geneId) {
           bool: {
             filter: [
               { term: { gene_ids: geneId } },
-              { exists: { field: 'main_transcript_hgvsp' } },
             ],
           },
         },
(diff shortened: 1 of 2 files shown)


packages/api/src/schema/types/clinvar.js
```diff
@@ -9,81 +10,62 @@ import { fetchAllSearchResults } from '../../utilities/elasticsearch'
...
-      _source: fields,
       body: {
         query: {
           bool: {
-            must: [
-              { term: { SYMBOL: geneName } },
-              { exists: { field: 'HGVS_P' } },
+            filter: [
+              { term: { gene_ids: geneId } },
+              { exists: { field: 'main_transcript_hgvsp' } },
```
mattsolo1: I forget, why is this filter applied?

nawatts: I don't know. It's been there since ClinVar variants were added to the API in https://github.com/macarthur-lab/gnomadjs/commit/e41e641e9f04420573c3e467620200d8f2c9d867.

mattsolo1: May want to try removing it -- it could be filtering splice variants.

PR #111, opened 2018-07-11
Update ClinVar variant query

nawatts:
The track in #90 displays ClinVar variants' main transcript's major consequence. Thus, a variant should be included in the track based on the gene ID of the variant's main transcript, not the list of all gene IDs in which the variant appears. Otherwise, the track may show an incorrect consequence for the current gene.


nawatts:
This filters out the intronic variants pointed out in the meeting this morning.

PR #113, opened 2018-07-11
Allow interaction with tooltip content

nawatts:
One of the requirements for the ClinVar track is showing a link to NCBI's website in a tooltip shown when hovering over a variant in the track.

Currently, tooltips are immediately removed when the mouse leaves the tooltip's anchor element. Thus, any link in a tooltip would be impossible to click because moving the mouse over the link would remove the tooltip and with it, the link.

This change keeps the tooltip visible while the mouse is over it. This allows putting interactive elements (such as links) in tooltips and allows users to select text shown in tooltips.

### DIFF
--- a/packages/ui/src/example/TooltipExample.js
+++ b/packages/ui/src/example/TooltipExample.js
@@ -1,0 +1,16 @@
+import React from 'react'
+
+import { withTooltip } from '..'
+
+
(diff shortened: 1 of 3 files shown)


nawatts:
Rebased to resolve conflicts.

PR #116, opened 2018-07-16
Scale expression plot to selected tissue

### LINKED ISSUE #114: Isoform expression scales should be adjusted per tissue
Currently, when you toggle between tissues with different gene expression, the scale doesn't change, making it look like none of the isoforms are expressed.

Example gene: BIN1

highly expressed in muscle (510 tpm) 

image

Not as highly expressed in esophagus, but still at appreciable levels (15) tpm. however scale is still 600 

image

nawatts:
Currently, the scale of the tissue expression plot is based on the maximum expression in *any* tissue. This often results in the plot points for all transcripts being indistinguishable from zero for all but one tissue. 

Resolves #114.

## Before
screen shot 2018-07-16 at 1 17 44 pm

## After
screen shot 2018-07-16 at 1 18 26 pm

screen shot 2018-07-16 at 1 16 29 pm


packages/track-transcript/src/GTEx.js
```diff
@@ -140,26 +145,35 @@ TissueIsoformExpressionPlotHeader.defaultProps = {
...
   const selectedTissue = currentTissue || 'median-across-all'
+  const maxExpressionValue = selectedTissue === 'median-across-all'
+    ? maxTissueExpressions.aggregate.median
+    : maxTissueExpressions.individual[selectedTissue]
+
   const padding = 10
 
   const gtexScale = scaleLinear()
-    .domain([0, maxTissueValue + (maxTissueValue * 0.3)])
+    .domain([0, maxExpressionValue])
```
mattsolo1: Perhaps make the max transcript value take 90% of the plot (rather than 100%)?

nawatts: With the padding added by `.nice()`, the max expressed transcript's usually won't be flush with the right edge of the plot.

https://github.com/d3/d3-scale#continuous_nice

Issue #51, opened 2018-05-11
Filtered variants (e.g. AC0) missing from Viewing in Table track


nawatts:
> Moved from https://github.com/macarthur-lab/gnomad_browser/issues/82
> Originally created by @konradjk on *Thu, 07 Dec 2017 15:37:35 GMT*

To reproduce:

- Go to http://35.184.206.249/gene/PCSK9
- Click LoF (observe variants in `Viewing in Table` track)
- Click Include filtered variants
- Variants are added to track above (e.g. note the first variant, a tiny AC0 circle), but not Viewing in Table track

nawatts:
> Moved from https://github.com/macarthur-lab/gnomad_browser/issues/82#issuecomment-387219715
> Originally created by @nawatts on *Mon, 07 May 2018 21:57:46 GMT*

It looks like the "Viewing in table" track filters out variants with an allele frequency of 0
https://github.com/macarthur-lab/gnomadjs/blob/3816f2be5f10e4d2ae6e8430ca19695f3637e614/packages/track-navigator/src/Navigator.js#L121

while the track above renders them
https://github.com/macarthur-lab/gnomadjs/blob/3816f2be5f10e4d2ae6e8430ca19695f3637e614/packages/track-variant/src/index.js#L154-L166

Filtered variants with allele frequency > 0 *do* appear in the "Viewing in table" track
<img width="1017" alt="screen shot 2018-05-07 at 5 56 19 pm" src="https://user-images.githubusercontent.com/1156625/39727282-16269ef0-5220-11e8-88ea-f744080fc6ed.png">

PR #123, opened 2018-07-17
Split variant track into multiple components

nawatts:
Rather than have one large configurable component for the VariantTrack, split it into two components: one for each configuration in use.

This isolates the piece (VariantAlleleFrequencyPlot) that needs to be replaced with canvas to support rendering more variants.


nawatts:
This also makes it easier to reuse the variant track's visualization in the navigator track, to prevent some inconsistencies between the two (#51).

mwoodbri:
Hi @nawatts. Just one VariantFX question on this: we've lost our transcripts as a result - any easy way to restore? This is how it used to look:

screen shot 2018-08-14 at 11 50 36

nawatts:
Hi @mwoodbri. It looks like the transcripts are still there, it's just that the transcript track defaults to showing only the canonical transcript.

There is a button to expand the track. However, it is difficult to notice because it uses a [Font Awesome](https://fontawesome.com/) icon and VariantFX does not load Font Awesome's CSS.

screen shot 2018-08-14 at 7 03 32 am

You could do one or both of:
Add Font Awesome's css to VariantFX's index.html to make the icon visible.
```html
<link rel="stylesheet" type="text/css" href="https://maxcdn.bootstrapcdn.com/font-awesome/4.7.0/css/font-awesome.min.css">
```

Dispatch the `toggleTranscriptFanOut` action after the Redux store is created in Main.js to expand the transcript track.
```js
import { actions as geneActions } from '@broad/redux-genes'

....
store.dispatch(geneActions.toggleTranscriptFanOut())
```

PR #126, opened 2018-07-18
Render variant track with canvas

nawatts:
Not having a React element per variant marker allows quickly rendering large numbers of variants.


nawatts:
That's unrelated to this PR. It's also happening on master. Not sure where that started.

nawatts:
I think it may have gotten introduced in #106. It looks like before #106, the variant filter was reset when the API request was made and after, it's reset when the response is received.

nawatts:
Default filter restored in #128.

PR #130, opened 2018-07-19
Remove redux-search

nawatts:
For large genes/regions, redux-search takes a long time to build an index. While it's building that index, entering *any* search query results in "No variants found" without any indication that searching won't work correctly yet.

Since the variant fields which are searched are not free form text and we don't order variants by relevance to the search query, we don't need the full text search that redux-search provides. A string match on the field's value will suffice.

By removing the indexing step, this change makes the variant search feature immediately available once a gene/region has loaded.

### DIFF
--- a/packages/redux-variants/variants.js
+++ b/packages/redux-variants/variants.js
@@ -2,7 +2,6 @@
 import keymirror from 'keymirror'
 import { Record, Set, OrderedMap, Map, List, fromJS } from 'immutable'
 import { createSelector } from 'reselect'
-import { createSearchAction, getSearchSelectors } from 'redux-search'
 
 import {
   isCategoryLoF,
@@ -19,6 +18,7 @@ export const types = keymirror({
   SET_FOCUSED_VARIANT: null,
   SET_SELECTED_VARIANT_DATASET: null,
   SET_VARIANT_FILTER: null,
+  SET_VARIANT_SEARCH_QUERY: null,
   SET_VARIANT_SORT: null,
   TOGGLE_VARIANT_QC_FILTER: null,
   ORDER_VARIANTS_BY_POSITION: null,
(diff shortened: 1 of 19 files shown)


packages/table/src/Table.js
```diff
@@ -634,12 +624,7 @@ const Table = ({
   onScroll,
   searchText,
   width,
-  filteredIdList,
 }) => {
-  if (searchText !== '' && filteredIdList.size === 0) {
-    return <NoVariants width={width} height={height}>No variants found</NoVariants>
-  }
```
nawatts: Moved this to VariantTable. The base Table should be agnostic about what it is rendering.

packages/redux-variants/variants.js
```diff
@@ -520,20 +506,3 @@ const sortedVariants = createSelector(
     return sortedVariants
   }
 )
-
-export const finalFilteredVariants = createSelector(
```
nawatts: Renamed `sortedVariants` to `finalFilteredVariants` to avoid changing all the imports.

packages/redux-variants/variants.js
```diff
@@ -520,20 +506,3 @@ const sortedVariants = createSelector(
...
-    if (filteredIdList.size !== 0 || variants.size === 0) {
-      return variants.filter((v) => {
-        return filteredIdList.includes(v.get('id'))
-      }).toList()
-    }
-    return variants.toList()
-  }
-)
-
-export const finalFilteredVariantsCount = createSelector(
```
nawatts: `VariantTable` used this connection, but didn't do anything with the value.

packages/table/src/example/VariantTableRedux.example.js
```diff
@@ -24,9 +24,14 @@ const sum = (oldValue, newValue) => oldValue + newValue
 const concat = (oldValue, newValue) => oldValue.concat(newValue)
 
 const appSettings = {
-  searchIndexes: ['variant_id', 'rsid', 'consequence', 'hgvsp', 'hgvsc'],
-  searchResourceSelector: (resourceName, state) => {
-    return state.variants.searchIndexed
+  variantSearchPredicate(variant, query) {
```
mattsolo1: Nice

Issue #160, opened 2018-08-03
Variant track does not display all variants


nawatts:
The browser loads variants located in or within 75bp of exon regions. However, the region viewer scale does not account for the first 75bp "padding" region, so some variants which appear in the table do not appear in the variant or navigator tracks.

For example, on the gene page for PCSK9, the first variant visible in the track is the 9th variant in the list.

<img width="1555" alt="screen shot 2018-08-03 at 2 11 34 pm" src="https://user-images.githubusercontent.com/1156625/43658700-9ffd308e-9727-11e8-8957-455c3c2de3a4.png">

<img width="1548" alt="screen shot 2018-08-03 at 2 15 44 pm" src="https://user-images.githubusercontent.com/1156625/43658776-d3605186-9727-11e8-96d3-7ec42ccd0520.png">

nawatts:
This is because the region viewer scale does not add padding to the first region, but the Elasticsearch query does.

https://github.com/macarthur-lab/gnomadjs/blob/43daf388096c9e773e609cc9c1aef83d88b07657/packages/utilities/src/coordinates/index.js#L66-L71

https://github.com/macarthur-lab/gnomadjs/blob/43daf388096c9e773e609cc9c1aef83d88b07657/packages/api/src/schema/types/elasticVariant.js#L257-L266

PR #177, opened 2018-08-09
Fix region viewer scale to account for points that fall between two regions

nawatts:
Some ClinVar variants may fall between the regions plotted in the region viewer. Currently, if a variant's position falls between two regions, its offset will be based on the first region and thus its scaled x position will be _way_ outside the region viewer to the right. With this change, it will be positioned correctly between the two regions.


nawatts:
This affects #160. Those variants in the first padding region will now all be positioned at the left edge of the region viewer, rather than off the canvas.

PR #179, opened 2018-08-10
Move getXpos function to api package

nawatts:
Currently, the API package has one dependency on another gnomadjs package: `getXpos` from utilities. Since utilities is not published to NPM and `npm install` in the Docker image doesn't use Yarn (and its workspaces configuration), this means that we have to jump through some hoops in the API's Docker build to install dependencies:

https://github.com/macarthur-lab/gnomadjs/blob/4edbf4c4166af836c7c15f6f6cc057c023e62490/packages/api/deploy/dockerfiles/gnomadgraphqlbase.dockerfile#L7-L20

`getXpos` is used in two places: the API and VariantFX. However, in VariantFX:
1. `getXpos` is called incorrectly:
https://github.com/macarthur-lab/gnomadjs/blob/4edbf4c4166af836c7c15f6f6cc057c023e62490/projects/variantfx/src/GenePage/fetch.js#L286
and thus each variant's `xpos` is set to `NaN`

2. Variants' `xpos` is never read 

Thus, we can remove `getXpos` from VariantFX, move it from utilities to API, and streamline the API Docker build.

### DIFF
--- a/packages/api/package.json
+++ b/packages/api/package.json
@@ -22,7 +22,6 @@
   "author": "",
   "license": "ISC",
   "dependencies": {
-    "@broad/utilities": "*",
     "compression": "^1.7.1",
     "core-js": "^2.5.0",
     "cors": "^2.8.1",
--- a/packages/api/src/schema/datasets/exac.js
+++ b/packages/api/src/schema/datasets/exac.js
@@ -7,7 +7,7 @@ import {
   GraphQLString,
 } from 'graphql'
 
-import { getXpos } from '@broad/utilities'
+import { getXpos } from '../../utilities/variant'
 
 import { VariantInterface } from '../types/variant'
 import {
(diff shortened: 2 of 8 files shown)


mwoodbri:
@nawatts: I have fixed this independently in one of my various un-upstreamed VariantFX changes. I'm currently working on getting closer to your mainline by basically rebasing these outstanding changes against your `master`. This will take a little while, but shouldn't prevent you correcting issues like this in the meantime, so thanks!

nawatts:
Hi @mwoodbri. Do you need the `getXpos` function in VariantFX? If so, I can leave it and duplicate to the API package. I've been looking at the `develop` branch of https://github.com/ImperialCardioGenetics/gnomadjs to attempt not to break anything in VariantFX, and I did not see variants' `xpos` fields being used there.

mwoodbri:
Hi @nawatts. Sorry for the confusion - it has taken me a while to sync up and test. But no need for `getXpos` from my perspective.

PR #186, opened 2018-08-13
Remove project/package specific Babel dependencies

nawatts:
Standardize on one version of Babel dependencies for all packages/projects, tracked in the root package.json.


packages/api/package.json
```diff
@@ -25,6 +25,7 @@
     "compression": "^1.7.1",
     "core-js": "^2.5.0",
     "cors": "^2.8.1",
+    "csvtojson": "^1.1.7",
```
nawatts: Moved this to dependencies because it is required by `api/utilities/elasticsearch.js`.

packages/api/package.json
```diff
@@ -38,16 +39,9 @@
...
   },
   "devDependencies": {
-    "babel-cli": "^6.18.0",
-    "babel-core": "^6.18.0",
-    "babel-eslint": "^7.1.0",
-    "babel-plugin-transform-object-rest-spread": "^6.16.0",
-    "babel-polyfill": "^6.16.0",
-    "babel-preset-latest": "^6.16.0",
-    "csvtojson": "^1.1.7"
+    "nodemon": "^1.11.0"
```
nawatts: Moved this to devDependencies because it is only used in `dev` NPM scripts.

nawatts:
I tested this locally, but not in Docker. Removing `babel-cli` and `babel-core` breaks the API because server.js imports both and after this change, they would not be included in the Docker image.

PR #189, opened 2018-08-14
Update API Docker build

nawatts:
* Remove `@broad/utilities` package
* Remove unnecessary files from image

Along with #187, this reduces the Docker image size by ~100MB

### DIFF
--- a/packages/api/.dockerignore
+++ b/packages/api/.dockerignore
@@ -1,0 +1,3 @@
+*
+!build
+!package.json
--- a/packages/api/deploy/dockerfiles/gnomadgraphql.dockerfile
+++ b/packages/api/deploy/dockerfiles/gnomadgraphql.dockerfile
@@ -2,10 +2,6 @@ FROM gcr.io/exac-gnomad/gnomad-api-base
 
 MAINTAINER MacArthur Lab
 
-# ENV NODE_ENV=production GRAPHQL_PORT=8000 MONGO_URL=mongodb://localhost:27017/exac
+COPY build /var/www/build
 
-COPY . /var/www
-WORKDIR /var/www
-
-ENTRYPOINT ["npm"]
-CMD ["run", "docker"]
+CMD ["node", "build/server.js"]
--- a/packages/api/deploy/dockerfiles/gnomadgraphqlbase.dockerfile
+++ b/packages/api/deploy/dockerfiles/gnomadgraphqlbase.dockerfile
@@ -2,11 +2,8 @@ FROM node:7.5.0
 
 MAINTAINER MacArthur Lab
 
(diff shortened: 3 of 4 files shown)


packages/api/.dockerignore
```diff
@@ -0,0 +1,3 @@
+*
+!build
+!package.json
```
nawatts: Fewer files in the build context means faster builds. Also, previously, we were running `npm install` in the Docker image and then also copying node_modules from the host.

packages/api/deploy/dockerfiles/gnomadgraphql.dockerfile
```diff
@@ -2,10 +2,6 @@ FROM gcr.io/exac-gnomad/gnomad-api-base
 
 MAINTAINER MacArthur Lab
 
-# ENV NODE_ENV=production GRAPHQL_PORT=8000 MONGO_URL=mongodb://localhost:27017/exac
+COPY build /var/www/build
```
nawatts: Copy only the necessary files in `build` instead of the entire api directory.

packages/api/deploy/dockerfiles/gnomadgraphql.dockerfile
```diff
@@ -2,10 +2,6 @@ FROM gcr.io/exac-gnomad/gnomad-api-base
...
 
-COPY . /var/www
-WORKDIR /var/www
```
nawatts: `WORKDIR` is inherited from the base image.

packages/api/deploy/dockerfiles/gnomadgraphql.dockerfile
```diff
@@ -2,10 +2,6 @@ FROM gcr.io/exac-gnomad/gnomad-api-base
...
-
-ENTRYPOINT ["npm"]
-CMD ["run", "docker"]
+CMD ["node", "build/server.js"]
```
nawatts: Directly start the server instead of going through the extra layer of NPM scripts.

packages/api/deploy/dockerfiles/gnomadgraphqlbase.dockerfile
```diff
@@ -2,22 +2,8 @@ FROM node:7.5.0
 
 MAINTAINER MacArthur Lab
 
-# ENV NODE_ENV=production GRAPHQL_PORT=8000 MONGO_URL=mongodb://localhost:27017/exac
-
-COPY package.json /var/www/
```
nawatts: Switch order of `WORKDIR /var/www` and `COPY package.json /var/www/` because `WORKDIR` doesn't change. Ordering layers from less frequently changed to more frequently changed makes the build cache more useful.

PR #192, opened 2018-08-14
Use Jest for tests

nawatts:
Standardize on one test framework for all packages. Jest has a lot of nice features like parallelization, snapshots, and support for monorepos.

A few packages were using mocha or tape for tests. This replaces them and fixes up any tests that were failing.

### DIFF
--- a/packages/gene-page/src/tests/test.js
+++ b/packages/gene-page/src/tests/test.js
@@ -1,21 +1,22 @@
-import test from 'tape'  // eslint-disable-line
-import { Record, OrderedMap, Set, Map, Seq, List } from 'immutable'  // eslint-disable-line
-import createGenePageStore from '../store/store'
-
-import data from '@resources/1506782078-gene-page-test-data.json'  // eslint-disable-line
+import { OrderedMap, Set } from 'immutable'
 
 import * as fromGenes from '@broad/redux-genes'
 import * as fromVariants from '@broad/redux-variants'
+import data from '@resources/1506782078-gene-page-test-data.json'
 
-function log (json) {
-  console.log(JSON.stringify(json, null, '  '))
-}
+import createGenePageStore from '../store/store'
+
+
 const sum = (oldValue, newValue) => oldValue + newValue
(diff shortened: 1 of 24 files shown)


packages/gene-page/src/tests/test.js
```diff
@@ -52,135 +57,60 @@ const appSettings = {
...
-})
-
-test('Variant search', (assert) => {
-  const store = createGenePageStore(appSettings)
-  store.dispatch(fromGenes.actions.receiveGeneData('ARSF', data.data.gene))
-  const state = store.getState()
-
-  const variants = fromVariants.visibleVariantsById(state)
-  assert.equal(variants.size, 720, 'all variants by default')
-  store.dispatch(fromVariants.actions.setVariantFilter('rare'))
```
nawatts: This variant filter doesn't exist any more.

packages/help/src/example/test.js
```diff
@@ -1,60 +1,20 @@
...
-test('Query can be set with an action.', (assert) => {
+test('Query can be set with an action.', () => {
   const store = createHelpStore()
   store.dispatch(helpActions.setHelpQuery('testing'))
   const state = store.getState()
-  assert.equal('testing', state.help.helpQuery)
-  assert.end()
-})
-
-test('Fetch stuff from database.', (assert) => {
```
nawatts: These tests depended on an API server and Elasticsearch instance which we don't have a test environment for.

packages/structure-viewer/src/redux.js
```diff
@@ -117,13 +117,10 @@ export const actions = {
        </queryRefinement>
       </orgPdbCompositeQuery>`
       dispatch(actions.startPdbSearch(currentGene))
-      superagent.post('http://www.rcsb.org/pdb/rest/search') // ?sortfield=Residue Count
+      return superagent.post('http://www.rcsb.org/pdb/rest/search') // ?sortfield=Residue Count
```
nawatts: Returning a promise was necessary for test code that waits until the following actions have been dispatched (previously the test would sleep for 2 seconds and assume the action had been dispatched by then).

PR #194, opened 2018-08-17
Update ClinVar query to be consistent with current gnomAD browser

nawatts:
Relies on the `gene_id_to_consequence_json ` field added to the ClinVar pipeline in https://github.com/macarthur-lab/hail-elasticsearch-pipelines/pull/51.


packages/api/src/schema/datasets/clinvar.js
```diff
@@ -0,0 +1,169 @@
...
+
+  return queryRegions.map(
+    ({ xstart, xstop }) => ({ range: { xpos: { gte: xstart, lte: xstop } } })
+  )
+}
+
+
+export const fetchClinvarVariantsInGene = async (geneId, ctx) => {
+  const geneExons = await lookupExonsByGeneId(ctx.database.gnomad, geneId)
+  const filteredExons = geneExons.filter(exon => exon.feature_type === 'CDS')
```
nawatts: Actually, this isn't quite consistent with the current browser. gnomAD includes 'UTR' and 'exon' regions as well. But elsewhere gnomAD beta uses only 'CDS' regions, so leaving it like this for now.

PR #199, opened 2018-08-29
Add gnomAD read data to API

nawatts:
Requires `READS_DIR` environment variable to be set in the production deployment.

For local development, `READS_DIR` can be left unset. In that case, the API will still function and the`reads` fields on gnomAD variants will be null.

### DIFF
--- a/packages/api/src/schema/datasets/gnomad.js
+++ b/packages/api/src/schema/datasets/gnomad.js
@@ -36,12 +36,12 @@ export const GnomadVariantType = new GraphQLObjectType({
           qualityMetrics: { type: VariantQualityMetricsType },
           reads: {
             type: ReadsType,
-            resolve: obj => {
+            resolve: async obj => {
               if (!process.env.READS_DIR) {
                 return null
               }
               try {
-                return resolveReads(process.env.READS_DIR, 'combined_bams_exomes', obj)
+                return await resolveReads(process.env.READS_DIR, 'combined_bams_exomes', obj)
               } catch (err) {
                 throw Error('Unable to load reads data')
               }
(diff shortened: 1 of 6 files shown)


packages/api/src/server.js
```diff
@@ -42,10 +43,13 @@ app.use(cors());
         },
       },
     }))
-    app.use('/export', (request, response) => {
-      console.log(request)
-      response.send('hello!')
-    })
+
+    if (process.env.READS_DIR) {
+      app.use('/reads', serveStatic(process.env.READS_DIR, {
```
mattsolo1: Should this default to the resources dir?

packages/api/src/schema/datasets/gnomad.js
```diff
@@ -120,17 +150,23 @@ export const fetchGnomadVariant = async (variantId, ctx) => {
   /* eslint-enable no-underscore-dangle */
   const commonData = exomeData || genomeData
 
-  return {
-    // common variant fields
+  const commonVariantFields = {
```
mattsolo1: May want to call this something different, since "common variant" already has a meaning in genetics. Perhaps `sharedVariantFields`?

packages/api/src/server.js
nawatts: I don't think so.

The IGV browser requires a reference, which is a 3GB file that I don't think we want to put in the resources directory.
https://github.com/macarthur-lab/gnomad_browser/blob/b4ff2705fd540037acc31f234f660b26162f5207/templates/variant.html#L900-L910

Even if we did include the reference, we wouldn't put all the other read data into the resources folder, so only certain variants would work correctly with resources as the default directory. Such being the case, I think it's preferable to default to not show read data at all in a development environment vs failing to load read data for most variants.

packages/api/src/schema/datasets/gnomad.js
nawatts: 👍 will change.

PR #201, opened 2018-08-29
Configure stylelint to require rules be grouped by type

nawatts:
We've been writing CSS with two different styles. Pick one and enforce it with stylelint.

https://github.com/necolas/idiomatic-css#declaration-order


.stylelintrc
```diff
@@ -1,15 +0,0 @@
-{
-  "processors": ["stylelint-processor-styled-components"],
-  "extends": [
-    "stylelint-config-standard",
-    "stylelint-config-styled-components"
-  ],
-  "syntax": "scss"
```
nawatts: There was a syntax error (no comma) here that prevented stylelint from parsing the configuration. Syntax errors are easier to see when the editor knows the file type, so I moved this to a JS file.

nawatts:
Much to my dismay, stylelint can't automatically fix problems in styled components because they rely on a preprocessor.

PR #208, opened 2018-09-05
Present empty cells as blank rather than "null" in tables

mwoodbri:
This is intended to prevent "null" being displayed in (e.g.) the HGVSc column of our Variant table.

Very open to suggestions for a cleaner and/or more appropriate solution!

### DIFF
--- a/packages/table/src/Table.js
+++ b/packages/table/src/Table.js
@@ -307,7 +307,7 @@ const getDataCell = (field, dataRow, searchText, i, onRowClick) => {
   const cellText = field.searchable ? (
     <Highlighter
       searchWords={searchText.split(/\s+/)}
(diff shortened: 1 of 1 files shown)


packages/table/src/Table.js
```diff
@@ -307,7 +307,7 @@ const getDataCell = (field, dataRow, searchText, i, onRowClick) => {
   const cellText = field.searchable ? (
     <Highlighter
       searchWords={searchText.split(/\s+/)}
-      textToHighlight={`${dataRow[dataKey]}`}
+      textToHighlight={`${dataRow[dataKey] || ''}`}
```
nawatts: This will also replace a value of `0` with an empty string, but since `searchable` is only applied to HGVSC/P and consequence columns, this should be fine.

PR #210, opened 2018-09-05
Restyle consequence category filter

### LINKED ISSUE #155: Allow filtering variants by any consequence category
Currently, the consequence category filter for variants is limited to 3 sets of categories: All variants, missense and LoF, or LoF only.

Each of the four categories should be able to be toggled individually, so that a user can view, for example, only missense variants.

### LINKED ISSUE #161: Remove default variant filter for large genes
For large genes, a default variant filter is applied (missense + LoF for > 40kbp and LoF only for > 100kbp). This was originally done because rendering large numbers of variants was too slow. Since moving the variant track to canvas this is is no longer an issue and there is no longer a need to apply the filter.

nawatts:
Allows toggling all consequence categories individually.

screen shot 2018-09-05 at 4 05 35 pm

Also adds keyboard shortcuts for toggling categories.

Closes #155.
Closes #161.

### DIFF
--- a/packages/dev-server/public/index.html
+++ b/packages/dev-server/public/index.html
@@ -6,9 +6,9 @@
     <meta content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=0" name="viewport" />
     <link type="image/x-icon" rel="shortcut icon" />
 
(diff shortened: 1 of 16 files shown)


nawatts:
This required changing `variantFilter` in the Redux store from a string to an object containing a boolean flag for each consequence category. In turn, that required changing the filter settings component for every project.

PR #213, opened 2018-09-06
Appropriately filter variants that have multiple consequences

mwoodbri:
I am _not_ suggesting that this is merged as-is. This is instead a request for comments.

In response to previous feedback from @nickywhiff I changed the category definitions such that variants with `Consequence` fields of the form `stop_gained&frameshift_variant` (e.g. 11-47364656-T-TATCA) would be classified as `lof`. Previously this would only occur if `Consequence` was either `stop_gained` or `frameshift_variant`.

I am not sure whether this is either a) logically sensible and b) if so, how we isolate this change to VariantFX (we previously talked about passing a function). Any suggestions very welcome!

### DIFF
--- a/packages/utilities/src/constants/categoryDefinitions.js
+++ b/packages/utilities/src/constants/categoryDefinitions.js
@@ -77,13 +77,13 @@ export function getCategoryFromConsequence(consequence) {
 }
 
 export function isCategoryLoF(consequence) {
-  return CATEGORY_DEFINITIONS_MAP.get('lof').contains(consequence)
+  return consequence.split('&').some(csq => CATEGORY_DEFINITIONS_MAP.get('lof').contains(csq))
 }
 export function isCategoryMissense(consequence) {
-  return CATEGORY_DEFINITIONS_MAP.get('missense').contains(consequence)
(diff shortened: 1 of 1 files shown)


nawatts:
This doesn't answer the question, but note that recent changes (1f09ecf5966817ee370b7398dd23eb56147847ea / #210) to the variants Redux code make it so that the filter is based on the result of `getCategoryFromConsequence` instead of `isCategory[LoF/Missense/MissenseOrLoF]`.

nawatts:
I think the short term solution to this problem is to have the variant reducer's configuration accept a  function along the lines of:
```js
variantMatchesConsequenceFilter: (variant, selectedCategories) => { ... }
```
which returns true/false to indicate whether to include the variant in the filtered list.

Longer term, I want to move the entire filter out of a shared package and into individual projects. `filteredVariantsById` in `redux-variants` is already "polluted" with project-specific code (`variantQcFilter` for gnomAD and `variantDeNovoFilter` for SCHEMA).

nawatts:
#214 contains the variants reducer configuration to support this.

nawatts:
Closing this in favor of #214.

Issue #164, opened 2018-08-03
Style region page consistent with gene page


nawatts:
The region viewer on the region page is positioned flush with the left side of the window, while on the gene page it's more centered. They should be consistent.

<img width="1680" alt="screen shot 2018-08-03 at 2 35 40 pm" src="https://user-images.githubusercontent.com/1156625/43659706-cf408780-972a-11e8-9a0d-c01b43c49c5e.png">

<img width="1680" alt="screen shot 2018-08-03 at 2 35 42 pm" src="https://user-images.githubusercontent.com/1156625/43659707-cf4a6444-972a-11e8-9763-5256366246cf.png">

PR #217, opened 2018-09-07
Size table based on container width instead of window size

nawatts:
It's much easier to reason about how a variant table component will fit into a page layout when it respects its container's size instead of sizing itself based on the window width.


nawatts:
The changes to the gnomAD region page here go a long way towards #164, though there are still some minor differences in the region viewer section.

PR #227, opened 2018-09-13
Access elasticsearch through a proxy rather than an exposed load balancer

mattsolo1:


packages/api/deploy/README.md
```diff
@@ -31,7 +31,7 @@ docker run --rm -ti --init \
    -p 8007:80 \
    -e "GRAPHQL_PORT=80" \
    -e "NODE_ENV=development" \
-   -e "ELASTICSEARCH_URL=elastic:9200" \
+   -e "ELASTICSEARCH_URL=localhost:8001/api/v1/namespaces/default/services/elasticsearch:9200/proxy" \
```
nawatts: This would be pointed at the container's port 8001. It should be the host's (`host.docker.internal:8001/api/...`).

packages/api/deploy/gnomad-api-controller.yaml
```diff
@@ -26,7 +26,7 @@ spec:
         - name: GNOMAD_MONGO_URL
           value: mongodb://gnomad-mongo:27017/exac
         - name: ELASTICSEARCH_URL
-          value: 10.4.0.11:9200
+          value: 10.7.245.23:9200
```
nawatts: Similar to how `GNOMAD_MONGO_URL` uses the `gnomad-mongo` service name, can this use `elasticsearch` instead of a hard coded IP address that will differ on every cluster.

packages/api/deploy/Makefile
```diff
@@ -27,11 +27,11 @@ gnomad-api-service:
 	--type="LoadBalancer"
 
 set-max-result:
-	curl -XPUT "http://elastic:9200/*/_settings" -d '{ "index" : { "max_result_window" : 500000 } }'
+	curl -XPUT "http://localhost:8001/api/v1/namespaces/default/services/elasticsearch:9200/proxy/*/_settings" -d '{ "index" : { "max_result_window" : 500000 } }'
 
 set-elastic-connect-timeout:
 	#  this doesn't work, must be set in elasticsearch.yaml
-	curl -XPUT "http://elastic:9200/_cluster/settings" -d '{ "transient" : { "transport.tcp.connect_timeout" : "60s" } }'
+	curl -XPUT "http://localhost:8001/api/v1/namespaces/default/services/elasticsearch:9200/proxy/_cluster/settings" -d '{ "transient" : { "transport.tcp.connect_timeout" : "60s" } }'
```
nawatts: If this doesn't work, can we remove it from this file and add it to the appropriate deployment configuration if it's needed?

Issue #204, opened 2018-08-30
Some ExAC variants not displayed correctly on gene page table


nawatts:
Some ExAC variants (ex. 1-55505552-A-ACTG) are missing allele count and frequency in the gene page's variant table. (To see this, view /gene/PCSK9 and filter variants with "55505552").

In Elasticsearch, this variant has an array of 6 values for AC_Adj instead of the expected 1. It is also a multi-allelic variant with 6 possible alternates. So it looks like the root cause is that some fields in the VDS file were not correctly split for multi-allelic variants before exporting to Elasticsearch.

nawatts:
The API also returns 369 errors in the response for PCSK9 data. They all look to be the same type: returning an array of numbers when the GraphQL schema expects a single number.

PR #231, opened 2018-09-14
Handle multi-allelic variants when loading ExAC data

nawatts:
The loading pipelines for ExAC variants currently does not split field values for multi-allelic variants. When the API encounters an array where it expects a single value, it throws an error and returns null for the field. This results in some fields for multi-allelic variants not showing up in the gene/transcript/region pages (#204).


nawatts:
The plan is to re-run VEP on ExAC and output a split Hail table and have the browser load that.

Issue #162, opened 2018-08-03
Combine exome/genome variant data in API


nawatts:
Currently, the browser requests exome/genome data for gnomAD variants individually and combines them client side. This results in downloading unnecessary data since many fields (VEP annotations, position, ID) are the same in the exome and genome data. Instead, the exome/genome data should be combined server side.

nawatts:
Done for querying individual variants in #174.

PR #242, opened 2018-09-20
Show variants on region page only for regions smaller than 10k base pairs

### LINKED ISSUE #234: gnomAD region page may show incomplete data
The region page fetches max 5000 variants from each of the gnomAD exomes and genomes and ExAC Elasticsearch indices. If the region being viewed contains more than 5000 exome/genome or ExAC variants, there is no indication that the region actually contains more variants than are displayed.

https://github.com/macarthur-lab/gnomadjs/blob/ee61e3f159bc3aae136972920289ca5ab560d69f/packages/api/src/schema/types/region.js#L209-L278

### LINKED ISSUE #235: Consequence category bar plot on region page is based on wrong dataset
For regions larger than 50k base pairs, the region page shows a stacked bar plot of binned variant consequence categories. However, this plot is based on a different Elasticsearch index than the ones used for all other gnomAD variant data. It seems to contain combined exome/genome variants from gnomAD 2.0.1.

https://github.com/macarthur-lab/gnomadjs/blob/ee61e3f159bc3aae136972920289ca5ab560d69f/packages/api/src/schema/types/region.js#L151-L186

nawatts:
Since the browser currently fetches all variants in the selected gene or region. However, there is a limit to how many can be loaded without degrading performance. Previously, the region page would only fetch the first 5000 variants (by position) from each of gnomAD exomes/genomes without any indication that the variant list was incomplete.

This change clarifies the limit. For regions containing > 30k variants, the region page does not fetch individual variants or show the variant tracks or table (replacing them with a message "This region has too many variants to display. To view individual variants, select a smaller region". The coverage and genes tracks are still available for regions with a large number of variants.

The limit isn't exact for gnomAD because it rejects any region than contains more than 30k variants in gnomAD exomes or 30k variants in gnomAD genomes. So a region than contains fewer than 30k in each of those, but more than 30k in the two combined (due to variants that don't overlap) is possible. This should be resolved with #162.

Resolves #234

### DIFF
--- a/packages/api/src/schema/types/elasticVariant.js
+++ b/packages/api/src/schema/types/elasticVariant.js
@@ -541,6 +541,24 @@ export const lookupElasticVariantsByInterval = ({ elasticClient, index, dataset,
   })
 }
 
+export const countVariantsInRegion = async ({ elasticClient, index, xstart, xstop }) => {
+  const response = await elasticClient.count({
+    index,
+    type: 'variant',
+    body: {
+      query: {
+        bool: {
+          filter: {
+            range: { xpos: { gte: xstart, lte: xstop } },
+          },
+        },
+      },
+    },
+  })
+
+  return response.count
+}
+
 export const lookupElasticVariantsInRegion = async ({ elasticClient, index, xstart, xstop }) => {
   const fields = [
     'hgvsp',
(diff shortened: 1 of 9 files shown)


nawatts:
I meant to remove the default variant filter and then forgot about it. The two pages should definitely be consistent. We can do that either by showing all variants by default or by setting the default filter on both pages based on the number of variants.

nawatts:
Removed the default filter for regions > 50kbp because:
a) that's how the gene page currently behaves
b) it was the simpler option

PR #246, opened 2018-09-26
Handle multi-allelic variants when loading ExAC data

nawatts:
The current loading pipeline for ExAC variants does not split field values for multi-allelic variants. When the gnomAD API encounters an array where it expects a single value, it throws an error and returns null for the field. This results in some fields for multi-allelic variants not showing up in the gene/transcript/region pages (#204).

It appears that the current pipeline also re-ran VEP on ExAC instead of using the VEP annotations from the VCF.

This new pipeline is split into two parts:
* export_exac_vcf_to_ht.py: Parses stringified VEP annotations in VCF, formats variants for the browser, adds derived fields, and writes a Hail table.
* export_ht_to_es.py: Reads the Hail table output by export_exac_vcf_to_ht.py and loads it into Elasticsearch. It stores sortedTranscriptConsequences values as nested objects instead of JSON encoding the list of all consequences.


projects/gnomad/data/export_exac_vcf_to_ht.py
```diff
@@ -0,0 +1,528 @@
+# Usage:
```
mattsolo1: How should we document loading script usage going forward? I greatly prefer it to be executable through a Makefile directive, but I realize this comes with the cost of keeping track of file paths/variables.

nawatts:
Added an optional `subset` argument to the script to convert the ExAC VCF to a Hail table and a README with the steps to load ExAC into Elasticsearch.

Issue #169, opened 2018-08-03
Load variants only for selected dataset


nawatts:
Currently, the gene page loads both ExAC and gnomAD variants. This makes switching between datasets fast, but it isn't feasible to continue once we add the various gnomAD subsets. Instead, the gene page should only load variants for the selected dataset.

PR #272, opened 2018-10-09
Combine gnomAD exome and genome data in API

### LINKED ISSUE #162: Combine exome/genome variant data in API

nawatts:
Currently, the browser loads gnomAD exome variants and genome variants separately and then combines them client side. Since #239 removed the option to view exomes/genomes separately, they can now be combined on the API. This saves bytes on having to download duplicate VEP annotations, ID fields, etc for variants that appear in both exomes and genomes. And treating gnomAD as one logical dataset instead of two that are combined makes #169 easier to implement.

Resolves #162


packages/api/src/schema/datasets/gnomad_r2_0_2/countGnomadVariantsInRegion.js
```diff
@@ -0,0 +1,39 @@
...
+    type: 'variant',
+    body: {
+      query: {
+        bool: {
+          filter: [{ term: { contig: chrom } }, rangeQuery],
+        },
+      },
+      aggs: {
+        unique_variants: {
+          cardinality: {
```
mattsolo1: This is cool

packages/api/src/schema/datasets/gnomad_r2_0_2/mergeExomeAndGenomeVariants.js
```diff
@@ -0,0 +1,106 @@
+const add = (n1, n2) => (n1 || 0) + (n2 || 0)
+
+const mergeExomeAndGenomeVariants = (exomeVariants, genomeVariants) => {
```
mattsolo1: We may need to think of a more general way to merge variant datasets down the road -- it will be a requirement for other projects as well.

PR #283, opened 2018-10-09
Prevent consequence filter control from wrapping at any window width

### LINKED ISSUE #282: Consequence filters on multiple lines
Didn’t notice this earlier but PR #277 appears to have caused the consequence filters and help topic icon to display on multiple lines at 100% zoom and full screen width, 15" mbp. 

image

nawatts:
Currently, certain window widths cause the consequence filter control to wrap to multiple lines.

46690264-d65eb580-cbcf-11e8-84ad-6d84f3d49a62

This change has the settings section switch from row to column layout at 1300px instead of 900px and adds a more phone-friendly layout for the consequence filter control for windows less than 700px wide.

screen shot 2018-10-09 at 3 29 55 pm

Resolves #282

### DIFF
--- a/packages/ui/src/ConsequenceCategoriesControl.js
+++ b/packages/ui/src/ConsequenceCategoriesControl.js
@@ -20,6 +20,17 @@ const categoryLabels = {
   other: 'Other',
 }
 
+// The max-width styles here are based on the filter settings
+// layout in the gnomAD browser.
+
+const ConsequenceCategoriesControlWrapper = styled.div`
+  @media (max-width: 700px) {
+    display: flex;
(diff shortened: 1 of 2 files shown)


nawatts:
Google Analytics says ~9% of users have a screen resolution 1280px wide, so they would see the column layout with a full screen window.

nawatts:
Another option is to, as window width decreases, first move "Include filtered variants" underneath the consequence categories and keep the search field right aligned, then switch to the vertical layout when consequence categories and the search field can't fit.

mattsolo1:
This looks good to me. 

image

PR #386, opened 2019-01-23
Kev/indel snp filters

knguyen142:
## Feature
Added filters for SNPs and Indels per https://github.com/macarthur-lab/gnomadjs/issues/288. 

## Design
Simple design, added checkboxes to the qc filter.
screen shot 2019-01-23 at 12 52 26 pm
Mobile
screen shot 2019-01-23 at 12 53 00 pm

## Functionality
By default, filtered variants unselected, and snp and indel both selected (on gnomad). Removing either filters out the corresponding variants.

## Implementation
Added the logic for checkboxes by copying the qcfilter checkbox, and the filter logic is compounded and done as a last step on the most filtered subset. The filter checks to see if the variant is a snp or indel based on variant id, and returns true if the corresponding filter is selected.


packages/redux-variants/variants.js
```diff
@@ -84,6 +96,7 @@ export const actions = {
 
 const defaultVariantMatchesConsequenceCategoryFilter = (variant, selectedConsequenceCategories) => {
   const category = getCategoryFromConsequence(variant.consequence) || 'other'
+  console.log(selectedConsequenceCategories)
```
nawatts: Looks like this is left over from development?

packages/redux-variants/variants.js
```diff
@@ -383,6 +412,24 @@ const filteredVariantsById = createSelector(
       filteredVariants = filteredVariants.filter(v => searchPredicate(v, query))
     }
 
+    // Indel and Snp filters.
+    filteredVariants = filteredVariants.filter(v => {
+      let splits = v.get('variant_id').split('-')
```
nawatts: I forgot to mention it, but we have ESLint and Prettier set up for a style guid... this doesn't pass the [prefer-const](https://eslint.org/docs/rules/prefer-const) rule.

packages/redux-variants/variants.js
```diff
@@ -383,6 +412,24 @@ const filteredVariantsById = createSelector(
...
+      if (splits.length === 4) {
```
nawatts: It's safe to assume that variant IDs have 4 parts.

knguyen142: Do we do validation somewhere before loading? My only concern with this is if for some reason this isn't true, the page will break (unless we have a root exception handler that swallows it).

nawatts: It's not validated anywhere, but it's defined as `chrom-pos-ref-alt` in our variant datasets.

https://github.com/macarthur-lab/hail-elasticsearch-pipelines/blob/4322dc60a22eb42d1ac35a90ad4df158c0421c61/hail_scripts/v02/utils/computed_fields/variant_id.py#L46-L59

PR #392, opened 2019-01-28
[gnomAD] Redirect HTTP requests to HTTPS

nawatts:


nawatts:
I haven't tried it yet. I wanted to get feedback on this proposed approach first.

For the files containing the absolute URL:
* .github/ISSUE_TEMPLATE.md and projects/gnomad/src/client/TermsPage.js have the URL in text visible to users.
* projects/exome-results-browsers/src/client/VariantDetails/VariantDetails.js is part of other applications (http://schema.broadinstitute.org and http://epi25.broadinstitute.org)
* projects/gnomad/src/client/opensearch.xml is used by web browsers to create a custom search engine
* projects/gnomad/start.sh sets the API URL for a browser running on localhost.

projects/gnomad/src/server/server.js
```diff
@@ -8,6 +8,15 @@ const port = process.env.PORT || 80
 const app = express()
 app.use(compression())
 
+// Redirect HTTP requests to HTTPS
+app.use((request, response, next) => {
+  if (request.get('x-forwarded-proto') === 'http') {
```
knguyen142: Looks like there is an express check for security https://expressjs.com/en/api.html#req.secure

projects/gnomad/src/server/server.js
```diff
@@ -8,6 +8,15 @@ const port = process.env.PORT || 80
...
+    response.redirect(307, `https://${request.get('host')}${request.url}`)
```
knguyen142: Is it a 301 redirect? https://support.google.com/webmasters/answer/6073543?hl=en, what happens when we don't specify a status code?

knguyen142:
I tried to run locally but couldn't through webpack, does webpack not use server.js? Also ran `node server.js` but ran into some import errors.

projects/gnomad/src/server/server.js
nawatts: I'm not sure if that will work, since the connection from load balancer to express is unencrypted.

projects/gnomad/src/server/server.js
nawatts: [307](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/307). MDN recommended 307/308 to avoid changing the request method/body. I was thinking that would be necessary for GraphQL requests to the API, but this code won't see those requests. So 301 should work.

nawatts:
> I tried to run locally but couldn't through webpack, does webpack not use server.js? Also ran node server.js but ran into some import errors.

Since all server.js has done thus far is serve static files, `yarn start`/`start.sh` do not run it. Those scripts only run `webpack-dev-server`. To run server.js locally, you can use one of:
```
NODE_ENV=development yarn build
PORT=8008 ./dist/server.js
```
or
```
PORT=8008 ../../node_modules/.bin/babel-node --config-file=../../babel.config.js ./src/server/server.js
```

projects/gnomad/src/server/server.js
nawatts: I was also thinking that this way would only redirect in the load balanced environment, while still allowing server.js to be run on HTTP locally.

projects/gnomad/src/server/server.js
knguyen142: Should we add a redirect or deny http requests for graphql too?

projects/gnomad/src/server/server.js
knguyen142: Do we need to support using this script non-https locally if we have webpack for development? Also, looks like it only checks 'https' == req.protocol;`? I can help with testing tomorrow too, should be able to spin up a local nginx pretty quick, or is there another way to test? Should we just do it in prod during non-peak time?

PR #394, opened 2019-01-28
[gnomAD] Improve formatting of transcript consequence details

### LINKED ISSUE #354: Improve formatting of LoF info on variant page
Example: http://gnomad.broadinstitute.org/variant/15-79058622-GC-G

nawatts:
Format details of LoF and missense transcript consequences so that each attribute appears on a different line.

---
LoF (15-79058622-GC-G)

Before:
lof_before

After:
lof_after

---
Missense (1-55505545-C-T)

Before:
missense_before

After:
missense_after

Resolves #354

### DIFF
--- a/projects/gnomad/src/client/VariantPage/TranscriptConsequenceDetails.js
+++ b/projects/gnomad/src/client/VariantPage/TranscriptConsequenceDetails.js
@@ -1,7 +1,39 @@
 import PropTypes from 'prop-types'
 import React from 'react'
+import styled from 'styled-components'
 
 import { getCategoryFromConsequence } from '@broad/utilities'
+
+const AttributeName = styled.dt`
+  display: inline;
+
+  ::after {
+    content: ': ';
+  }
+`
+
+const AttributeValue = styled.dd`
+  display: inline;
+  margin: 0;
+`
+
+const AttributeList = styled.dl`
+  display: flex;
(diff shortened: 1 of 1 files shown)


nawatts:
Making the keys bold can't be done with the current typography in this section. Making them bold would give them the same level of emphasis as the gene name heading. I think keeping everything left aligned makes this section easier to scan. There is already two levels of indentation for the genes and scripts... I don't think we need to add a third.

projects/gnomad/src/client/VariantPage/TranscriptConsequenceDetails.js
```diff
@@ -1,8 +1,40 @@
...
+    content: ': ';
+  }
+`
+
+const AttributeValue = styled.dd`
+  display: inline;
+  margin: 0;
+`
+
+const AttributeList = styled.dl`
```
knguyen142: Should we do validation on AttributeList's children, a la https://stackoverflow.com/questions/27366077/only-allow-children-of-a-specific-type-in-a-react-component/36424582#36424582?

nawatts: I had not considered validating children/props of styled components, but that makes sense for cases where there's some expected nesting.

PR #395, opened 2019-01-29
[gnomAD] Add tooltips with exact numbers to constraint table

### LINKED ISSUE #368: Add tooltip with precise constraint numbers
Numbers in the gene constraint table are truncated to 1-2 decimal places. These should have a hover tooltip with the precise value.

nawatts:
In the gene constraint table, the expected number of variants is truncated to 1 decimal place. In some cases, this makes the observed/expected ratio appear incorrect. This adds a tooltip with exact values for numbers shown in the table.

For example, [OR7C2](http://gnomad.broadinstitute.org/gene/ENSG00000127529):
screen shot 2019-01-29 at 10 50 37 am

Resolves #368

### DIFF
--- a/projects/gnomad/src/client/GenePage/GnomadConstraintTable.js
+++ b/projects/gnomad/src/client/GenePage/GnomadConstraintTable.js
@@ -1,7 +1,7 @@
 import PropTypes from 'prop-types'
 import React from 'react'
 
-import { BaseTable } from '@broad/ui'
+import { BaseTable, TooltipAnchor, TooltipHint } from '@broad/ui'
 
 import { Query } from '../Query'
 import StatusMessage from '../StatusMessage'
@@ -98,7 +98,17 @@ Graph.propTypes = {
   upper: PropTypes.number.isRequired,
 }
 
-const renderNumber = (num, precision) => (num === null ? '—' : num.toFixed(precision))
+const renderRoundedNumber = (num, precision = 1, tooltipPrecision = 3) => {
+  if (num === null) {
+    return '—'
+  }
+
+  return (
+    <TooltipAnchor childRefPropName="innerRef" tooltip={num.toFixed(tooltipPrecision)}>
+      <TooltipHint>{Number(num.toFixed(precision)).toString()}</TooltipHint>
+    </TooltipAnchor>
+  )
+}
 
 const renderOECell = (constraintData, category, style) => {
   const value = constraintData[`oe_${category}`]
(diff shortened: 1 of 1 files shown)


projects/gnomad/src/client/GenePage/GnomadConstraintTable.js
```diff
@@ -98,7 +98,17 @@ Graph.propTypes = {
   upper: PropTypes.number.isRequired,
 }
 
-const renderNumber = (num, precision) => (num === null ? '—' : num.toFixed(precision))
+const renderRoundedNumber = (num, precision) => {
+  if (num === null) {
```
knguyen142: Would it make sense to return '-' by default, and only return tooltipanchor if it is what we expect (num/float)? i.e.
psuedocode:
```
if typeof num is number:
  return tooltip
return '-'
```
Kind of like whitelisting instead of blacklisting.

nawatts: I would argue that if something of an unexpected type gets passed to this function, then it's a sign of some logic mistake and `num.toFixed(precision)` throwing an error and calling attention to the problem is preferable to accepting it and returning "-".

The values which are passed to this function are properties of gnomAD gene constraint, which GraphQL coerces to numbers (or null) when they are returned from the API.
https://github.com/macarthur-lab/gnomadjs/blob/c239ac4deaaa9846359ea889e18932634d4f1416/packages/api/src/schema/datasets/gnomad_r2_1/GnomadConstraintType.js#L3-L34

knguyen142: Agreed that identifying invalid forms and bringing this up to devs is best (i.e. fail silently and record error). If this throws and exception, is it silently caught and reported?

nawatts: If it throws, it's not caught, it prevents the component from rendering. But that same risk is present in every component. PropTypes will warn about an incorrect type in development, but won't prevent it from causing an error. Verifying the type of every argument/prop at run-time would be a huge amount of work.

With the exception of variants being created as Immutable.js Records with a defined set of fields, the API response schema is not validated on the client side.

nawatts: Makes more sense after talking in person... Validating the API response is a good place to start for greater type safety since the API can be deployed separately from the browser UI. Also filed #396 for tracking exceptions in the browser.

nawatts:
The precision shown in the tooltip is the same precision available in the Hail table.

```python
ds = hl.read_table("constraint.ht")
row = ds.filter(ds.transcript == "ENST00000248072").take(1)[0]
row.exp_lof
# 0.1254235603820917
```

nawatts:
Updated precision based on feedback from @konradjk

PR #397, opened 2019-02-07
Update gnomAD K8S configuration

nawatts:
Update the Kubernetes configuration files for the gnomAD browser to reflect reality.

* Remove unused volume mounts - only the API containers mount those disks
* Remove unnecessary security context configuration
* Remove unused environment variables
* Remove unnecessary restart and DNS policy configuration - those settings are the defaults
* Remove image pull policy - now that we tag images instead of always using `latest`, we can use the default `IfNotPresent` setting
* Remove gnomad-service.yaml - this has been replaced by gnomad-ingress.yaml


knguyen142:
Awesome, does that mean you can apply these changes and they're safe (so would not change anything in prod). How was this validated/tested?

nawatts:
It's based on the current configuration of deployments in GKE, so it's "tested" in production 😬.

Though it can't be applied with `kubectl` since our deployments were not originally created with `kubectl apply` or `kubectl create --save-config`. Also we haven't been tagging images with "latest", so `gcr.io/exac-gnomad/gnomad-browser-beta` doesn't match an image. I'll update the deployment scripts to add that tag.

knguyen142:
Oh interesting, what would happen if we do kubectl apply? So is this an exact copy and paste of the prod config?

nawatts:
I stand corrected... `apply` worked on the development deployment.

I didn't think it would work based on https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands#apply
> To use 'apply', always create the resource initially with either 'apply' or 'create --save-config'.

It's not an exact copy and paste because the YAML on GKE fills in default values for more fields in the spec.

knguyen142:
That might be because GKE internally calls kubectl apply.
I think the benefits of having a source controlled config file are to store configs for PR purposes as well as quick automation.
What do you think of the following?
1. We eventually get the source control to deployable state (where diff is null) and we use that to deploy changes.
2. If 1 is too much work, we use the file as a store for the GKE yaml, and we organize the GKE yaml so that every time we make a change, we can just copy and paste our file into the user section so it's synchronized?

nawatts:
Found an explanation for how `apply` works at https://kubernetes.io/docs/concepts/cluster-administration/manage-deployment/#kubectl-apply
> Currently, resources are created without this annotation, so the first invocation of kubectl apply will fall back to a two-way diff between the provided input and the current configuration of the resource. During this first invocation, it cannot detect the deletion of properties set when the resource was created. For this reason, it will not remove them.

> 1. We eventually get the source control to deployable state (where diff is null) and we use that to deploy changes.
> 2. If 1 is too much work, we use the file as a store for the GKE yaml, and we organize the GKE yaml so that every time we make a change, we can just copy and paste our file into the user section so it's synchronized? And if if we ever need to recreate it, we can copy and paste ours.

With the exception of the image tag, I believe these files are deployable. I don't think that the source YAML and GKE version can be kept exactly the same or copied back and forth, since some of the metadata and status information that GKE adds is only relevant there.

How about this?
* [ ] `kubectl apply` production config in this PR (this gives us a starting point for future `kubectl apply`s)
* [ ] In deploy scripts, tag image with `latest` so that creating a new deployment with these YAML files gets the latest image
* [ ] Add to gnomad/deploy/README.md a section about process for updating environment, etc: commit to YAML files, PR and review, kubectl apply, verify and merge

knguyen142:
> I don't think that the source YAML and GKE version can be kept exactly the same or copied back and forth, since some of the metadata and status information that GKE adds is only relevant there.

Right, but according to the docs you sent, this won't be a problem when we do kubectl apply since it won't delete it, right?

Yeah I can see splitting up 'apply' and 'deploy', we don't need to advertise the image tag in this file (or this repo at least). We might consider splitting up the project repo with our specific configuration/deployment to another repo, where version pinning might make sense.

nawatts:
> We might consider splitting up the project repo with our specific configuration/deployment to another repo, where version pinning might make sense.

Yeah, this might be good. This repository has already accumulated some configuration values specific to our deployment like GCP disk names and IP addresses.

PR #404, opened 2019-03-08
[gnomAD] Update download page links to r2.1.1

mattsolo1:


projects/gnomad/src/client/DownloadsPage.js
```diff
@@ -39,202 +39,202 @@ const DownloadSection = styled.div`
...
+    '17.8 GiB',
+    'https://storage.googleapis.com/gnomad-public/release/2.1.1/vcf/genomes/gnomad.genomes.r2.1.1.sites.X.vcf.bgz',
   ],
 ]
 
 const createVcfLink = file => {
   const fileSize = file[0]
   const url = file[1]
-  const chrom = /chr(\w+)/.exec(url)[0]
+  const chrom = /sites.(\w+)/.exec(url)[0]
```
nawatts: Since the only difference between VCF URLs is the chromosome number, it would be a lot simpler if the list contained only the chromosome instead of the full URL and this function built the URL from the chromosome instead of parsing the chromosome out of the URL.

projects/gnomad/src/client/DownloadsPage.js
```diff
@@ -249,14 +249,33 @@ export default () => (
   <InfoPage>
     <PageHeading>Downloads</PageHeading>
     <p>
-      The variant dataset files below contain all subsets and are large in size. We recommend using{' '}
-      <ExternalLink href={'https://Hail.is/'}>Hail 0.2</ExternalLink> to work with gnomAD data. It
-      is easiest to download these files in parallel on the commandline using{' '}
+      gnomAD is available for download in VCF and Hail Table (.ht) formats. The variant dataset
+      files below contain all subsets (non-neuro, non-cancer, controls only, and non-TOPMed). The
+      files can be downloaded in parallel on the commandline with{' '}
```
nawatts: I think this is usually "command line" or "command-line" vs "commadline".

projects/gnomad/src/client/DownloadsPage.js
```diff
@@ -249,14 +249,33 @@ export default () => (
...
+      <code>gsutil ls gs://gnomad-public/release/2.1.1</code>
+    </p>
+
+    <p>
+      Download all exome VCF files locally, run:{' '}
+      <code>gsutil -m cp gs://gnomad-public/release/2.1.1/vcf/exomes .</code>
+    </p>
+    <p>
+      To work effciently with gnomAD, we recommend using{' '}
+      <ExternalLink href={'https://Hail.is/'}>Hail </ExternalLink> and our{' '}
```
nawatts: Our ESLint rules specify that props with string values should not use curly-braces. `href="..."` instead of`href={'...'}`. Same comment for several links below.

projects/gnomad/src/client/DownloadsPage.js
```diff
@@ -249,14 +249,33 @@ export default () => (
...
+      For example, after installing gsutil, start navigating with this command:{' '}
+      <code>gsutil ls gs://gnomad-public/release/2.1.1</code>
+    </p>
+
+    <p>
+      Download all exome VCF files locally, run:{' '}
+      <code>gsutil -m cp gs://gnomad-public/release/2.1.1/vcf/exomes .</code>
+    </p>
+    <p>
+      To work effciently with gnomAD, we recommend using{' '}
```
nawatts: Efficiently is misspelled.

projects/gnomad/src/client/DownloadsPage.js
nawatts: The Hail team requested in October that we specify the version number here. I'm not sure if that's still relevant now that Hail 0.2 has been out for a while. @konradjk

projects/gnomad/src/client/DownloadsPage.js
```diff
@@ -286,17 +305,18 @@ export default () => (
               >
                 gs://gnomad-public/release/2.1/coverage/exomes/gnomad.exomes.r2.1.coverage.ht
               </ExternalLink>
+              <br />
```
nawatts: This `br` element is unnecessary.

projects/gnomad/src/client/DownloadsPage.js
```diff
@@ -286,17 +305,18 @@ export default () => (
...
             </li>
           </ul>
         </DownloadSection>
         <DownloadSection>
           <h3>VCFs</h3>
           <ul>
             <li>
-              <ExternalLink href="https://storage.googleapis.com/gnomad-public/release/2.1/vcf/exomes/gnomad.exomes.r2.1.sites.vcf.bgz">
-                All chromosomes (59.23 GiB){' '}
+              <ExternalLink href="https://storage.googleapis.com/gnomad-public/release/2.1.1/vcf/exomes/gnomad.exomes.r2.1.sites.vcf.bgz">
```
nawatts: These two links are incorrect. They have the "2.1.1" folder but the filename contains "2.1.sites".

projects/gnomad/src/client/DownloadsPage.js
```diff
@@ -338,18 +356,18 @@ export default () => (
...
               </ExternalLink>
-              <ExternalLink href="https://storage.googleapis.com/gnomad-public/release/2.1/vcf/genomes/gnomad.genomes.r2.1.sites.vcf.bgz.tbi">
+              <ExternalLink href="https://storage.googleapis.com/gnomad-public/release/2.1.1/vcf/genomes/gnomad.genomes.r2.1.1.sites.vcf.bgz.tbi">
                 (.tbi)
               </ExternalLink>
             </li>
             <li>
-              <ExternalLink href="https://storage.googleapis.com/gnomad-public/release/2.1/vcf/genomes/gnomad.genomes.r2.1.exome_calling_intervals.sites.vcf.bgz">
-                Exome calling intervals (9.76 GiB){' '}
+              <ExternalLink href=" https://storage.googleapis.com/gnomad-public/release/2.1.1/vcf/genomes/gnomad.genomes.r2.1.1.exome_calling_intervals.sites.vcf.bgz">
```
nawatts: This URL shouldn't start with a space.

projects/gnomad/src/client/DownloadsPage.js
```diff
@@ -365,13 +383,24 @@ export default () => (
...
-            <ExternalLink href="https://console.cloud.google.com/storage/browser/gnomad-public/release/2.1/ht/constraint">
-              Hail Table (gs://gnomad-public/release/2.1/ht/constraint/constraint.ht)
+            <ExternalLink href="https://console.cloud.google.com/storage/gnomad-public/release/2.1.1/constraint/gnomad.v2.1.1.lof_metrics.by_transcript.ht">
+              Hail Table
+              (gs://gnomad-public/release/2.1.1/constraint/gnomad.v2.1.1.lof_metrics.by_transcript.ht)
+            </ExternalLink>
+          </li>
+          <li>
+            <ExternalLink href="https://storage.googleapis.com/gnomad-public/release/2.1.1/constraint/gnomad.v2.1.1.lof_metrics.by_transcript.txt.bgz">
+              TSV (gnomad.v2.1.1.lof_metrics.by_transcript.txt.bgz)
```
nawatts: Now that there are multiple gene constraint TSVs, I think these should have better labels than "Hail Table" and "TSV".

Maybe
* LoF Metrics by Transcript (Hail Table)
* LoF Metrics by Transcript (TSV)
* LoF Metrics by Gene (TSV)
* LoF Metrics Downsamplings (TSV)

@konradjk

projects/gnomad/src/client/DownloadsPage.js
```diff
@@ -249,14 +249,33 @@ export default () => (
...
 
-    <code>gsutil ls gs://gnomad-public/release/2.1</code>
+    <p>
+      For example, after installing gsutil, start navigating with this command:{' '}
+      <code>gsutil ls gs://gnomad-public/release/2.1.1</code>
+    </p>
+
+    <p>
+      Download all exome VCF files locally, run:{' '}
+      <code>gsutil -m cp gs://gnomad-public/release/2.1.1/vcf/exomes .</code>
```
nawatts: This command won't work. Since exomes is a directory, `cp` would need a `-r` flag.

Also, I don't think we should encourage users to run this, since this will download both the all sites VCF and the per chromosome VCFs, resulting in ~50 GB of duplicate data.

projects/gnomad/src/client/DownloadsPage.js
konradjk: Just say 0.2, that's fine

projects/gnomad/src/client/DownloadsPage.js
konradjk: Yep, sounds good

projects/gnomad/src/client/DownloadsPage.js
```diff
@@ -365,13 +234,24 @@ export default () => (
         <h3>Gene constraint</h3>
         <ul>
           <li>
-            <ExternalLink href="https://console.cloud.google.com/storage/browser/gnomad-public/release/2.1/ht/constraint">
-              Hail Table (gs://gnomad-public/release/2.1/ht/constraint/constraint.ht)
+            <ExternalLink href="https://console.cloud.google.com/storage/gnomad-public/release/2.1.1/constraint/gnomad.v2.1.1.lof_metrics.by_transcript.ht">
```
nawatts: For other Hail tables on the page, links point to the table's parent directory instead of the table directory itself.

For example, https://console.cloud.google.com/storage/browser/gnomad-public/release/2.1.1/ht/exomes instead of gs://gnomad-public/release/2.1.1/ht/exomes/gnomad.exomes.r2.1.1.sites.ht

PR #422, opened 2019-06-18
Configure autism browser ingress/https

mattsolo1:


projects/exome-results-browsers/browsers/asc/deploy/asc-ingress.yaml
```diff
@@ -0,0 +1,35 @@
...
+kind: Ingress
+metadata:
+  name: asc-ingress
+  annotations:
+    ingress.kubernetes.io/ssl-redirect: "true"
+    kubernetes.io/ingress.global-static-ip-name: asc-browser-https
+    ingress.gcp.kubernetes.io/pre-shared-cert: asc-cert
+spec:
+  backend:
+    serviceName: asc-browser
```
nawatts: This service doesn't exist. This should be `asc-browser-nodeport`.

Also, since this ingress exposes only one service, either the `backend` or `rules` configuration could be removed.
https://kubernetes.io/docs/concepts/services-networking/ingress/#the-ingress-resource

PR #534, opened 2019-12-10
Refactor defaultDataset

mhebrard:
I noticed that `defaultDataset` is a variable defined in `routes.js` and in `Searchbox.js`

In this PR, I modified `gnomad/src/client/datasets.js` to be the unique source that define the default dataset, the datasetIDs, the labels , the reference genome and the top level datasets.

* `routes.js` get the default dataset from `datasets.js`
* **SearchBox** module get the default dataset, the top level datasets and the labels from `datasets.js`
* **DatasetSelector** module get the default dataset from `datasets.js`

[Edit]
I also made some modification into **RegionPage** directory to align the Containter and the VariantIn*  with **GenePage**

On a side note I also renamed `GnomadHeading.js` to `Heading.js` for convenience.


nawatts:
Hi @mhebrard. Thanks for the PR.

The changes to make the region page queries consistent with the gene page look good. Those are cherry picked and merged those in 5d452a4b4ff1b1aec686c73ec3e1aa6f497fb1c5.

However, I think the dataset changes introduces too much coupling between the variant datasets metadata and the presentation of the dataset menu and search box components.

mhebrard:
The main goal was to avoid to define the default dataset in 2 different files. But I understand that my chages might not be optimum and introduce some unwanted/unnecessary changes...
Thanks for your review.

PR #543, opened 2020-01-21
Fix for new HGNC header name

shechter:


shechter:
Closed and replaced by #544

PR #548, opened 2020-02-05
population ac_hom equal 0

mhebrard:
I had some issue when a population is absent from the dataset, the API return a null object then the front end could not find the population name of null object.

In the API, if we sett the population feilds to 0, then an object wuth the proper shape is sent, and the frontend can handle a population object properly.

### DIFF
--- a/projects/gnomad-api/src/schema/datasets/gnomad_r2_1/fetchGnomadVariantDetails.js
+++ b/projects/gnomad-api/src/schema/datasets/gnomad_r2_1/fetchGnomadVariantDetails.js
@@ -25,7 +25,7 @@ const formatPopulations = variantData => [
       id: `${popId.toUpperCase()}_${subPopId.toUpperCase()}`,
       ac: (variantData.AC_adj[popId] || {})[subPopId] || 0,
       an: (variantData.AN_adj[popId] || {})[subPopId] || 0,
-      ac_hemi: null,
+      ac_hemi: 0,
       ac_hom: (variantData.nhomalt_adj[popId] || {})[subPopId] || 0,
     })),
     {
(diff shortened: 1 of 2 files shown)


projects/gnomad-api/src/schema/datasets/gnomad_r2_1/fetchGnomadVariantDetails.js
```diff
@@ -25,7 +25,7 @@ const formatPopulations = variantData => [
       id: `${popId.toUpperCase()}_${subPopId.toUpperCase()}`,
       ac: (variantData.AC_adj[popId] || {})[subPopId] || 0,
       an: (variantData.AN_adj[popId] || {})[subPopId] || 0,
-      ac_hemi: null,
+      ac_hemi: 0,
```
nawatts: This should stay `null`. gnomAD v2 doesn't have subpopulations broken down by sex, so the number of hemizygotes is unknown, not necessarily 0.

This is used by the PopulationsTable component to show a "-" instead of a number.
https://github.com/macarthur-lab/gnomadjs/blob/269157843cdbbec9b7e140151819f9be2e3dc85f/projects/gnomad/src/client/VariantPage/PopulationsTable.js#L263

mhebrard: I see. Let me revert that

mhebrard:
Hi @nawatts 

I get your point with `GraphQLNonNull` in the API schema to warn the front-end
The error I was facing was actually on the back-end. If I read the error well, it was due to one population been absent from the data then `ac_hom` was not assagned and the API raised the error

> "message":"Cannot return null for non-nullable field GnomadVariantPopulation.ac_hom."

PR #612, opened 2020-10-09
Set Clinvar as optional for region

mhebrard:
Similar to `VariantsInGene.js`, I set clinvarVariants as optional in `VariantsInRegion.js`, and displayed only if present


nawatts (approved):

nawatts:
Thanks @mhebrard. This just needs one small change to merge: a default value for the `clinvarVariants` prop.

Similar to VariantsInGene.js:
https://github.com/broadinstitute/gnomad-browser/blob/37b70191e15b4b9e0f6c5ca7d4c8a071795c7f13/projects/gnomad/src/client/GenePage/VariantsInGene.js#L97-L99

mhebrard:
Hi @nawatts, I added the default props value.

PR #613, opened 2020-10-09
read data tables on app startup instead of in queries

cseed:
@nawatts I'd be curious to see how this performs.  I have not run it.


nawatts:
Closing this since we're no longer actively experimenting with using Hail as a backend for the browser.

PR #897, opened 2022-04-21
Popmax filtering allele frequency

ch-kr:
This PR adds the populations excluded from popmax faf to the faf description


jkgoodrich (approved):
LGTM, thank you for adding KC!

PR #910, opened 2022-06-22
fix modified time for Google Cloud Storage

phildarnowsky-broad:
@mattsolo1 there's a long comment in this change that explains the rationale behind it, but please let me know if you'd like more detail.

### DIFF
--- a/data-pipeline/src/data_pipeline/pipeline.py
+++ b/data-pipeline/src/data_pipeline/pipeline.py
@@ -23,8 +23,14 @@ class GoogleCloudStorageFileSystem:
         return hl.hadoop_exists(path)
 
     def modified_time(self, path):  # pylint: disable=no-self-use
+        # The Hail docs say that stat["modification_time"] should be a string,
+        # but in the case of Google Cloud Storage, it returns an epoch timestamp
+        # as an int. There is a Google Cloud Storage filesystem in the Hail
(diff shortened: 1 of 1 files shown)


mattsolo1:
I think this looks good. But first need to resolve https://github.com/broadinstitute/gnomad-browser/issues/913 so that CI passes.

mattsolo1 (changes requested):
I think this looks good. But first need to resolve [#913](https://github.com/broadinstitute/gnomad-browser/issues/913) so that CI passes.

phildarnowsky-broad:
@mattsolo1 I rebased this and now black runs, but as you can see [here](https://github.com/broadinstitute/gnomad-browser/runs/7029857211?check_suite_focus=true) CI is now failing because black complains about other files in that directory that this PR doesn't touch.

There are probably two ways to solve this: one would be to open a separate branch, let black reformat everything, and rebase on top of that. Or, assuming black is like other formatters I've used in the past, there's probably a way to have it check only files modified in the PR.

The first approach leaves you with a neater codebase, but also introduces noise into the commit history that can complicate certain tasks later. Do you have a preference?

mattsolo1:
I think I'm fine with the first approach of running the formatter on these files. Can you make that a PR that we can look to see how extensive the formatting changes are?

mattsolo1 (approved):

PR #922, opened 2022-06-28
Add Hail to pipeline requirements.txt

phildarnowsky-broad:
CONTRIBUTING.md called for Hail to be installed manually separately, but it seems to work fine when specified as a dependency in data-pipeline/requirements.txt.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -25,9 +25,7 @@ This should be enough to use the Docker Compose development environment. However
 - For data pipeline development, install [Python](https://www.python.org/), dependencies, and development tools.
 
   ```
-  pip install hail
   pip install -r data-pipeline/requirements.txt
-
   pip install -r requirements-dev.txt
   ```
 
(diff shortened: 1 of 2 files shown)


phildarnowsky-broad:
@mattsolo1 I deleted my Python install, reinstalled everything according to the new directions, and did the first part of the Hail GWAS tutorial as a smoke test. Seems to work fine.

mattsolo1:
If we're going to add hail to `requirements.txt` does it also make sense to include the version? This could be the most recent version that we're confident works properly with the pipeline (which I'm actually not sure of... I guess we could add this later).

mattsolo1 (approved):

phildarnowsky-broad:
> Thing is, I think 0.2.81 is pre-Log4Shell

Whoops...that's a dealbreaker then, unpinned it is. Still, I'll keep my run with 0.2.81 going in the background just to satisfy my own curiosity.

> It's possible the individual datasets have always been updated at different times, and hence with different Hail versions.

Given that, it might be an idea to record Hail versions that are known to work with the different pipelines somewhere or another. It could be as simple as a comment at the head of a file saying "Ran successfully on 2021-02-03 against Hail 0.1.2".

PR #931, opened 2022-07-12
Set up for snapshot testing, add example test

### LINKED ISSUE #930: Create testing suite
Currently, it is difficult to carry out large refactoring efforts while being confident that no external-facing behavior has broken.

@phildarnowsky-broad proposed a snapshot-based approach for ensuring exact HTML/source is generated. Also includes an approach to mocking calls to API/graphql.

phildarnowsky-broad:
resolves #930 

@mattsolo1 @rileyhgrant for this, I used the snapshot capabilities built into the Jest testing library that we're already using: https://jestjs.io/docs/snapshot-testing. 

The snapshot file, `browser/src/__snapshots__/DatasetSelector.spec.jsx.snap`, is checked into version control as I mentioned earlier today. If you read it carefully, you'll see that it's a funny-looking but vaild JS module that exports a bunch of long string literals containing HTML. Behind the scenes, when we call `expect(tree).toMatchSnapshot()`, `toMatchSnapshot` automatically figures out which of those literals goes with the current test, renders the `tree` of React components to HTML, and compares the rendered HTML to the snapshot. If they don't match, the test fails, and the test output will show diffs of the actual versus expected markup to help you diagnose.

When a snapshot test fails, this doesn't always mean there is a bug. Consider that the rendered markup might not match the snapshot due to a deliberate change you made to the code, or due to an unintentional but benign side-effect of such a change. In that case, the thing to do is to update the snapshot by running `yarn test -u`, as opposed to just `yarn test`.

Similarly, when you write a new snapshot test, there won't be a corresponding snapshot in the codebase yet. The first time you actually run the new test, Jest assumes that the result of rendering `tree` gives the correct markup, the new test passes, and it creates the appropriate snapshot. Due to that, when your aim as in this PR is to prevent regressions rather than to cover new features, it's important to run the test and create that "before" snapshot before you start changing the code proper.

Finally, I want to caution one more time against using snapshot tests and only snapshot tests. They're useful, but best when complemented by well-chosen unit tests. In the current situation though, where we're interested in quickly growing a test suite on a complicated codebase from basically nothing, they're an excellent place to start.

### DIFF
--- a/browser/src/DatasetSelector.spec.jsx
+++ b/browser/src/DatasetSelector.spec.jsx
@@ -1,0 +1,18 @@
+import React from 'react'
+import { Router } from 'react-router'
+import renderer from 'react-test-renderer'
+import DatasetSelector from './DatasetSelector'
+import { createBrowserHistory } from 'history'
+
+import { allDatasetIds } from './datasets'
+
+describe.each(allDatasetIds)('DataSelector with "%s" dataset selected', (datasetId) => {
+  test('has no unexpected changes', () => {
+    const tree = renderer.create(
+      <Router history={createBrowserHistory()}>
+        <DatasetSelector selectedDataset={datasetId} datasetOptions={{}} />
+      </Router>
+    )
+    expect(tree).toMatchSnapshot()
+  })
+})
--- a/browser/src/__snapshots__/DatasetSelector.spec.jsx.snap
+++ b/browser/src/__snapshots__/DatasetSelector.spec.jsx.snap
@@ -1,0 +1,5251 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`DataSelector with "exac" dataset selected has no unexpected changes 1`] = `
+<ul
+  className="DatasetSelector__NavigationMenuWrapper-sc-1x96sz-0 dOhyHC"
+>
+  <li
+    className="DatasetSelector__NavigationMenuItem-sc-1x96sz-2 eioFKY"
+  >
+    <a
+      className="DatasetSelector__TopLevelNavigationLink-sc-1x96sz-1-Link hSieFP"
+      data-item="current_short_variant"
+      href="/?dataset=exac"
+      onBlur={[Function]}
+      onClick={[Function]}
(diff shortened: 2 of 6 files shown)


mattsolo1 (approved):
Some things I tried while reviewing:

Added a new dataset to "datasets.js" and then ran `yarn test`. The new snapshot was generated as expected. When I remove this dataset and run `yarn test`, jest tells me there is an obsolete snapshot that I can remove with `yarn test -u`. So that all looks good.

I tried removing the `onBlur` prop from the `DatasetSelector` and see a nice error message.

```
    Snapshot name: `DataSelector with "gnomad_sv_r2_1_non_neuro" dataset selected has no unexpected changes 1`

    - Snapshot  - 2
    + Received  + 0

    @@ -6,11 +6,10 @@
        >
          <a
            className="DatasetSelector__TopLevelNavigationLink-sc-1x96sz-1-Link hSieFP"
            data-item="current_short_variant"
            href="/?dataset=gnomad_r2_1"
    -       onBlur={[Function]}
            onClick={[Function]}
            onKeyDown={[Function]}
          >
            gnomAD v2.1.1
          </a>
    @@ -254,11 +253,10 @@
        >
          <a
            className="DatasetSelector__TopLevelNavigationLink-sc-1x96sz-1-Link hSieFP"
            data-item="current_sv_dataset"
            href="/?dataset=gnomad_sv_r2_1_non_neuro"
    -       onBlur={[Function]}
            onClick={[Function]}
            onKeyDown={[Function]}
          >
            gnomAD SVs v2.1 (non-neuro)
          </a>

      14 |       </Router>
      15 |     )
    > 16 |     expect(tree).toMatchSnapshot()
         |                  ^
      17 |   })
      18 | })
      19 |

      at Object.<anonymous> (browser/src/DatasetSelector.spec.jsx:16:18)

 › 15 snapshots failed.
```

So far this looks great!

rileyhgrant:
> @rileyhgrant can you also give this a try and submit a review before we merge?

Yep, will do.

rileyhgrant (approved):
This all looks good. Excited for the strong start of a robust test suite!

PR #934, opened 2022-07-14
Don't lint snapshots in pre-commit

phildarnowsky-broad:
Technically a snapshot file is valid JS source code, but doesn't and shouldn't adhere to customary JS formatting conventions.


phildarnowsky-broad:
@mattsolo1 small fix to the push I made the other day introducing snapshots, which addresses a problem that came up in my current refactoring work-in-progress.

PR #935, opened 2022-07-14
Refactor isSubset to expedite adding future datasets

phildarnowsky-broad:
@mattsolo1 @rileyhgrant I opened this PR not because the refactoring in it is particularly interesting, but because the tests include some more new things that I'd like to make standard practice.

Looking at those in turn:

The `ReadData` component that I snapshot in this PR needs data from the API to render, but clearly I don't want a real network call in a test suite, so I've mocked it using again the functionality built into Jest: https://jestjs.io/docs/mock-functions. @rileyhgrant not sure if this technique is something you would know so if this sounds new to you let's talk about that in a different thread, it's important and there are subtleties and traps in it.

Furthermore, to write an effective test I need my mock API to respond with data that's up to some minimal standard of realism, and we're dealing with complex object graphs here that are a pain to write out by hand. Enter the Fishery library from the good folks at Thoughtbot: https://github.com/thoughtbot/fishery. Fishery gives you a concise API to write factory functions that spare you from most of the pain of setting up dummy test data, as you can see in `graphql-api/src/graphql/factories.js`.

Speaking of "minimal standard of realism", you may notice that the factories defined here are bare-bones in terms of their contents, and you might think that for realism it would be better to populate these factories with some data copied from production. In practice though it turns out to be best to start with very minimal factories like we have here, that have just enough to them that code using dummy data from these factories won't crash. Once you've made a factory, it's easy to specialize it further, for instance the classic example of an `adminFactory` that is equivalent to a `userFactory` except for setting a superuser flag on the returned data.

Finally, as I've said a couple of times snapshot tests are good but shouldn't be your only tool: the unit tests in `dataset-metadata/metadata.spec.js` are an example of what I have in mind when I say that snapshot tests need something to complement them. Notice too how I've made use of `describe.each` to effectively get 15 tests for the price of one: judicious use of `each` in tests can help cover a lot of ground quickly.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,4 +1,5 @@
 ---
+exclude: '\.snap$'
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v4.0.1
@@ -7,7 +8,7 @@ repos:
       - id: end-of-file-fixer
       - id: trailing-whitespace
   - repo: https://github.com/pre-commit/mirrors-prettier
-    rev: v2.0.5  # This should be kept in sync with the version in package.json
+    rev: v2.0.5 # This should be kept in sync with the version in package.json
     hooks:
       - id: prettier
   - repo: https://github.com/psf/black
--- a/browser/src/ReadData/IGVBrowser.js
+++ b/browser/src/ReadData/IGVBrowser.js
@@ -75,19 +75,21 @@ class IGVBrowser extends Component {
       promisified: true,
     }
 
-    igv.createBrowser(this.el, browserConfig).then(browser => {
-      if (this.mounted === false) {
-        igv.removeBrowser(browser)
-        return
-      }
+    igv
+      .createBrowser(this.el, browserConfig)
+      .then(browser => {
+        if (this.mounted === false) {
+          igv.removeBrowser(browser)
+          return
+        }
 
-      this.browser = browser
+        this.browser = browser
 
-      const resetButton = document.createElement('i')
-      resetButton.className = 'igv-app-icon'
-      resetButton.innerText = '⟲'
-      resetButton.title = 'Reset'
-      resetButton.style.cssText = `
+        const resetButton = document.createElement('i')
+        resetButton.className = 'igv-app-icon'
+        resetButton.innerText = '⟲'
+        resetButton.title = 'Reset'
+        resetButton.style.cssText = `
         position: relative;
         top: -1px;
         font-style: normal;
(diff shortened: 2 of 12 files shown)


mattsolo1 (approved):

rileyhgrant (approved):
Very cool on the creating mocks. I'm not super familiar with them at this moment, one course had us create mocks for unit testing of an application using the MVC architecture - so I'm definitely excited to talk more elsewhere about good practice and patterns when using mocks.

phildarnowsky-broad:
@mattsolo1 thanks for the feedback, for clarity I'll address that one point at a time

> Out of curiosity, why did you decided to put the factories in the graphql folder (even though API is not using the factories)? Could put factories closer to where the component needs them or perhaps directly in the test file? The factory names & data shape seem specific to this component.
>
> Does it make sense to call factories.js something more specific like mock-factories.js? Also if we plan on making many of them, should we start a new folder alongside types and resolvers and store them in appropriately named files?

On reflection I agree that it'd be better to move the factories closer to the test. We could fold them into the test file and that would be fine, except that I anticipate re-using these factories in another test shortly, so I'd just end up pulling them out into their own module anyway. If we later have factories that are indeed used in a single test only, it probably would make sense to have them within the test file itself.

So what's the right place for factories that stand on their own? I think a good approach would be to imitate the convention that the snapshot tests use, that directories with leading and trailing double-underscores in the name are for test use only. That is, for factories used to generate mock data to populate a component in `/path/to/Component.js`, we could put the factories in `/path/to/__factories__/Component.js'.

> If we modify the graphql API schema, I take it we will need to remember to make changes to the factories to match the real data shape? How should we make sure factories stay in sync as the API evolves? Is it worthwhile thinking about routinely testing the frontend/API contract through integration or end-to-end tests, and if would those sorts of tests cover what is being tested here?

This is true, and what's more, given that each factory corresponds to a specific query against the API, which being GraphQL the result can have an arbitrary shape, we'd want to ensure that our factories didn't diverge from the queries either. I did experiment briefly with trying to use our programatically-defined schema to validate factory-created objects automatically upon creation, but that turns out to be less straightforward than you might think. and anyway that wouldn't address the question of staying in sync with the queries, just the schema.

So my suggestion is that we look deeper into this, but not right now, as it's desirable but something we can do without for now that doesn't have a clear solution.

>>    you might think that for realism it would be better to populate these factories with some data copied from production. In practice though it turns out to be best to start with very minimal factories like we have here
>
> Curious to learn more about why this is. 

I realized that while I know this is said to be a best practice I couldn't say why, so I thought about it a while. The conclusion I came to is that a minimal response is, in a sense, generic. This means that, if we invoke a factory like `const almostMinimalFoo = FooFactory({someField: 2})`, where `someField` is an optional field not set on a minimal response, we can say that the only distinguishing characteristic of `almostMinimalFoo` is that `someField` is set, and hence it's obvious that `someField` is critical to the functionality we're trying to test.

> Could another strategy be to snapshot real API responses rather than making factories? Then we could occasionally regenerate the snapshots when needed using the network but most of the time they would be run without accessing the network.

This reminded me that I've worked on a project that did exactly that. My thinking on that is that it would actually be best to use both strategies, that they would complement each other nicely. It looks like there's at least one library that would facilitate this, though of course who knows how well it works until we've tried it: https://github.com/ijpiantanida/talkback.

So I think what I'll do is put the final touches on this PR as it stands and get it merged, then experiment a while with that library, and perhaps see if there are other alternatives. If it proves quick and easy to record a few real API responses and thereby add another layer of assurance to the test suite, great. If not, and it turns out that we would need to dig deeper and/or write our own solution, I'd recommend we hold off on that.

mattsolo1:
Sounds great. Thanks for your responses!

phildarnowsky-broad:
@mattsolo1 regarding recording live responses for later playback, it looks like it will need a nontrivial amount of setup and glue code to make it work, so I'll table that for now but make an issue to remind us to look at it again later.

PR #951, opened 2022-07-25
Add Team page, modify About and Policies pages

rileyhgrant:
Opening a PR for this right now, because the addition and modification of the pages is complete, despite not all the content being present / in its final form.

The listing of contributors, the lack of headshots for many individuals, and the final wording on some of the policies are all pending, and will be changed later, once that info has been finalized. However, I figured it wouldn't hurt to open the PR now, so that if anyone has a spare moment at any point in the week, they can take a look at the PR for code quality and best practices.

I do have a specific question: I'm curious if there's a better practice on using images. I added a `browser/assets/images` directory and am using `<image>` tags with the `src` attribute pointing to that local directory. I did that to keep all the information regarding biography/headshots in one JSON file for ease of mapping over. Is there a better way to go about handling these images?

I'll add comments to a few parts in particular that I'd like opinions on.


browser/src/TeamPage/TeamMembers.json
```diff
@@ -0,0 +1,150 @@
+{
+  "gnomadCouncil": [
+    {
+      "name": "Heidi Rehm, Ph.D., FACMG,",
+      "bio": "(MPI) is an Institute Member at the Broad Institute, Chief Genomics Officer at Massachusetts General Hospital (MGH) and Professor of Pathology at Harvard Medical School (HMS). She brings long-standing experience in genomic resource management, clinical sequencing, variant interpretation and gene discovery for rare disease. She co-chairs the gnomAD council and is responsible for jointly overseeing all activities.",
+      "headshotSource": "../assets/images/hrehm_headshot.jpg"
```
rileyhgrant, line 6: Here I included the relative local filepath as a property, so that it can be used when mapping

browser/src/TeamPage/TeamPage.js
```diff
@@ -0,0 +1,277 @@
...
+
+// Component and Styling for a Card
+// Is there better practice between taking in a single (prop) with fields, or with
+//   taking in the destructured fields of a given prop/object?
+const TeamCard = ({ title, description, headshotSource }) => {
+  return (
+    <TeamHeadshotAndDescription>
+      <Row>
+        <ImageColumn>
+          <Headshot alt="headshot" src={headshotSource} />
```
rileyhgrant, line 73: Here its used in a styled `` tag, with the source attribute. This was the best way I could figure to be able to map over the data and keep this component clean while having the data be easily adjustable. Is there a better way to deal with the headshot images?

mattsolo1:
Looking really good! One thing that jumped out at me was how uneven. These sections look:

image

mattsolo1:
Fixes https://github.com/broadinstitute/gnomad_production/issues/561

sambaxter:
Do we know what the URL will be for the new team page? https://gnomad.broadinstitute.org/team ?

rileyhgrant:
> Do we know what the URL will be for the new team page? https://gnomad.broadinstitute.org/team ?

@sambaxter Yep, that's the intended URL for the teams page. 

- `.../team` is new
- `.../about`'s URL is untouched
-  `.../terms` changed to `.../policies`

mattsolo1:
Looking fabulous @rileyhgrant. Remember to redirect from terms -> polices, in case someone out there is using this URL. Maybe use something like https://v5.reactrouter.com/web/api/Redirect. We are using the /terms route on HomePage so that will need to be changed.

@sambaxter does it make make sense to modify the text "About" page to link out to the new Teams & Policy pages?

rileyhgrant:
> Looking fabulous @rileyhgrant. Remember to redirect from terms -> polices, in case someone out there is using this URL. Maybe use something like https://v5.reactrouter.com/web/api/Redirect. We are using the /terms route on HomePage so that will need to be changed.
> 
> @sambaxter does it make make sense to modify the text "About" page to link out to the new Teams & Policy pages?

Aha! Thanks for referring to that Redirect component. Super useful!

mattsolo1:
Should we convert this from draft -> full PR?

mattsolo1 (approved):
This looks great @rileyhgrant!

Issue #944, opened 2022-07-19
Evaluate autogenerating Typescript types from GraphQL types


mattsolo1:

mattsolo1:
@rileyhgrant I looked back at my previous attempt to generate Typescript from GraphQL & yes I used the same Apollo codegen tool you described

PR #955, opened 2022-07-26
Add apollo-generated types for gnomAD gene and variant page

mattsolo1:
Proof-of-concept approach for https://github.com/broadinstitute/gnomad-browser/issues/944... not to be merged.

To generate types from gnomAD API

`cd browser`

`yarn download-schema`

`yarn generate-types`

### DIFF
--- a/browser/__generated__/globalTypes.ts
+++ b/browser/__generated__/globalTypes.ts
@@ -1,0 +1,32 @@
+/* tslint:disable */
+/* eslint-disable */
+// @generated
+// This file was automatically generated and should not be edited.
+
+//==============================================================
+// START Enums and Input Objects
+//==============================================================
+
+export enum DatasetId {
+  exac = "exac",
+  gnomad_r2_1 = "gnomad_r2_1",
+  gnomad_r2_1_controls = "gnomad_r2_1_controls",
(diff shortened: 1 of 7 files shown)


mattsolo1:
Any thoughts on his this compares to https://github.com/broadinstitute/gnomad-browser/pull/954?

mattsolo1:
I kind of wish the generated interface name was the same as `__typename`. Wonder if there is a way to configure that or if it would lead to name clashes.

For instance, ideally this interface would be called `TranscriptConsequence`, rather than `GnomadVariant_variant_transcript_consequences`

```typescript
export interface GnomadVariant_variant_transcript_consequences {
  __typename: "TranscriptConsequence";
  domains: string[] | null;
  gene_id: string;
  gene_version: string | null;
  gene_symbol: string | null;
  hgvs: string | null;
  hgvsc: string | null;
  hgvsp: string | null;
  is_canonical: boolean | null;
  is_mane_select: boolean | null;
  is_mane_select_version: boolean | null;
  lof: string | null;
  lof_flags: string | null;
  lof_filter: string | null;
  major_consequence: string | null;
  polyphen_prediction: string | null;
  sift_prediction: string | null;
  transcript_id: string;
  transcript_version: string;
}
```

Issue #945, opened 2022-07-19
New branch for migration of gnomAD browser to Typescript


mattsolo1:
Phil is trying this out on a new branch.

mattsolo1:
Closed via #958

PR #958, opened 2022-07-29
Migrate to TS

phildarnowsky-broad:
This contains updates to our configuration, plus light changes to the code itself (mainly consisting of inserting `any` types and `ts-expect-error` directives), to move the `browser` subtree over to TypeScript. It's expected that, if this goes well, we'll convert the other subprojects (e.g. `graphql-api` and friends) in the near future too.

This should definitely get a thorough shakedown from a real end-user before merging, but per the last commit message, here are some basic QA steps I already undertook:

```
    * From main `gnomad-browser` directory, running `yarn format` makes no changes
    * From main `gnomad-browser` directory, `yarn lint:js` and `yarn test` both run with no failures
    * From `browser` subdirectory, `yarn build` builds new bundle
    * From `browser` subdirectory, `yarn start` starts server
    * Modifying source file while dev server is running triggers rebuild and reload
    * With dev server running, can look up BRCA1 gene and click through a number of variant links
```

Notably, one thing I _didn't_ do here is try to correct any significant number of errors flagged by `ts-error-expect`. Generally I feel fine about leaving them in and only removing them by attrition, but looking for a chance to remove some in bulk might be a good side project for the odd free hour.

I apologize in advance for any tricky rebases that you have in your near future.


phildarnowsky-broad:
CI action doesn't seem to have triggered, investigating.

phildarnowsky-broad:
note to self: maybe CI won't run until I resolve the merge conflicts? need to do that sooner or later anyway.

phildarnowsky-broad:
@mattsolo1 I've completed the rebase onto latest `main` and made the necessary adjustments to make CI green again, this is once again ready for review.

phildarnowsky-broad:
@mattsolo1 @rileyhgrant I want to draw your attention particularly to one test utility library I added here, https://github.com/styled-components/jest-styled-components

browser/package.json
```diff
@@ -46,5 +46,10 @@
     "react-window": "^1.7.1",
     "styled-components": "^4.2.0",
     "whatwg-fetch": "^3.5.0"
+  },
+  "devDependencies": {
+    "jest-styled-components": "^6.0.0",
+    "ts-migrate": "^0.1.30",
```
mattsolo1, line 52: Is this a dependency so that we can periodically re-ignore? Should we add an npm script for easily/consistently doing that?

browser/src/AboutPage.tsx
```diff
@@ -2,18 +2,31 @@ import React from 'react'
 import styled from 'styled-components'
 import { PageHeading } from '@gnomad/ui'
 
+// @ts-expect-error TS(2307) FIXME: Cannot find module '../about/about.md' or its corr... Remove this comment to see the full error message
```
mattsolo1, line 5: Wonder why this couldn't be found.

browser/src/CategoryFilterControl.tsx
```diff
@@ -1,6 +1,6 @@
+// @ts-expect-error TS(2307) FIXME: Cannot find module '@fortawesome/fontawesome-free/... Remove this comment to see the full error message
 import Check from '@fortawesome/fontawesome-free/svgs/solid/check.svg'
 import { darken, hideVisually, transparentize } from 'polished'
-import PropTypes from 'prop-types'
```
mattsolo1, line 3: I like how it stripped out PropType imports... or did we do that manually?

browser/src/ClinvarVariantsTrack/ClinvarVariantPropType.ts
```diff
@@ -1,6 +1,33 @@
 import PropTypes from 'prop-types'
 
-const ClinvarVariantPropType = PropTypes.shape({
+type ClinvarVariantPropType = {
```
mattsolo1, line 3: I love that this tool generated these types so nicely. And kudos to @nawatts for disciplined use of PropTypes that enabled this.

browser/src/ClinvarVariantsTrack/ClinvarVariantPropType.ts
```diff
@@ -1,6 +1,33 @@
...
+  in_gnomad: boolean
+  major_consequence?: string
+  pos: number
+  review_status: string
+  transcript_id: string
+  variant_id: string
+}
+
+// @ts-expect-error TS(2322) FIXME: Type 'Requireable<InferProps<{ clinical_significan... Remove this comment to see the full error message
+const ClinvarVariantPropType: PropTypes.Requireable<ClinvarVariantPropType> = PropTypes.shape({
```
mattsolo1, line 30: Sometimes PropTypes are stripped out, sometimes they are not... wonder how this is determined.

browser/src/ConstraintTable/ConstraintTable.tsx
```diff
@@ -90,37 +116,4 @@ const ConstraintTable = ({ datasetId, geneOrTranscript }) => {
   )
 }
 
-ConstraintTable.propTypes = {
```
mattsolo1, line 93: An example of where they PropTypes are removed.

mattsolo1 (commented):
This looks excellent @phildarnowsky-broad! 

I spot-checked a few files but have not gone through all 280 changed files. But most errors look like things we can fix as we go.

Last things to do before merging:

(1) Make sure we can build & deploy in a production setting
(2) I propose we let Riley get his Team/Policy/About page changes in first (which are time-sensitive), then rebase/ts-migrate those files

phildarnowsky-broad:
@mattsolo1 

> I propose we let Riley get his Team/Policy/About page changes in first (which are time-sensitive), then rebase/ts-migrate those files

Yes, I think let's let Riley get it in and then I'll take care of the rebase and migration after, since I've got a good feel for that process now that should be fastest.

browser/package.json
phildarnowsky-broad: The `reignore` command can be helpful, but I also found a lot of manual editing and cleanup to be necessary afterwards, so the payoff of using it is not as high as you might think. I'd suggest that we actually remove `ts-migrate` as a dependency after we merge this PR, and if we think we want it in the future we can always reinstall it on a moment's notice.

browser/src/AboutPage.tsx
phildarnowsky-broad: I think it's to do with the fact that this is a Markdown document rather than TS source, and `tsc` doesn't seem to understand the idea of including assets--at least not as currently configured. If you grep for `Cannot find module`, you'll see references to different sorts of assets: graphics, JSON, fonts, etc., as well as Markdown.

There should be a way to change the configuration to get rid of these warnings, but I left it for now since the build overall seems to work despite the warnings.

browser/src/ClinvarVariantsTrack/ClinvarVariantPropType.ts
phildarnowsky-broad: My only beef with PropTypes really is that they're a half measure, in a hypothetical world where I couldn't have types checked at compile time on everything, types checked at runtime on props only would be better than nothing.

babel.config.js
```diff
@@ -26,11 +26,7 @@ module.exports = {
       },
     ],
   ],
-  plugins: [
-    '@babel/plugin-proposal-class-properties',
-    'react-hot-loader/babel',
```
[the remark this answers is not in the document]
phildarnowsky-broad: It was causing the build to break at some point, I don't recall the exact problem, but since we also have `react-hot-loader` installed and hot-reloading seems to work without `react-hot-loader/babel`, I decided just to ditch the latter as redundant instead of trying to fix the immediate problem.

phildarnowsky-broad:
@mattsolo1 To reiterate, I do think it's important that we have an end-user QA this on a demo server, going through as many reasonable workflows as they can think of, looking for regressions, before we merge this. Since we want to stand up a demo server for smoke testing anyway, we could keep that running and use that. It might be best if one of us also collaborates on a test plan with this hypothetical QA tester, in the interest of finding more edge cases to prod than either might find alone.

mattsolo1:
fixes #945

mattsolo1 (approved):

PR #963, opened 2022-08-04
Allow use of interactive debugger in tests

phildarnowsky-broad:
Chrome has an interactive Node debugger built in, see https://jestjs.io/docs/troubleshooting for instructions on how to access it. The script added here runs the test suite under that debugger.

See https://jestjs.io/docs/troubleshooting.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,8 @@
     "format": "prettier --write './**/*.js'",
     "lint:css": "stylelint './**/*.js'",
     "lint:js": "eslint .",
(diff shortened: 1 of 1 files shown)


phildarnowsky-broad:
FYI @mattsolo1 this will now be possible and is very useful, I could also add a script to run the browser code itself with the debugger if we want.

rileyhgrant (approved):
Nice, this seems like it'll be quite useful.

PR #966, opened 2022-08-09
Add Cypress and sample basic tests to project

### LINKED ISSUE #941: Evaluate Cypress in our testing suite
See discussion: https://the-tgg.slack.com/archives/C03P7FA3W3T/p1658156228150599

(1) Make assertions about specific parts of DOM against part the real browser actually rendered.

OR 

(2) take real screenshot (e.g. Percy).

rileyhgrant:
Add [Cypress](https://www.cypress.io/) to the project to be used for e2e testing.

Add sample basic tests that right now only check basic navigation of home page, in `home_page.cy.ts`

To see tests, open both a local development server and Cypress:

In one shell

```
$ yarn start
```

In a different shell

```
$ yarn cypress:open
```

Resolves #941


rileyhgrant:
Opening as a draft right now, as part of the issue is to evaluate whether we want to be using Cypress.

browser/cypress/e2e/home_page.cy.ts
```diff
@@ -0,0 +1,102 @@
...
+        pageUrl: '/publications',
+        pageHeader: 'publications by the gnomAD group',
+      },
+      {
+        pageTitle: 'Feedback',
+        pageUrl: '/feedback',
+        pageHeader: 'Tell us how you use gnomAD',
+      },
+      {
+        pageTitle: 'Help',
```
mattsolo1, line 51: Can confirm that the Cypress test that checks for a properly rendered `pageHeader` at `/help` is failing... which is the true state of the world. Very cool!

image

mattsolo1 (commented):
I think this example is already useful to merge as-is. Would be interested to hear feedback from @phildarnowsky-broad. 

Also, how do we make sure this is run as part of CI/Github Actions?

rileyhgrant:
> I think this example is already useful to merge as-is. Would be interested to hear feedback from @phildarnowsky-broad.
> 
> Also, how do we make sure this is run as part of CI/Github Actions?

I did not yet add anything to the CI to run the Cypress tests, if we do want to use Cypress I was going to read up and get them running as part of our existing CI, fortunately their docs are pretty helpful and thorough, and they have a [section dedicated to exactly this](https://docs.cypress.io/guides/continuous-integration/github-actions)

mattsolo1:
Nice, CI working! This is awesome.

browser/cypress/support/commands.ts
```diff
@@ -0,0 +1,37 @@
+/// <reference types="cypress" />
```
mattsolo1, line 1: Is this file boilerplate that we can remove?

browser/cypress/fixtures/example.json
```diff
@@ -0,0 +1,5 @@
+{
```
mattsolo1, line 1: Is this file boilerplate that we can remove?

browser/cypress/support/e2e.ts
```diff
@@ -0,0 +1,20 @@
+// ***********************************************************
```
mattsolo1, line 1: Is this file boilerplate that we can remove?

package.json
```diff
@@ -10,7 +10,9 @@
     "lint:css": "stylelint './**/src/**/*.(js|ts|jsx|tsx)'",
     "lint:js": "eslint .",
     "test": "jest",
-    "test:debug": "node --inspect-brk node_modules/.bin/jest --runInBand"
+    "test:cypress": "cd browser && cypress run",
+    "test:debug": "node --inspect-brk node_modules/.bin/jest --runInBand",
```
mattsolo1, line 14: Maybe calling this `test:jest` is more informative

mattsolo1: Actually... what's the purpose of `test:debug`?

browser/cypress/fixtures/example.json
rileyhgrant: Good catch this one we can remove

browser/cypress/support/commands.ts
rileyhgrant: Removing this broke Cypress, so despite the large amount of comments in it, it is needed. We could remove all the documentation on it, but honestly it might be useful in the future if it's needed to dive into creating custom commands - just to have it in this location?

browser/cypress/support/e2e.ts
rileyhgrant: This can be removed if a certain flag it set, but I think it'd be better to keep this to have the option open to modify behavior of Cypress in the future if needed.

package.json
phildarnowsky-broad: I added that the other day, it lets you use an interactive debugger with tests. See https://jestjs.io/docs/26.x/troubleshooting#tests-are-failing-and-you-dont-know-why

rileyhgrant: Good question `test:debug`, upon brief research, appears to be a script using Jest's [recommended way](https://jestjs.io/docs/troubleshooting) of getting more info out of a failing test by running it in a Node process that allows use of `debugger` statements. 

As such, it's now been renamed `test:jest:debug`

phildarnowsky-broad: Also as long as we're doing this, perhaps we want a task something like `test:all` or `test:full` that runs Jest, Cypress, and whatever else we might later add?

browser/cypress/support/commands.ts
phildarnowsky-broad: My guess is, this file would be the established place to put such custom commands as you might have, and something in Cypress is doing `import * from 'support/commands'` or similar. In which case we'd be stuck with it, and I'd favor keeping the content as a handy quick reference because my estimation is we'll find a use for these custom commands if we stick with Cypress.

browser/cypress.config.ts
```diff
@@ -0,0 +1,11 @@
+// eslint-disable-next-line import/no-extraneous-dependencies
```
phildarnowsky-broad, line 1: I'm surprised that TS thinks `defineConfig` is "extraneous", seeing as we use it right below

browser/cypress/support/e2e.ts
```diff
@@ -0,0 +1,20 @@
...
+// You can change the location of this file or turn off
+// automatically serving support files with the
+// 'supportFile' configuration option.
+//
+// You can read more here:
+// https://on.cypress.io/configuration
+// ***********************************************************
+
+// Import commands.js using ES2015 syntax:
+import './commands'
```
phildarnowsky-broad, line 17: oh hey I found where it does `import './commands'`

browser/cypress/e2e/home_page.cy.ts
```diff
@@ -0,0 +1,108 @@
...
+  it('Home page loads successfully', () => {
+    cy.visit('/')
+
+    // Make sure the page has the landing splash
+    cy.contains('Genome Aggregation Database')
+    cy.contains('gnomAD browser')
+
+    // Make sure page content headers are there
+    cy.contains('New to gnomAD?')
+    cy.contains('About')
```
phildarnowsky-broad, line 14: Looks good, but perhaps we could interact with the version selector here? Consider that that's going to change soon, inasmuch as v4 will shortly be an option.

browser/cypress/e2e/home_page.cy.ts
```diff
@@ -0,0 +1,108 @@
...
+  })
+
+  // json for each header page for ease of iterating for tests
```
phildarnowsky-broad, line 17: good call

package.json
rileyhgrant: > Also as long as we're doing this, perhaps we want a task something like `test:all` or `test:full` that runs Jest, Cypress, and whatever else we might later add?

Good call, I've added `test:full` per your recommendation

browser/cypress/e2e/home_page.cy.ts
```diff
@@ -0,0 +1,108 @@
...
+      cy.contains('gnomAD browser')
+      cy.contains(route.pageHeader)
+      cy.url().should('include', route.pageUrl)
+    })
+  })
+
+  // TODO: stub these? so that you see that the page properly loads with dummy data
+  //   However, it'd be good to use real data from the API for well understood
+  //   happy paths so the strategy could be:
+  //   - Stub in general? To not query real API
```
phildarnowsky-broad, line 78: Definitely let's stub here, as a rule hitting a remote service in a test suite is setting yourself up for trouble. As a rule anything that makes tests fail intermittently habituates people to test failures, and getting habituated to test failures means missing meaningful test failures.

I can see the argument for going to the trouble for a handful of top-priority happy paths, but I'd also say those are the exact paths that should be tested most thoroughly at the lower levels too.

phildarnowsky-broad (changes requested):

mattsolo1 (approved):

phildarnowsky-broad (changes requested):
@rileyhgrant code itself looks impeccable, the last thing I'd suggest you do is a little bit of rebase trickery. Specifically, I would rebase so as to fold the stubbing of API interaction (https://github.com/broadinstitute/gnomad-browser/pull/966/commits/6fa7c941de797fa7baebf9275e9c9863aaa33cea) into the commit that introduces the pertinent test.

Otherwise I think we're in business! I pulled this and played with Cypress myself a bit, the possibilities are exciting.

phildarnowsky-broad:
@rileyhgrant it just occurred to me that, if you want to give a talk sometime, a brief introduction to Cypress and E2E testing in general might make a decent topic.

phildarnowsky-broad:
@rileyhgrant sorry, just noticed one more small thing, another rebase. Specifically, let's take that very last commit [Add explicit Jest imports](https://github.com/broadinstitute/gnomad-browser/pull/966/commits/10a724aacec727408bad9a803e3553ea897be267) and rebase to fold it into the first commit, where you introduce Cypress. 

Then please go through this branch commit by commit (there should be 4 remaining at this point), and verify that the Jest and Cypress tests, as they are in each respective commit, all pass. Probably good to try `yarn build` in the `browser` dir as well, even though there shouldn't be any changes that affect that.

Generally, the ideal is for each commit to be clean, meaning that if we choose any arbitrary commit out of the history and check that out in our working copy, we should be able to run all tests and build the whole project, as they exist at that point in time.

As a corollary, I'd suggest that we adopt as SOP the idea that, when you open a PR, and that PR introduces a bug in commit A, which is later fixed in commit B in the same PR, once you feel confident in B, that you rebase in such a way that B gets merged into A, replacing them both with a new clean commit A'.

One trick that can help you rebase with confidence is to make a temporary backup branch. Say you have a git history that looks like this:

```
123456 did a thing
234567 did second thing
345678 whoops first thing was wrong, fixed here
```

The desired end state is then:

```
121212 did a thing (any rumor that I ever did this thing wrong is just a rumor)
232323 did second thing
```

Now in this case it's simple because you `git rebase` to move `345678` to directly after `123456`, then `fixup` so they are merged together into the commit we call `121212` here. But if you've got a lot of changes, say after a thorough code review, it can be much trickier, and if you have to resolve conflicts manually you might wonder if you did so correctly.

The trick hinges on this fact: in terms of the actual _content_, commit `345678` and `232323` should be totally indistinguishable. This means that you can use your pre-rebase version, `345678` here, as a reference to compare the post-rebase version to. You simply do something like:

```
$ git checkout -b topic_branch_name.BAK1
$ git checkout topic_branch_name
```

...then make your changes on branch `topic_branch_name`, and when you're done, just `git diff topic_branch_name topic_branch_name.BAK1`. If that comes up with anything, you know that you goofed somewhere.

rileyhgrant:
> @rileyhgrant sorry, just noticed one more small thing, another rebase. Specifically, let's take that very last commit ...

Aha, ok very cool. Thanks for this detailed and helpful explanation, particularly my working definition of a clean commit history was faulty (Clean commit history being a commit history that the history itself is clean in a legible and logical way, vs a history of commits that are each individually 'clean' in the way you described above). 

This is super helpful, makes a ton of sense, and I'm definitely on board to adopt the principle you mentioned, of rebasing and squashing bug fixes into commits that introduced the bug itself.

Exciting! And thanks for all the guidance, as per usual.

rileyhgrant:
> @rileyhgrant sorry, just noticed one more small thing, another rebase. Specifically, let's take that very last commit ...

@phildarnowsky-broad Alrighty, all commits are clean and the history has been modified to keep clean commits and keep relevant code changes in the expected commit.

This also now includes a very minor fix for a `@ts-expect-error` message that was being printed as text on the Team Page

Thanks for your review and comments.

phildarnowsky-broad (approved):
🎉

PR #968, opened 2022-08-09
Fix help page

rileyhgrant:
Update `buildHelp.js` so that it correctly imports `.tsx` files to be used in the FAQ.

Add unrelated edit to `eslintrc.js` to allow JSX syntax in `.ts` and `.tsx` files

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -8,7 +8,7 @@ module.exports = {
   rules: {
     'prettier/prettier': 'error',
     'func-names': ['warn', 'as-needed'],
-    'react/jsx-filename-extension': ['error', { extensions: ['.js'] }],
+    'react/jsx-filename-extension': ['error', { extensions: ['.js', '.jsx', 'ts', '.tsx'] }],
     'import/prefer-default-export': 0,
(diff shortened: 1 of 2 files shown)


.eslintrc.js
```diff
@@ -8,7 +8,7 @@ module.exports = {
   rules: {
     'prettier/prettier': 'error',
     'func-names': ['warn', 'as-needed'],
-    'react/jsx-filename-extension': ['error', { extensions: ['.js'] }],
+    'react/jsx-filename-extension': ['error', { extensions: ['.js', '.jsx', 'ts', '.tsx'] }],
```
rileyhgrant, line 11: As a question: is this fair game to have as its own commit?

I made this change as part of this PR, and as it's own individual commit as its distinct from the spirit of fixing the loading of the help page.

Is this better as it's own commit as part of this PR, or as a separate small PR.

phildarnowsky-broad: When it comes to small stuff like this, I generally like to have it all in the same PR if it's even tangentially related to the main point of the PR. If I spot a bug that's totally unrelated, but very small, I'll even add a commit to the PR to fix that separate bug and just note in the PR description "Contains an unrelated fix for such-and-such". In the end, PRs are transitory, it's the commit history that lasts.

phildarnowsky-broad (approved):

PR #970, opened 2022-08-10
Modify header routes and order

rileyhgrant:
Resolves broadinstitute/gnomad-blog#76

Modify the header routes so `Changelog` is to the left of `News`, keeping the two blog/news related pages together in the header.

Modifies the routes of `Changelog` and `News` to point to the live site (`https://gnomad.broadinstitute.org/news/` rather than `/news/`), allowing for one-way correct navigation in dev environments and demo deployments, rather than relying on the production site re-direct alone.

Unrelated - correct a mistake in the acroynm for a current funding source (NIHCHD -> NICHD)


mattsolo1 (changes requested):
Same feedback from https://github.com/broadinstitute/gnomad-blog/pull/77

rileyhgrant:
Order is now [..., `News`, `Changelog`, ...]

mattsolo1 (approved):

Issue #969, opened 2022-08-09
Age distribution FAQ question does not display Histograms


rileyhgrant:
### What you did:
Navigate to the question `What is the age distribution of in gnomAD` on the `help` page.

This bug currently happens on the live production gnomAD browser, in addition to in the current state of the `main` branch.

### What you expected to see after you did that:
Based on the relevant imported file, there should be a histogram that uses the gnomAD metadata to display the age distribution for different versions.

### What you actually saw after you did that:
The headers and paragraphs load, and there is a wrapper and svg element, but the element itself is blank.

* Screenshot
![age_distribution](https://user-images.githubusercontent.com/59549713/183766338-69abab6e-a394-46f3-97cd-3ec8e29bdb4d.jpg)

rileyhgrant:
Will look at this more tomorrow, currently trying to trace when this happened to see what might have caused the issue.

The history of the [faq question](https://github.com/broadinstitute/gnomad-browser/commits/babd364256df5093b5ca7b05b9056a29eeb785ea/browser/help/faq/technical-details/what-is-the-age-distribution-in-gnomad.js?browsing_rename_history=true&new_path=browser/help/faq/technical-details/what-is-the-age-distribution-in-gnomad.tsx&original_branch=main) itself has not been touched for a while, nor has the [Histogram Component](https://github.com/broadinstitute/gnomad-browser/commits/babd364256df5093b5ca7b05b9056a29eeb785ea/browser/src/Histogram.js?browsing_rename_history=true&new_path=browser/src/Histogram.tsx&original_branch=main). This bug is currently in production, so it does not seem to be a related to the `ts-migrate` branch.

phildarnowsky-broad:
@rileyhgrant 
> This bug is currently in production, so it does not seem to be related to the ts-migrate branch, as production does not have this PR merged.

Yup--you anticipated my first question. :)

phildarnowsky-broad:
FYI right now I am using `git bisect` to try and pinpoint the problem. There's still a bit to go, but so far I can tell that this has been broken since at least June 7, 2021--that's more than 14 months. Another example of the value of tests.

phildarnowsky-broad:
@rileyhgrant For your information, courtesy of `git bisect`. Hope this helps, please let me know if you want more help with this.

```
1d1eec15fae944c8daf272a26cb2b688417535bb is the first bad commit
commit 1d1eec15fae944c8daf272a26cb2b688417535bb
[...]

PR #975, opened 2022-08-24
Convert existing tests and buildHelp.js script to TypeScript

phildarnowsky-broad:
Here, we convert our backlog of existing tests in JS to TypeScript. We also convert `buildHelp.js` in the same commit, which is a thing that I would normally call out myself on the theory that converting that script would merit its own separate commit. However, there is a dependency in there that's probably not readily obvious, at least, it sure took me a while to figure out why changing the tests over to TS broke the apparently unrelated script. I can go into more detail if you want about that, but it's not all that interesting.


phildarnowsky-broad:
TL;DR: the code in `buildHelp.js` that generated the list of JS/TS (as opposed to Markdown) files to use was erroneously including tests, i.e. `spec.tsx` files, which of course never came up before since there weren't any tests. I'm not 100% clear on the details, but it seems that whereas JS accepted this uncritically, TS has higher standards. Fixing that was as simple as throwing in a `filter` that matched all files except for tests. Everything else I did in that file was just a bit of refactoring, and adjusting a few things that the TS linter didn't like.

rileyhgrant (approved):
Looks good to me!

phildarnowsky-broad:
@rileyhgrant if you haven't yet, could you please pull this branch and manually verify that the build script is giving good results? Given that this (apparently obvious) script has been trouble a couple of times in a brief period already, I'd feel better if you double-checked this one for me on your own machine.

phildarnowsky-broad:
@rileyhgrant specifically, I'd appreciate if you could test both `yarn build` and `yarn start` and verify by eye that the help pages, especially the haplotype ones that you modified recently, haven't regressed

rileyhgrant:
> @rileyhgrant if you haven't yet, could you please pull this branch and manually verify that the build script is giving good results? Given that this (apparently obvious) script has been trouble a couple of times in a brief period already, I'd feel better if you double-checked this one for me on your own machine.

@phildarnowsky-broad Oops yeah, I should have clarified when I approved that things are working on my machine. Good points on wanting to be especially sure seeing as its this script. 

Both running `yarn start`, as well as `yarn build` then checking the `npx live-server` from the `gnomad-browser/browser/dist/public` folder result in a functional help page on my machine. The `.tsx` files that get used as part of the build, and were causing issues for a tiny bit, are working, as are the `.md` files. The exception of the one noted in issue #969, which is not directly related to the `buildHelp.ts` file, and has a standalone issue filed.

phildarnowsky-broad:
@rileyhgrant great, thanks--I figured you probably had checked these on your own initiative, but I was just looking over this PR and saw that, while I had meant to ask you explicitly to do so at the time I opened the PR, apparently I forgot to do so. anyway, good to hear that we're in business, I'll merge.

PR #976, opened 2022-09-06
New test helpers

phildarnowsky-broad:
This introduces two new test helpers that I spun off from another branch I'm working on. They're useful in their own right, so no reason to hold them back from `main` until the other branch is complete.

Those helpers are:

* `withDummyRouter`: many of our components need to be wrapped in a `React.Router`, this streamlines the process of doing so in a test
* `forAllDatasets`: a bit of syntactic sugar over `import { allDataSetIds } ; ... ; describe.each(allDatasetIds)`

I will probably have more spinoff PRs with more test helpers to follow.


browser/src/DatasetSelector.spec.tsx
```diff
@@ -1,20 +1,17 @@
...
 import renderer from 'react-test-renderer'
-import { createBrowserHistory } from 'history'
 import DatasetSelector from './DatasetSelector'
 
-import { allDatasetIds } from './datasets'
+import { forAllDatasets } from '../../tests/__helpers__/datasets'
+import { withDummyRouter } from '../../tests/__helpers__/router'
 
-describe.each(allDatasetIds)('DataSelector with "%s" dataset selected', (datasetId: any) => {
+forAllDatasets('DataSelector with "%s" dataset selected', (datasetId) => {
```
rileyhgrant, line 11: Neat - the helper make it concise and legible

dataset-metadata/metadata.ts
```diff
@@ -1,4 +1,24 @@
-import { datasetLabels } from '../browser/src/datasets.ts'
```
rileyhgrant, line 1: Good call changing this to be more clearly named.

rileyhgrant (approved):

dataset-metadata/metadata.ts
phildarnowsky-broad: part of my overall scheme for the branch that this was spun off from, is to dismantle `browser/src/datasets.ts` and move the usable parts to this new `dataset-metadata` project.

PR #977, opened 2022-09-06
Ensure types checked

phildarnowsky-broad:
Due to a mistake in the Webpack config, TypeScript code was being compiled properly, but no actual type checking was done. Corrected this, and made necessary changes to the code accordingly.


phildarnowsky-broad:
@rileyhgrant just for my own peace of mind, could you please definitely make sure to try this locally? specifically, try introducing a type error (dealer's choice, LMK if you have trouble coming up with something), and make sure that both `yarn build` from the `browser` subdirectory, and `yarn jest` from the main directory, both fail with appropriate type erros.

browser/help/faq/mitochondrial-dna/mtdnaHaplogroupPerNuclearAncestryPopulationData.ts
```diff
@@ -0,0 +1,247 @@
...
+  'T',
+  'U',
+  'V',
+  'W',
+  'X',
+  'Y',
+  'Z',
+] as const
+
+export type Haplogroup = typeof haplogroups[number]
```
rileyhgrant, line 33: aha, this was what caused you to make that post in the SoftEng slack channel about this topic. Interesting.

phildarnowsky-broad: Yeah, it bothered me that I apparently had to repeat myself at the type level and at the value level, so I dug a little and found out this little gem. Typescript's type system is a big and sophisticated one so don't worry about trying to take it all in at once.

rileyhgrant (commented):
Commenting and not explicitly approving as I'm a bit confused.

Introducing a type error does indeed cause both `yarn build` from the `browser` subdirectory, and `yarn jest` from the top level directory to fail with appropriate type error messages.

With no type errors, `yarn jest` performs as expected, but `yarn build` produces 60 errors of the `[tsl] TS1005: ':' expected.` or `TS1003: Identifier expected.` flavor on my machine. Despite this, I can still navigate the browser with `yarn start` or `npx live-server` from the `dist/public` output directory. 

Is there some linting still happening after the transpilation to javascript? / from a curiosity point of view, why does this error that gets flagged during compilation and results in `webpack compiled in 60 errors` not actually causing anything to break (as in, I can still visit localhost and see and use the browser)? Ignoring these errors seems to me like a case of getting devs acclimated to ignoring errors, which could be a bad thing. See attached screenshot for a few of the errors.

compilation_errors

In general, jest is working as expected, with type errors causing a test to fail as expected with an appropriate error message referencing the type mismatch. `yarn build` on my machine is producing many errors, one of which is a one regarding a type mismatch if there is one.

phildarnowsky-broad:
@rileyhgrant interesting--note that the name of the source JS file it's complaining about appears to be autogenerated, and the fact that it's under the `dist/public` directory. Looks like the problem here is, for some reason, Webpack is trying to include the result of a previous compilation as a source for the current compilation.

What happens if you `rm -rf` that `dist/public` directory and try the build again? (Or you could `mv` that directory elsewhere if `rm`ing it makes you nervous.)

phildarnowsky-broad:
@rileyhgrant as for the counterintuitive behavior regarding the site still working after a build failure, this is something I've also gotten caught on many times: you're looking at the _last_ successful build of the site. That does mean that, if you create a type error, then don't run tests before checking in your code, it's possible to push code with a type error to Github.

That's not catastrophic, because CI will catch that when it tries to build the bundle from scratch, and it's infinitely better to have CI catch that than to let it slip by. But it would also be better to catch that earlier in the pipeline. I wonder if we could rig up some kind of widget on the page itself that only displays in dev mode, something very prominent in a deliberately garish color scheme, with a warning to the effect of `YOU ARE LOOKING AT A STALE BUILD. 6 build failures since 2022-09-13 09:25.`

phildarnowsky-broad:
I thought about the possibility of a pre-commit hook that ensures that the source builds from a clean slate, but the problem is that that could turn very time-consuming. I wonder though if there's a way to just typecheck everything without actually having to compile everything, that should be much faster if possible, and while it won't catch every bug it should catch a great many of them.

rileyhgrant:
Aha, thanks for so thoroughly answering my questions. You were, of course, correct on the errors on my machine coming from some previously built files in the `dist/public` that were still hanging around. As an aside, is removing the contents of that directory something I should do regularly? Or is this a particular case in which there were older files still hanging around and this is the first time TypeScript was properly trying to compile things to JavaScript with type checking?

Removing the contents of that directory causes the intended behavior when introducing a type error, and you already touched upon the interesting bit of falling back on the previous successful build (thanks for the clarification). Interesting idea on both of the fronts re: stale build warning for the site in dev mode, or just checking types without compilation. For my two cents, I feel like a stale build warning message on the site would be sufficient? Happy to discuss more offline and hear more about peoples opinions.

Anyhoo, thanks for answering my questions. Looks good!

rileyhgrant (approved):

phildarnowsky-broad:
@rileyhgrant no need to delete `dist/public` regularly, no, it's only something you need to do if like in this case you think it's specifically implicated.

I'll make an issue about giving a stale-build situation a higher profile generally, feels like we can wait to move on that one until we all three have a chance to talk it over.

phildarnowsky-broad:
> That's not catastrophic, because CI will catch that when it tries to build the bundle from scratch

Well, never mind, because it turns out CI will only catch such a bug _if it's in a file that some test exercises_, which is almost none of the codebase as it currently stands. So it looks like we'll need to add in an explicit test that the code compiles, I'm thinking now I will add a CI step for that to this PR. More commits to follow.

phildarnowsky-broad:
@rileyhgrant as I mentioned above, I added a new commit with a step that builds the whole project. I verified that a type error that got past jest was caught by the new step, in run https://github.com/broadinstitute/gnomad-browser/actions/runs/3045979908/jobs/4908216219.

rileyhgrant:
> @rileyhgrant as I mentioned above, I added a new commit with a step that builds the whole project. I verified that a type error that got past jest was caught by the new step, in run https://github.com/broadinstitute/gnomad-browser/actions/runs/3045979908/jobs/4908216219.

That it does! Very nice.

PR #980, opened 2022-09-14
Centralized dataset metadata part 1

### LINKED ISSUE #853: Update components that render conditional on dataset (part 1)
Many components render differently based on the current dataset. Check for conditionals such as `datasetId.startsWith('gnomad_r3')`.

Ideally, instead of checking the dataset ID directly, abstract checks into functions in [dataset-metadata](https://github.com/broadinstitute/gnomad-browser/tree/main/dataset-metadata). For example, `datasetHasGenomes` vs `datasetId !== 'exac'`.

phildarnowsky-broad:
In this PR, we begin to draw the metadata for our various datasets together into a centralized, strongly typed, test-covered library.

This branch addresses https://github.com/broadinstitute/gnomad-browser/issues/853. There are still many more changes that I intend to make related to that ticket. I decided, however, to get this first part into review and get feedback on it while continuing to move forward on those changes (hence "part 1"). I've made a separate issue, https://github.com/broadinstitute/gnomad-browser/issues/981, to track work on the remaining changes.

There are a couple of reasons I thought breaking up the work like this would be advantageous:

* It's already a big changeset, and there's still much more to come, so I'm concerned about it getting unwieldy for purposes of review, rebasing, etc.
* At this point I think I've done most or all of the groundwork, and the rest should be straightforward if bulky. If review turns up a fundamental problem with the changes I've made here, better to find that out now than wait until all the other changes are done.

Resolves #853


phildarnowsky-broad:
@rileyhgrant you'll likely find it much easier to review this one commit by commit, rather than as a single diff.

phildarnowsky-broad:
@rileyhgrant can you please review ASAP? I also owe you a couple of reviews, I will try and knock them out today on the theory that it's a good use of between-meeting time.

rileyhgrant:
> @rileyhgrant can you please review ASAP? I also owe you a couple of reviews, I will try and knock them out today on the theory that it's a good use of between-meeting time.

Oop, yeah will do. My bad for letting this one hang out unreviewed for quite a while.

browser/src/ConstraintTable/ConstraintTable.spec.tsx
```diff
@@ -0,0 +1,109 @@
...
+          geneOrTranscript={transcriptFactory.build({ chrom: 'M' })}
+        />
+      )
+      expect(tree).toMatchSnapshot()
+    })
+  })
+})
+
+test('ConstraintTable with exac dataset and gene with available constraints has no unexpected changes', () => {
+  const tree = createShallowRenderer().render(
```
rileyhgrant, line 66: More just curiosity - what's the reasoning behind when to use `createShallowRenderer()`? It looks like the exac table has less other components in it, and the use case seems to be for testing of components in isolation, regardless of how many other components they are built up of? So I would've expected `createShallowRenderer()` to be used either everywhere, or more heavily for datasets that aren't `exac`. I may just ask in zoom, but am curious.

phildarnowsky-broad: This might just be a mistake on my part, I'm looking into that.

rileyhgrant (changes requested):
The code itself looks good and I am very on board to centralize the metadata and get it thoroughly tested.

Requesting changes because of a bug I'm seeing on my machine that first occurs in (oddly) the commit:
  `introduce hasShortVariants metadata field`
  `2108b7f55f035060f9c59874eddc299d2a195c52`

Starting at this commit, the DatasetSelector begins to not display the currently selected dataset on the left pane, and selecting a dataset on the right pane updates the text in the left pane.

See screenshot below, in that screenshot I have selected an option from the right pane (it's in grey), and its showing the same 'SV' label on the left pane.

I was poking around trying to figure out why this was occurring on my own for a bit, and am happy to continue also, but figured you are closer to these changes and files. I went through commit by commit, and am only noticing this behavior starting with the commit mentioned above. 


interesting_behavior

rileyhgrant:
Here's another screenshot elaborating on the behavior

On the header, we can see the selected dataset is `v3.1.2`, and can confirm this by seeing the coverage track is all green (no exome data). However, the two-paned dataset selector did not update when the page reloaded, and the left pane is still showing `v2.1.1`




odd_behavior_2

phildarnowsky-broad:
@rileyhgrant figured out the problem regarding short variants, added a commit to deal with it, and verified that the bad behavior you flagged is gone. Nice catch there.

Notice that I added the commit to fix that at the end of the chain of commits, rather than `rebase`-ing it after the commit it fixes, in apparent contradiction of a practice I described the other day. However, starting the commit message with the `fixup: ` prefix as I did here is a convention that indicates that, after the PR is approved, I intend to first `rebase` the new commit into the obvious place, resolve any conflicts, then `rebase` again to `fixup` that new commit into the one that fixes. One advantage of that is that you don't have to do any force-pushes until you're almost at the end of the process, which tends to make the review process simpler.

browser/src/ConstraintTable/ConstraintTable.spec.tsx
phildarnowsky-broad: @rileyhgrant `createShallowRenderer` is in fact not appropriate here, and in moving away from that, I saw another chunk of refactoring that I want to include in this PR, so watch this space for a re-review request.

The basic difference between shallow and regular rendering is that regular rendering recursively renders all children of the rendered node, until it doesn't have any React components in it, just HTML elements, text, other raw elements of a Web page. By contrast, shallow rendering doesn't render the children: instead of rendering a child, it includes a kind of dummy node indicating what child was called and with what params.

One common use case for shallow rendering would be something like this: say you have a `List` widget that contains some number of children, each of which is a heavyweight React component in its own right, with maybe a bunch of network calls that have to be stubbed, test data that has to be set just right, and other complications to the test setup. Suppose you're writing a test to verify that, if you render a `List` with the right arguments, you get a certain sequence of child widgets--the content itself of the widgets is of no interest to you in this one particular test, you just want to know that each is the right kind called with the correct set of arguments. Here shallow rendering lets you test what you're concerned about without having to do a lot of extra setup just to ensure that the parts you don't really care about don't fail the test.

rileyhgrant (approved):

PR #983, opened 2022-09-21
Update clinvar.py to allow ingesting of ClinVar data

### LINKED ISSUE #961: Latest Clinvar release cannot be parsed
```
Java stack trace:
org.apache.spark.SparkException: Job aborted due to stage failure: Task 1434 in stage 0.0 failed 20 times, most recent failure: Lost task 1434.20 in stage 0.0 (TID 2070) (vep85-sw-q2n2.c.gnomadev.internal executor 63): is.hail.utils.HailException: Caught exception while reading hdfs://vep85-m/tmp/clinvar_variants.tsv: is.hail.utils.HailException: could not convert "Un:113557" to locus<GRCh37> in column "locus_GRCh37" 
  offending line: Un:113557	["T", "A"]	chr4:190174726	["T", "A"]	{"clinvar_var...
```

```
Hail version: 0.2.81-edeb70bc789c
Error summary: HailException: Invalid locus 'Un:113557' found. Contig 'Un' is not in the reference genome 'GRCh37'.

[Stage 0:====================================================>(1999 + 6) / 2000]
```

rileyhgrant:
Add a fix to `clinvar.py` that allows the pipeline to be run despite two anomalous variants with Chromosome "Un" that were included in ClinVar releases since at least August.

Includes additional documentation for the processing of updating ClinVar variants.

Includes additional flags in `setup.py` to allow clusters to be created successfully. 

Resolves #961

### DIFF
--- a/data-pipeline/src/data_pipeline/datasets/clinvar.py
+++ b/data-pipeline/src/data_pipeline/datasets/clinvar.py
@@ -105,8 +105,18 @@ def _parse_variant(variant_element):
     location_elements = variant_element.findall("./InterpretedRecord/SimpleAllele/Location/SequenceLocation")
     for element in location_elements:
         try:
+            chromosome = element.attrib["Chr"]
+            # A release in August 2022 introduced several Variants with a Chromosome of 'Un'
+            #   which caused failure of this pipeline when compared to the reference genome
+            if chromosome == "Un":
+                variant["locations"] = {}
+                allele_element = variant_element.findall("./InterpretedRecord/SimpleAllele")
+                print(
+                    f'Skipping variant with Allele ID: {allele_element.attrib["AlleleID"]} due to anomalous Chromosome value of "Un"'
+                )
+                break
             variant["locations"][element.attrib["Assembly"]] = {
-                "locus": element.attrib["Chr"] + ":" + element.attrib["positionVCF"],
+                "locus": chromosome + ":" + element.attrib["positionVCF"],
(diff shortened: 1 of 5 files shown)


data-pipeline/src/data_pipeline/datasets/clinvar.py
```diff
@@ -105,8 +105,13 @@ def _parse_variant(variant_element):
     location_elements = variant_element.findall("./InterpretedRecord/SimpleAllele/Location/SequenceLocation")
     for element in location_elements:
         try:
+            chromosome = element.attrib["Chr"]
+            # A release in August 2022 introduced several Variants with a Chromosome of 'Un'
+            #   which caused failure of this pipeline when compared to the reference genome
+            if chromosome == "Un":
```
phildarnowsky-broad, line 111: Looks OK but I wonder if we want to add a warning to the effect of "Skipping Variant with bad chromosome value 'Un'", what do we do in comparable cases elsewhere in this codebase?

phildarnowsky-broad (approved):

data-pipeline/src/data_pipeline/datasets/clinvar.py
rileyhgrant: I'm on board with adding a bit more output in special cases. In a similar case (resulting from a key error when parsing) simply skips the variant without output. It seems this "Un" is a bit more specific, and I've included output per your suggestion now.

PR #985, opened 2022-09-21
Add Jest snapshot tests for static pages

rileyhgrant:
Includes Jest snapshot tests for the static pages:

- Home
- About
- Team
- Downloads
- Policies
- Publications
- Feedback
- Help

And accompanying helper files and minor changes to let the snapshots be both created and correct (a space that got dropped here and there, a jest transformer to mock out `.md` files, etc)

These represent the mostly static pages that are unlikely to change. Snapshot tests for these unchanging pages could help in catching unintentional regressions going forward.


rileyhgrant:
While this transformer doesn't render the `.md` content exactly as it would be on the browser, I'd argue it still represents a very useful set of tests. 
- `@gnomad/markdownLoader` is tested in the [gnomad-browser-toolkit repository](https://github.com/broadinstitute/gnomad-browser-toolkit/tree/main/packages/markdown-loader/src/__tests__)
- Snapshotting the raw `.md` files as strings catches regressions of the content itself.

Thoughts / opinions? This seems like an arguably correct division of responsibility to me. Also, candidly, implementing an Async transformer to use the `@gnomad/markdown-loader` was proving incredibly difficult for me.

rileyhgrant:
The Jest testing function `toMatchInlineSnapshot()` seems like a good fit for many tests that don't rely on looping over datasets, however the pre-commit hook `trailing-whitespace` won't accept the string literals the Jest function outputs.

I removed the pre-commit hook for now, but there may be a better way to do this.

rileyhgrant:
Also - there seem to be some strange configuration differences happening, Phil and I were unable to replicate the same error on both our local machines. I have re-created a snapshot for the `DatasetSelector.spec.tsx` test, as my machine was mirroring the errors that CI was outputting. Doing so is causing CI to succeed now, however there still seems to be possibly something weird going on, and may warrant discussion in the near future.

rileyhgrant:
Re: discussion Phil and I had about differing newline characters possibly causing snapshot test issues, I have checked and both the previous and new snapshot for `DatasetSelector.spec.tsx`, one that was causing particular issues, used `LF` as the newline character, per 

```
git ls-files --eol
```

Still stumped on what actually changed, and what is causing the different development environments on CI, my machine, and Phil's machine.

rileyhgrant:
> The Jest testing [function `toMatchInlineSnapshot()`](https://jestjs.io/docs/snapshot-testing#inline-snapshots) seems like a good fit for many tests that don't rely on tests that use some variation of a loop ...

Re: a discussion with Phil some time back, this PR has moved away from doing inline snapshots, and kept the pre-commit hooks intact. Writing a comment here to keep a logical flow in the thread, and also added in edit to the initial message just so it's clear that that is no longer the case.

browser/src/DownloadsPage/DownloadsPage.spec.tsx
```diff
@@ -0,0 +1,15 @@
...
+
+import React from 'react'
+import renderer from 'react-test-renderer'
+
+import DownloadsPage from './DownloadsPage'
+
+import { withDummyRouter } from '../../../tests/__helpers__/router'
+
+// TODO: (rgrant) add mocked data for each of the different types of layout effects on line 44 of DownloadsPage.tsx
+test('Downloads Page has no unexpected changes', () => {
```
phildarnowsky-broad, line 12: @rileyhgrant there's an easy way to do this, remember that these specs are just ordinary code, so something like this ought to work:

```
const locationPrefixes = ['#v2-liftover', '#v3-', '#exac-', '#research-']

locationPrefixes.forEach((locationPrefix) => {
  test(`Downloads page with ${locationPrefix} location prefix`, () => {
    const tree = renderer.create(withDummyRouter(<DownloadsPage location={{ hash: locationPrefix }} />))
    expect(tree).toMatchSnapshot()
  })
})
```

...and you get four tests for the price of one.

phildarnowsky-broad (changes requested):
one suggestion to allow you to remove that TODO right now

browser/src/DownloadsPage/DownloadsPage.spec.tsx
rileyhgrant: Aha - ok that makes sense. Thanks!

phildarnowsky-broad (approved):

Issue #981, opened 2022-09-14
Update components that render conditional on dataset (part 2)


phildarnowsky-broad:
See https://github.com/broadinstitute/gnomad-browser/issues/853 for details.

PR #991, opened 2022-09-28
Centralized metadata part 2

phildarnowsky-broad:
Per https://github.com/broadinstitute/gnomad-browser/issues/981 this is the second in the "centralized metadata" series. 

Aside from continuing to extract the dataset metadata, the big change here is the introduction of some central `Gene` and `Transcript` types. I'm guessing a declaration of `Variant` may follow soon.


phildarnowsky:
testing just because I'm curious if I can comment from my personal account, please disregard

phildarnowsky-broad:
resolves #981

phildarnowsky-broad:
@mattsolo1 you might be interested in seeing this too, sorry I didn't tag you earlier, still getting used to you being back

rileyhgrant (approved):
@phildarnowsky-broad 
Looks good to me! Nice to have these PRs to review as good reference for how to touch up TypeScript typing and adding Testing while touching any given parts of the code.

One minor difference in behavior, that I personally think is acceptable or even preferable, will comment below saying what it is.

rileyhgrant:
I found one minor difference in the amount of sig figs displayed in the constraint table. The live browser shows a few less digits for a few parameters, but hovering a mouse over these figures shows that they are rounded values (0.000 real value -> 0 in the browser). To what place this rounding is happening, differs in the browser and the local branch depending on where the first decimal that's not a 0 occurs.

It seems to me that if the specificity is there, reducing the amount of sig figs is a choice? Going the other direction is a no no - implying more precision where there is none is bad, but I've never had discussed with me the implications reducing specificity when its there. To my eyes it looks fine, but I am of course, less close to the underlying science.

In the screenshots below, expected SNVs for missense, PLI, and one value in the o/e for synonymous all gain more sig figs on the local branch (e.g. 435 -> 435.0). However, on both branches, hovering the mouse over either of these values shows the value its using, which goes to 3 decimal points of precision (hovering the 435 shows instead 434.988). In both cases, the number is being truncated, the only difference is how much this is happening.

Live browser:

Screen Shot 2022-10-03 at 09 23 30



This branch:

Screen Shot 2022-10-03 at 09 23 35

phildarnowsky-broad:
@rileyhgrant good catch, that is a legit regression--I see where you're coming from about perhaps wanting to expand the number of sigfigs, but as a rule, when I'm doing a big refactoring like this, even if I see a really obvious and tempting opportunity for improvement, I save it for later.

rileyhgrant (changes requested):
> @rileyhgrant good catch, that is ...

That makes sense - purely update under the hood. As an aside, I'd be interested on hearing your and Matt's thoughts on this case of number of sig figs - just seems like an interesting thing I've not thought about, reducing specificity by choice.

I've re-requested myself for review to turn my approval a request changes, per your comment implying that you believe this to be not ready to be merged.

phildarnowsky-broad:
@rileyhgrant should be good to go here--note that I've also taken the opportunity to add a few tests to the broken function itself, always a worthwhile thing to do when you fix a regression.

rileyhgrant (approved):

PR #996, opened 2022-10-12
Add robots.txt

phildarnowsky-broad:
As written, this `robots.txt` allows bots access to any page in the top black navbar, and nothing else.

### DIFF
--- a/browser/webpack.config.js
+++ b/browser/webpack.config.js
@@ -91,6 +91,9 @@ const config = {
     new CopyWebpackPlugin({
       patterns: [path.resolve(__dirname, './src/opensearch.xml')],
     }),
+    new CopyWebpackPlugin({
+      patterns: [path.resolve(__dirname, './src/robots.txt')],
+    }),
     new EnvironmentPlugin({
       REPORT_VARIANT_URL: null,
       REPORT_VARIANT_VARIANT_ID_PARAMETER: null,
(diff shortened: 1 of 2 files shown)


browser/webpack.config.js
```diff
@@ -79,6 +79,15 @@ const config = {
           loader: '@gnomad/markdown-loader',
         },
       },
+      {
+        test: /robots\.txt$/,
+        use: {
+          loader: 'file-loader',
+          options: {
+            outputPath: 'assets/images',
```
mattsolo1, line 87: Not sure if it makes sense to go to images folder... but not a big deal I guess

rileyhgrant, line 87: Does this `outputPath` parameter need to change? Googling suggests that a `robots.txt` file has to be at the root of the build output. 

Perhaps `'assets/images'` -> `'/'`

phildarnowsky-broad:
@mattsolo1 @rileyhgrant @sjahl stand by, this PR isn't right--worked on my machine of course.

browser/robots.txt
```diff
@@ -0,0 +1,11 @@
+User-Agent: *
+Disallow: *
```
sjahl, line 2: We might want to allow just the home page/root URL, which I think would be done by:

`Allow: /$`

phildarnowsky-broad:
@rileyhgrant @mattsolo1 @sjahl let's try that again.

sjahl (approved):
This looks good to me. Google has a robots.txt tester that might be worth using to confirm that this config has the desired effect: https://support.google.com/webmasters/answer/6062598?hl=en

Though, it does look like the file has to be deployed onto the live site in order to test.

PR #1005, opened 2022-10-14
Fix faulty GraphQL queries

rileyhgrant:
Fixes several faulty GraphQL queries that caused variant pages network requests to fail (and displayed "`Unknown operation ""`) , and changes the default behavior of GraphQL queries to succeed if `operationName` is not provided.

When doing minor refactoring of GraphQL queries to include the `operationName` field in the body, which helps helps with mocking out network requests in tests, a few queries were missed due to an incomplete method of finding all files that use `BaseQuery`. The default `operationName` prop of an empty string (`""`) caused these queries that were not included in the refactoring to fail. 

The field `operationName` has been added to these queries, and the default prop has been changed to `null`, which causes the queries to act exactly as before in the case that an `operationName` is not provided as a prop to `BaseQuery`.

### DIFF
--- a/browser/src/GenePage/MitochondrialGeneCoverageTrack.tsx
+++ b/browser/src/GenePage/MitochondrialGeneCoverageTrack.tsx
@@ -37,6 +37,7 @@ const MitochondrialGeneCoverageTrack = ({ datasetId, geneId }: Props) => {
 
   return (
     <Query
+      operationName={operationName}
       query={query}
       variables={{ geneId, datasetId, referenceGenome: referenceGenome(datasetId) }}
       loadingMessage="Loading coverage"
--- a/browser/src/MNVPage/MNVDetailsQuery.tsx
+++ b/browser/src/MNVPage/MNVDetailsQuery.tsx
@@ -2,8 +2,9 @@ import React from 'react'
 
 import { BaseQuery } from '../Query'
 
+const operationName = 'MultiNucleotideVariant'
 const query = `
-query MultiNucleotideVariant($variantId: String!, $datasetId: DatasetId!) {
+query ${operationName}($variantId: String!, $datasetId: DatasetId!) {
   multiNucleotideVariant(variant_id: $variantId, dataset: $datasetId) {
     variant_id
     chrom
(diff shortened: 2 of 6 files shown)


phildarnowsky-broad (changes requested):
@rileyhgrant code looks good overall, however I'd like to also see some tests to guard against regression. In this case it'd be appropriate to have a test that explicitly shows that the API works with a `null` `operationName`. That would probably best go in the `graphql-api` subtree.

rileyhgrant:
As discussed on zoom, this is known to be not best practice, but I am opening an issue to come back to this and think about / write tests to catch this in an automated way. These requires some thought, as as I understand it, these tests need to validate an incoming http requests body's field as being valid queries for GraphQL. Tests from the frontend that mock out network responses would not have caught this error, as this came from live API responses.

For the time being, here are three screenshots of Firefox's inspect window showing the behavior obtained by changing the default of `operationName` to `null`. An issue has been opened (#1009) and I've assigned myself as a sort of 'IOU' to come back to this, and think about and write automated unit tests for the backend that can catch this without the need for manual testing.

Here is the behavior before `operationName` was introduced, `200` response to a successful network request, when there is no `operationName` included in the body.

no_operationName



Here are the two behaviors after this change, one when `operationName` is defined as the Query's operation name, and one where `operationName` was not passed to the frontend Query, causing it to default to `null`, both of these network requests respond with `200`s and return the expected data.

string_operationName

null_operationName

I've also navigated to each major type of page (Gene, Transcript, Region, and Variant) each with a few flavors of data type, to round out the manual testing of this PR, and noticed no bugs.

rileyhgrant:
> @rileyhgrant code looks good overall, however I'd ...

@phildarnowsky-broad Per the zoom call earlier today, I'm hoping to get this approved without the inclusion of the recommended automated tests that can catch regressions of this type. As such I have re-requested review.

This is with the understanding that this is ***not*** the preferred workflow for typical PRs. For this particular fix, automated tests will be created at a later date, allowing this PR to be merged to fix the faulty behavior, and to allow for further changes to be merged into main and deployed as part of browser changes for ASHG 2022.

mattsolo1 (approved):

PR #1027, opened 2022-11-08
Display all variants at stop position in variant list

rileyhgrant:
Per user email and related discussion in this [slack thread](https://atgu.slack.com/archives/CNL7NA6H2/p1667939930278419?thread_ts=1667886205.261139&cid=CNL7NA6H2).

The frontend performs some logic to filter the list of variants recieved down to only variants in specific regions. This PR changes the `Lodash` function used in the stop index of that filtering from `sortedIndexBy` to `sortedLastIndexBy` so that lists of variants that are filtered include all variants that occur at the last position, instead of just the first one.

The example in the email now correctly shows all 3 variants when the region is specified as the single location

Screen Shot 2022-11-08 at 16 05 45

---

Includes unrelated change to the regional genomic constraint track, previously tiny regions that did not have genomic constraint data would use the same message as regions that are too large. Now small regions without data show a message that correctly describes why nothing is shown.

### DIFF
--- a/browser/src/RegionViewer/filterVariantsInZoomRegion.ts
+++ b/browser/src/RegionViewer/filterVariantsInZoomRegion.ts
@@ -1,4 +1,4 @@
-import { sortedIndexBy } from 'lodash-es'
+import { sortedIndexBy, sortedLastIndexBy } from 'lodash-es'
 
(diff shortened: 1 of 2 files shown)


rileyhgrant:
> Tip for the future: it is ...

Aha, that makes total sense - I'll use your suggestions when making PRs in the future!

Issue #929, opened 2022-07-08
Upgrade to Elasticsearch 7.17+ and ECK operator


mattsolo1:
See: https://github.com/broadinstitute/gnomad-browser/issues/928

peternixon:
The below patch allows deployment with the current code base:
diff --git a/deploy/deployctl/subcommands/setup.py b/deploy/deployctl/subcommands/setup.py
index 219a25a8..94a501b7 100644
--- a/deploy/deployctl/subcommands/setup.py
+++ b/deploy/deployctl/subcommands/setup.py
@@ -210,7 +210,7 @@ def create_cluster() -> None:
             f"--zone={config.zone}",
             "--release-channel=stable",
             "--enable-autorepair",
-            "--enable-autoupgrade",
+            "--cluster-version=1.21.14-gke.3000",
             "--maintenance-window=7:00",
             f"--service-account={config.gke_service_account_full_name}",
             f"--network={config.network_name}",

sjahl:
Just following some threads on what our elasticsearch export looks like, It kinda seems like we may be limited to Elasticsearch 7 on the Hail end of things: https://github.com/hail-is/hail/blob/b124e86925221bc1c82662ff429e4ab480c4b080/hail/python/hail/methods/impex.py#L2950

I see some discussion in the Hail zulip that indicates at least one person has managed to compile hail with ES 8 support. On one hand, I don't like the idea of doing any trailblazing here, but on the other, Elastic is EOL-ing ES 7 in August 2023. So, I think it makes sense to upgrade to 7 since we have to do that anyway to make it to 8, and then on the longer term we can start trying to work out ES 8 support status in Hail.

sjahl:
ES7 Compatibility (both in the k8s manifests and API service) was shipped in https://github.com/broadinstitute/gnomad-browser/commit/65e4099bfcea184512eb620a44cf489f65d20c17

PR #1033, opened 2022-11-16
Force GKE to an older version known to be compatible to the ECK 1.2.1…

peternixon:
… (until upstream has a chance to upgrade)

The current codebase doesn't appear to be installable from scratch without this change!


mattsolo1:
Thanks @peternixon. Rather than pinning the older GKE version, which will soon be unsupported by Google Cloud, we are planning to upgrade the ECK operator today (keep an eye on: https://github.com/broadinstitute/gnomad-browser/issues/929 and a linked PR coming soon). 

If installation still doesn't work for you after this, please let us know in a new issue.

PR #1043, opened 2022-11-22
Fix external dbSNP link

### LINKED ISSUE #1032: External Resource Link to dbSNP returns error
On VariantPage the link to dbSNP returned an error - SNP is valid and I have been experiencing it for all SNPs,  It looks like they have removed that URL.  I am getting this on gene.iobio (and, a separate issue, from them, their variant link to gnomAD goes to https://gnomad.broadinstitute.org/variant/11-128485068-T-C which can't be found (ah! because they are not adding _?dataset=gnomad_r3_))

I have traced the ExternalLink to lines 36 and 54 in the ReferenceList module:

https://github.com/broadinstitute/gnomad-browser/blob/main/browser/src/VariantPage/ReferenceList.tsx

Both lines use : href={`https://www.ncbi.nlm.nih.gov/projects/SNP/snp_ref.cgi?rs=<rsID>

This page, for example:  https://www.ncbi.nlm.nih.gov/projects/SNP/snp_ref.cgi?rs=rs143246089  = Error 404

Correct URL appears to be https://www.ncbi.nlm.nih.gov/snp/rs143246089 - but haven't tested for line 36 (multiple rs IDs)

Oliver

rileyhgrant:
Resolves #1032 

At some point, the url convention for ncbi's dbSNP changed, causing gnomAD's links to dbSNP's variants by rsID to be broken links. This is a straightforward PR the changes instances in which this resource is linked to (`ReferenceList` and `MitochondrialVariantReferenceList`) to use the correct URL.

It also DRYs up the links a bit, and adds very small scoped unit tests. 

Localhost links for convenience:
* [This variant page](http://localhost:8008/variant/1-55505552-A-ACTGCTGCTG) has both a dbSNP and ClinVar link
* [This mitochondrial variant page](http://localhost:8008/variant/M-8602-T-C?dataset=gnomad_r3) has both a dbSNP and ClinVar link

---

Unrelated - in looking at all these components that use `<Link>` and `<ExternalLink>` that have related `ts` errors, I will likely make an issue to take a look at those (they're part of `gnomad-browser-toolkit`) to fix those errors so we can remove a whole slew of `ts-expects`. Looks like there are quiiiite a few `ts-expects` hanging around the codebase that could be resolved by taking some time to look at static typing errors that arise from the interaction of `gnomad-browser` and `gnomad-browser-toolkit` at some point.

### DIFF
--- a/browser/src/MitochondrialVariantPage/MitochondrialVariantDetailPropType.ts
+++ b/browser/src/MitochondrialVariantPage/MitochondrialVariantDetailPropType.ts
@@ -1,5 +1,23 @@
 import PropTypes from 'prop-types'
 
+type ClinvarVariant = {
+  clinical_significance: string
+  clinvar_variation_id: string
+  gold_stars: number
+  last_evaluated: string
+  review_status: string
+  submissions: {
+    clinical_significance: string
+    conditions: {
+      name: string
+      medgen_id: string
+    }[]
+    last_evaluated: string
+    review_status: string
+    submitter_name: string
+  }
+}
+
 type MitochondrialVariantDetailPropType = {
   alt: string
   an: number
@@ -26,6 +44,8 @@ type MitochondrialVariantDetailPropType = {
   ref: string
   reference_genome: string
   variant_id: string
+  rsids?: string[]
+  clinvar: ClinvarVariant
 }
 
 // @ts-expect-error TS(2322) FIXME: Type 'Requireable<InferProps<{ alt: Validator<stri... Remove this comment to see the full error message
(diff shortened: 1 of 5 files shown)


phildarnowsky-broad (changes requested):
Looks fine to me except for one small style note, `ncbiReference` and `clinvarReference` are React components, so should be capitalized, `NcbiReference` and `ClinvarReference` respectively.

mattsolo1:
@phildarnowsky-broad for very minor changes like those requested here (capitalization, etc), does it make sense to approve the PR under the assumption they will be an easy fix (rather than going through the whole request changes/approval cycle again)?

phildarnowsky-broad (approved):

phildarnowsky-broad:
@mattsolo1 yes good call on that

phildarnowsky-broad:
@mattsolo1 one thing about that though, we'd still need an approval to merge, so if we wanted to do as you suggest we'd have to approve but with a note saying "please do X before merging". Not a problem except for the possibility of prematurely merging before you actually do X, so that's something we'd have to be careful about.

rileyhgrant:
Makes sense to me too on giving approval with minor stipulations (the caps here) going forward.
 
Thanks all!

PR #1044, opened 2022-11-25
Make the Google Storage Buckets configurable to allow 3rd party insta…

peternixon:
…llation

### DIFF
--- a/deploy/deployctl/config.py
+++ b/deploy/deployctl/config.py
@@ -14,7 +14,8 @@ class Configuration:
     domain: str = None
     environment_tag: str = None
     authorized_networks: str = "0.0.0.0/0"
-    data_pipeline_output: str = None
+    data_pipeline_output: str = "gs://gnomad-browser-data-pipeline"
+    data_snapshot_bucket: str = "gs://gnomad-browser-elasticsearch-snapshots"
 
     def __init__(self, config_path):
         self._config_path = config_path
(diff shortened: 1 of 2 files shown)


deploy/deployctl/config.py
```diff
@@ -14,7 +14,8 @@ class Configuration:
     domain: str = None
     environment_tag: str = None
     authorized_networks: str = "0.0.0.0/0"
-    data_pipeline_output: str = None
+    data_pipeline_output: str = "gs://gnomad-browser-data-pipeline"
+    data_snapshot_bucket: str = "gs://gnomad-browser-elasticsearch-snapshots"
```
mattsolo1, line 18: Would be good to add the snapshot configuration to README here: https://github.com/broadinstitute/gnomad-browser/blob/007388ffe8568940f71720bbde20874f9c1bd9ce/data-pipeline/README.md?plain=1#L52

deploy/deployctl/config.py
```diff
@@ -14,7 +14,8 @@ class Configuration:
     domain: str = None
     environment_tag: str = None
     authorized_networks: str = "0.0.0.0/0"
-    data_pipeline_output: str = None
+    data_pipeline_output: str = "gs://gnomad-browser-data-pipeline"
```
mattsolo1, line 17: Can we leave the defaulting to `None`, since most people can't write to our bucket?  Looks like we check to see if the configuration is complete here: https://github.com/broadinstitute/gnomad-browser/blob/007388ffe8568940f71720bbde20874f9c1bd9ce/deploy/deployctl/subcommands/data_pipeline.py#L24.

You could add another check for the snapshots bucket under that.

mattsolo1 (changes requested):

mattsolo1:
Closing due to inactivity.

PR #1047, opened 2022-11-29
Fix clinvar pipeline bug

### LINKED ISSUE #1046: AttributeError: 'list' object has no attribute 'attrib'
We're trying to get the gnomAD data pipeline running following the guides in this repository. I'm hitting this error while running the `clinvar_grch37` data pipeline step which is parsing the [2022-11-20](https://ftp.ncbi.nlm.nih.gov/pub/clinvar/xml/clinvar_variation/weekly_release/ClinVarVariationRelease_2022-1120.xml.gz) Clinvar XML file:

```
Parsing XML file
  1%|          | 12476/1100000 [00:13<20:42, 875.10it/s]Failed to parse variant 151314
Traceback (most recent call last):
  File "/tmp/1176299b30c0490db29348b9b1b477a1/clinvar_grch37.py", line 80, in <module>
    run_pipeline(pipeline)
  File "/tmp/1176299b30c0490db29348b9b1b477a1/pyfiles_e3ra1vi2.zip/data_pipeline/pipeline.py", line 200, in run_pipeline
  File "/tmp/1176299b30c0490db29348b9b1b477a1/pyfiles_e3ra1vi2.zip/data_pipeline/pipeline.py", line 167, in run
  File "/tmp/1176299b30c0490db29348b9b1b477a1/pyfiles_e3ra1vi2.zip/data_pipeline/pipeline.py", line 132, in run
  File "/tmp/1176299b30c0490db29348b9b1b477a1/pyfiles_e3ra1vi2.zip/data_pipeline/datasets/clinvar.py", line 191, in import_clinvar_xml
  File "/tmp/1176299b30c0490db29348b9b1b477a1/pyfiles_e3ra1vi2.zip/data_pipeline/datasets/clinvar.py", line 115, in _parse_variant
AttributeError: 'list' object has no attribute 'attrib'
```

This is run on dataproc using the command `./deployctl data-pipeline run --cluster vep85 clinvar_grch37`. It appears to be caused by the f-string in this print statement:

```python
allele_element = variant_element.findall("./InterpretedRecord/SimpleAllele")
print(f'Skipping variant with Allele ID: {allele_element.attrib["AlleleID"]} due to anomalous Chromosome value of "Un"')
```

Seems like the `findall` method is returning a list of XML nodes when the f-string expects a single node to be present. The cluster is using python 3.8 if that helps.

rileyhgrant:
Resolves #1046 

Quick one line that fixes the output on the ClinVar pipeline to allow it to parse the xml without errors. I tested running the pipeline on a dataproc cluster, and it once again successfully parses the xml.

---

**Context / Opinion Question**

A release at some point introduced variants with a chromosome called '`Un`'. 

Additional logic was added to skip these and let the pipeline run. Later, a couple lines were added to give additional information to the runner of the pipeline when it skipped a variant, output currently looks like:

```
| [1 files][  2.0 GiB/  2.0 GiB]      0.0 B/s
Parsing XML file
  1%|          | 12990/1100000 [00:13<19:28, 930.40it/s] Skipping variant with Allele ID: 161065 due to anomalous Chromosome value of "Un"
 44%|████▍     | 484551/1100000 [05:52<07:45, 1321.52it/s] Skipping variant with Allele ID: 1676726 due to anomalous Chromosome value of "Un"
 83%|████████▎ | 913219/1100000 [11:42<02:00, 1549.61it/s]
```

This gives more information to the user, but also causes the progress bar to behave differently - it leaves behind lines of output when it skips a chromosome and notifies the user.

This PR fixes the output so that the pipeline can be run without error, but I'm curious as to opinions on if the output should be removed altogether - so that the pipeline just skips these variants quietly, and causes the progress bar to not leave behind afterimages.

Previously, it wouldn't leave behind afterimages:

```
| [1 files][  2.0 GiB/  2.0 GiB]      0.0 B/s
Parsing XML file
 83%|████████▎ | 913219/1100000 [11:42<02:00, 1549.61it/s]
```

Output - yay or nay?

### DIFF
--- a/data-pipeline/src/data_pipeline/datasets/clinvar.py
+++ b/data-pipeline/src/data_pipeline/datasets/clinvar.py
@@ -112,7 +112,7 @@ def _parse_variant(variant_element):
                 variant["locations"] = {}
                 allele_element = variant_element.findall("./InterpretedRecord/SimpleAllele")
                 print(
-                    f'Skipping variant with Allele ID: {allele_element.attrib["AlleleID"]} due to anomalous Chromosome value of "Un"'
+                    f' Skipping variant with Allele ID: {allele_element[0].attrib["AlleleID"]} due to anomalous Chromosome value of "Un"'
                 )
                 break
             variant["locations"][element.attrib["Assembly"]] = {


phildarnowsky-broad:
FWIW I don't mind the breaks in the output, but it seems like there'd be an easy way to avoid that. What if, when you hit an `Un`, you don't `print` anything right away but just add it to a running list, and then you print the whole list at the end in one shot?

rileyhgrant:
> FWIW I don't mind the ...

Aha, that makes sense. If we think that skipping an 'Un' variant is useful information, I'm down to move to the list and print at the end.

mattsolo1:
I don't mind progress bar breaks either, but happy with Phil's suggestion too. I don't know about the importance of this 'Un' chromosome.

rileyhgrant:
Alright, in that case, if no one strongly objects to the breaks lets keep it that way.

If there are more ClinVar xml quirks to sort through in the future, we can 'burn that bridge when we get to it'.

PR #1050, opened 2022-12-01
Remove `requester-pays-allow-buckets` parameter from Dataproc cluster creation

lgruen:
I couldn't find the `gnomad-public-requester-pays` bucket being mentioned anywhere else in the code, so I'm hoping the setting isn't necessary?

Removing this would provide some peace of mind when running the pipelines with workers outside the US :).


mattsolo1 (approved):

sjahl:
Yes, that requester pays bucket is no longer used after gnomAD data was made available in the [GCP public datasets program](https://cloud.google.com/life-sciences/docs/resources/public-datasets/gnomad).

I could perhaps see there being some value in making this a configurable/parameter so that users can whitelist other requesterpays buckets as needed, but I don't think that needs to be done as part of this PR.

lgruen:
Awesome, thanks for confirming!

I can't merge this, as I don't have write access.

PR #1052, opened 2022-12-01
Add back to gene button on variant page

### LINKED ISSUE #1025: Add way to easily navigate from Variant page to Gene page
Elevate list of genes higher up on page.

---

Per Jan 17, 2023 roadmapping meeting:

Very often, clicking on a link to a Gene from any page in gnomAD will direct the user to an external resource instead of another gnomAD page. This results in users being hesitant to click on links.

Also, the current location of the link to a Gene in the Vep section (halfway down the page) makes it easy to lose if a user opens a bunch of unrelated Variant pages. In addition, knowing which Gene this Variant occurs in is a useful piece of information to have at a glance, when many Variant pages are open.

The suggestion is to add a link somewhere up top, to both give context to the variant, as well as a way to navigate to the corresponding gene. This could be located next to the "Copy Variant ID" button so it doesn't take up a lot of space. 

We could use MANE to select the Gene to show if there are multiple Genes associated with a Single variant. Alternatively, we could keep the view as it is (only have the Gene's be present in the VEP section) if there are multiple Genes, and only add the link up top if the Variant is associated with a single Gene.

rileyhgrant:
Resolves #1025 

Edit: Updated screenshot and description in initial text here for clarity:

Adds a "to gene" button on a variant page.

Checks to see if in the transcript consequences there's a MANE select transcript, if so, it links to that gene. If there's no MANE select transcript, it checks for a single canonical transcript, if there are multiple canonical transcripts (multiple genes) and no MANE select, it will not display a "Gene page" button.

- [Link to](http://localhost:8008/variant/1-55051215-G-GA?dataset=gnomad_r3) a Variant with a MANE select transcript ("Gene page" button links to Gene associated with MANE select transcript, here, PCSK9)
  - Screenshot 2023-04-05 at 12 25 59

- [Link to](http://localhost:8008/variant/9-38577187-G-A?dataset=gnomad_r2_1) a Variant with multiple canonical transcripts, and no MANE select transcript (no "Gene page" button)
  - Screenshot 2023-04-05 at 12 26 10

If there are multiple 

---

## Outdated, kept for context (originally Dec 01, 2023):

Adds a more prominent link to the Gene page from the Variant page.

Localhost [link here](http://localhost:8008/variant/1-55051215-G-GA?dataset=gnomad_r3).

Screen Shot 2022-12-01 at 10 11 45

Uses the ENSG ID from the canonical transcript for a given gene.

### DIFF
--- a/browser/src/VariantPage/VariantPage.tsx
+++ b/browser/src/VariantPage/VariantPage.tsx
@@ -687,7 +687,44 @@ type VariantPageProps = {
   variantId: string
 }
 
+// Returns the gene_id to be linked to in the Variant page header
+// The vast majority of cases have a single gene associated, the logic for those that don't is:
+//   If there is a mane select transcript, use that to determine which gene to link to
+//   If there is no mane select transcript, check if there is a single canonical transcript
+//     If there is a single canonical transcript, link to that gene
+// As such, if a variant has multiple associated genes, multiple canonical transcripts, and
+//   no MANE select transcript, the "Gene page" button will not appear as it is non-trivial,
+//   and users should navigate further down the page to the VEP consequences section
+const checkGeneLink = (transcript_consequences: TranscriptConsequence[] | null) => {
+  if (!transcript_consequences) {
+    return null
+  }
+
+  const maneSelectTranscript = transcript_consequences.filter(
+    (transcript: TranscriptConsequence) => transcript.is_mane_select
+  )
+
+  if (maneSelectTranscript.length === 1) {
+    return {
+      ensembleId: maneSelectTranscript[0].gene_id
+    }
+  }
+
+  const canonicalTranscripts = transcript_consequences.filter(
+    (transcript: TranscriptConsequence) => transcript.is_canonical
+  )
+
+  if (canonicalTranscripts.length !== 1) {
+    return null
+  }
+
+  return {
+    ensembleId: canonicalTranscripts[0].gene_id,
+  }
+}
+
 const VariantPage = ({ datasetId, variantId }: VariantPageProps) => {
+  let gene = { ensembleId: '' }
   return (
     // @ts-expect-error TS(2746) FIXME: This JSX tag's 'children' prop expects a single ch... Remove this comment to see the full error message
     <Page>
@@ -746,6 +783,14 @@ const VariantPage = ({ datasetId, variantId }: VariantPageProps) => {
               liftover: data.liftover,
               liftover_sources: data.liftover_sources,
             }
+
+            // In this branch, a variant was successfully loaded. Check the symbol
+            //   and ensemble ID to create a 'Gene page' button with the correct link
+            const geneData = checkGeneLink(variant.transcript_consequences)
+            if (geneData) {
+              gene.ensembleId = geneData.ensembleId
+            }
+
             pageContent = <VariantPageContent datasetId={datasetId} variant={variant} />
           }
 
@@ -763,17 +808,29 @@ const VariantPage = ({ datasetId, variantId }: VariantPageProps) => {
                 }}
                 selectedDataset={datasetId}
                 extra={
-                  navigator.clipboard &&
-                  navigator.clipboard.writeText && (
-                    <Button
-                      onClick={() => {
-                        navigator.clipboard.writeText(variantId)
-                      }}
-                      style={{ margin: '0 1em' }}
-                    >
-                      Copy variant ID
-                    </Button>
-                  )
+                  <>
+                    {navigator.clipboard && navigator.clipboard.writeText && (
+                      <Button
+                        onClick={() => {
+                          navigator.clipboard.writeText(variantId)
+                        }}
+                        style={{ margin: '0 1em' }}
+                      >
+                        Copy variant ID
+                      </Button>
+                    )}
+
+                    {gene.ensembleId && (
+                      <Button
+                        onClick={() => {
+                          location.href = `/gene/${gene.ensembleId}?dataset=${datasetId}`
+                        }}
+                        style={{ margin: '0 1em 0 -0.33em' }}
+                      >
+                        Gene page
+                      </Button>
+                    )}
+                  </>
                 }
               >
                 <VariantPageTitle variantId={variantId} datasetId={datasetId} />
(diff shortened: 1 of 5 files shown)


mattsolo1:
When variant is found in multiple genes, should make plural (i.e. "Variant Found in Genes").

Example: http://localhost:8008/variant/9-38577187-G-A?dataset=gnomad_r2_1

(note this gene not greatest example b/c it has same gene name but with different gene ids, but you get the idea)

image

mattsolo1:
When you think it is ready, can you post a screenshot on slack to Heidi as well since she requested this feature?

rileyhgrant:
Amazing - thanks for the feedback and on the example of the variant with multiple different associated ENSG IDs for canonical transcripts. Worth it to include ENSG ID in parentheses? 

I didn't realize that there could be multiple ENSG IDs associated with a single gene symbol.

mattsolo1:
It's the first example I've seen with identical gene symbols (and the only example I could easily find using Slack search). I think this is rare enough scenario that we don't need the ENSG in parentheses, unless others ask for it.

rileyhgrant:
Per comments from the Roadmapping Meeting on Jan 17, 2023, here are a few possible solutions.

I'm a bit partial to Anne's suggestion of having a link to the Gene come before the (SNV, Insertion, etc). 

As I understand it, the idea is to at a glance get a sense of what Gene this Variant occurs in, if it's that simple. If the variant is part of the small number that occur in more than one Gene, than we can remove the Gene on the header (and maybe suggest users go further down the page if there's room to say that). 

Here are a couple screenshots of possible locations the Gene symbol could occur. I welcome any and all feedback, as this has turned into a more discussion provoking request than I thought at first. Any thoughts?




---
All these are shown in the simplest case (Variant occurs cleanly only in a single Gene, decided by the single canonical transcript).

Gene symbol before the Variant information 
Screenshot 2023-01-31 at 09 14 44


Gene symbol after the Variant information
Screenshot 2023-01-31 at 08 46 07


Gene symbol after (or could be before) the 'Copy Variant ID' button.
Screenshot 2023-01-31 at 08 47 57

---

The last screenshot additionally has a temporary placeholder unicode arrow. Part of this issue is that Gene links often direct to external resources, which makes some users shy away from clicking Gene links. Is there some icon or styling that we can apply to intuitively suggest that certain links are internal? Maybe if we make a SVG Favicon that's simple and recognizable, this SVG could come after/before internal links? Just a thought.

@phildarnowsky @mattsolo1 - opinions are encouraged. I'd feel good about having another set of eyes/thoughts from the team before going back into the thread in the gnomad browser channel for another round of comments.

phildarnowsky-broad:
Looking at the screenshots you have here, I think the first, with the gene symbol leading, is my preference.

There definitely are some idiomatic icons that suggest an internal link, not sure what they're called but imagine a diagonal arrow coming out of a square...I see that often. Do we have a preferred way to add small icons like that @mattsolo1 ?

browser/src/VariantPage/VariantPage.tsx
```diff
@@ -76,7 +86,7 @@ type VariantPageContentProps = {
     lof_curations?: any[]
     in_silico_predictors?: any[]
     non_coding_constraint: NonCodingConstraint | null
-    transcript_consequences?: any[]
+    transcript_consequences?: TranscriptConsequence[]
```
phildarnowsky-broad, line 89: @rileyhgrant GraphQL actually returns `null` not `undefined` for missing values like this, so `transcript_consequences: TranscriptConsequence[] | null` is what you want here, assuming that it is valid to be missing `transcript_consequences`.

browser/src/VariantPage/VariantPage.tsx
```diff
@@ -506,7 +518,27 @@ type VariantPageProps = {
   variantId: string
 }
 
+const checkGeneLink = (variant: any) => {
```
phildarnowsky-broad, line 521: let's put some proper types rather than `any` for this and `transcript` below.

browser/src/VariantPage/VariantPage.tsx
```diff
@@ -583,20 +624,22 @@ const VariantPage = ({ datasetId, variantId }: VariantPageProps) => {
...
+                      <Button
+                        onClick={() => {
+                          navigator.clipboard.writeText(variantId)
+                        }}
+                        style={{ margin: '0 1em' }}
+                      >
+                        Copy variant ID
+                      </Button>
+                    )}
+                    <GeneLink to={`/gene/${gene.ensembleId}`}>{`⬉ ${gene.symbol}`}</GeneLink>
```
phildarnowsky-broad, line 638: ah OK the icon here is similar to the one I had in mind meaning "external link"

browser/src/VariantPage/VariantPageTitle.tsx
```diff
@@ -88,6 +98,10 @@ const VariantPageTitle = ({ datasetId, variantId }: Props) => {
       </TooltipAnchor>
       <Separator> </Separator>
       <span>({referenceGenome(datasetId)})</span>
+      {/* <span>{','}</span>
```
phildarnowsky-broad, line 101: this can come out

phildarnowsky-broad (changes requested):
couple very small changes requested

mattsolo1:
Can't say I'm a huge fan of the link. How about making it a button? This makes it stylistically consistent with the surrounding buttons.

image

rileyhgrant:
That's fair @mattsolo1. Though, a button's extra margins become problematic on Single Nucleotide Variants and multi base insertions.

Screenshot 2023-02-03 at 14 11 42

Is the whole fully spelled out SNV text necessary? Or could we move to a 'SNV' with a hover text that's the full thing. I'd imagine anyone browsing gnomAD variant pages probably knows what an SNV is. 

It seems the behavior of adding an ellipses to multi base insertions or deletions that are too long exists as an edge case, but that doesn't seem like the thing to do for every single SNV.


Just Gene symbol gets it short enough to do elipses
Screenshot 2023-02-03 at 14 21 59


Sample multi base insertion with elipses on prod:
Screenshot 2023-02-03 at 15 03 28

mattsolo1:
I don't think it's advisable to truncate SNVs ids with ellipses. I do think we could make a case for the SNV abbreviation w/ hover to see the full thing.

phildarnowsky-broad:
> I don't think it's advisable to truncate SNVs ids with ellipses. 

I concur--this is a place where small details matter a lot and shouldn't be glossed over.

mattsolo1:
Heidi says we can definitely shorten to SNV, and that button should say "Go to gene page" or maybe we can get away with just "Gene page". The word "page" being important so that people don't think it is an external resource being linked to. Will that work?

rileyhgrant:
> Heidi says we can ...

Yep that all makes sense - thanks for checking on that and following up!

phildarnowsky-broad (approved):

PR #1060, opened 2023-01-10
Dual rare variants table on gene page

phildarnowsky-broad:
As illustrated in the attached screenshots, a user can now switch the table in the upper-right-hand side of the gene page to show variant cooccurrence count data.

Screen Shot 2023-01-09 at 3 03 55 PM
Screen Shot 2023-01-09 at 3 03 58 PM


rileyhgrant (changes requested):
Looks good to me! With the caveat of I didn't poke around myself yet on a localhost site, this is based on your screenshots and what I could glean from reading through the code.

I assume that I can poke around, but with some adjustments to which backend I point at (i.e. the phil gnomadev cluster). The phil gnomadev cluster with the test data loaded appears to be down for the moment (which makes sense as its not in use) and this means I was unable to poke around myself at this moment.

The only reason I marked as 'Request changes' is because the data is not yet ready for production, so I think it makes sense to make clearly mark that, and make it easy to come back to later and address a few lines of code in the pipeline when the data is finalized, and ready to be run on prod.

The code looks good to my eyes! The testing in particular is quite cool.

yarn.lock
```diff
@@ -8911,9 +8911,9 @@ prettier-linter-helpers@^1.0.0:
     fast-diff "^1.1.2"
 
 prettier@^2.0.5:
-  version "2.8.2"
-  resolved "https://registry.yarnpkg.com/prettier/-/prettier-2.8.2.tgz#c4ea1b5b454d7c4b59966db2e06ed7eec5dfd160"
-  integrity sha512-BtRV9BcncDyI2tsuS19zzhzoxD8Dh8LiCx7j7tHzrkz8GFXAexeWFdi22mjE1d16dftH2qNaytVxqiRTGlMfpw==
+  version "2.7.1"
```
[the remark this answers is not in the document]
phildarnowsky-broad: this change is most likely an accidental leftover of some of the elaborate rebase dances I was doing right before I opened this PR, so most likely prettier can in fact stay at 2.8.2. will check.

graphql-api/src/graphql/types/gene.graphql
```diff
@@ -55,6 +86,7 @@ type Gene {
   mitochondrial_coverage(dataset: DatasetId!): [MitochondrialCoverageBin!] @cost(value: 5)
 
   short_tandem_repeats(dataset: DatasetId!): [ShortTandemRepeat!]! @cost(value: 5)
+  variant_cooccurrence_counts: VariantCooccurrenceCountsPerSeverityAndAf! @cost(value: 5)
```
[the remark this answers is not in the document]
phildarnowsky-broad: this is something I ought to revisit, because the current version of this imports much less data than my earlier cuts at it (well under half of the original amount), so I suspect that cost might be higher than is appropriate now. I don't know how you'd pick a good cost based on first principles, so my strategy is to see what costs other fields of comparable size have.

browser/src/GenePage/VariantCooccurrenceCountsTable.tsx
```diff
@@ -0,0 +1,390 @@
...
+        severity={severity}
+        afCutoff="af_cutoff_0_0001"
+      />
+    </>
+  )
+}
+
+const ConsequenceHeaderContent = () => {
+  const tooltip =
+    'The displayed consequence is the least severe consequence of both variants in the variant pair (e.g., LoF + moderate missense also includes all individuals with LoF + strong missense and LoF + LoF). Missense category is determined by REVEL score (strong ≥ 0.932, moderate ≥ 0.773, supporting ≥ 0.644) in line with ClinGen recommendation: www.biorxiv.org/content/10.1101/2022.03.17.484479v1).'
```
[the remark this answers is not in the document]
phildarnowsky-broad: most likely--I will check with Sarah just to be sure that we do want a link to the latest rather than to v1 specifically.

data-pipeline/src/data_pipeline/pipelines/genes.py
```diff
@@ -181,6 +185,10 @@
     {"path": "gs://gcp-public-data--gnomad/release/2.1.1/constraint/gnomad.v2.1.1.lof_metrics.by_transcript.ht"},
 )
 
+pipeline.add_task(
+    "prepare_variant_cooccurrence_counts", prepare_variant_cooccurrence_counts, "/variant_cooccurrence_counts.ht"
```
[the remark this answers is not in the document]
phildarnowsky-broad: yes that seems like it'd be good just for neatness

phildarnowsky-broad:
@rileyhgrant 

> I assume that I can poke around, but with some adjustments to which backend I point at (i.e. the phil gnomadev cluster). The phil gnomadev cluster with the test data loaded appears to be down for the moment (which makes sense as its not in use) and this means I was unable to poke around myself at this moment.

I'm planning (pending the changes I'm awaiting that I mention elsewhere) to spin up a demo cluster for QA, will make sure to let you know when I do so you can see for yourself.

data-pipeline/src/data_pipeline/pipelines/variant_cooccurrence_counts.py
```diff
@@ -1,5 +1,4 @@
 import hail as hl
-from data_pipeline.pipeline import Pipeline, run_pipeline
 
 RAW_DATA_PATH = "gs://gnomadev-data-pipeline-output/pwd-2022-12-06/external_sources/chet_per_gene.tsv"
```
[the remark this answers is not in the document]
mattsolo1: Not sure if I'm following this conversation exactly, but yes in general it's often very useful to track whole process for how data got into our pipeline. Even if you do copy it to a bucket we control, perhaps write in a comment how it was produced or where it originally came from.

browser/src/GenePage/VariantCooccurrenceCountsTable.tsx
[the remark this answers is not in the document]
phildarnowsky-broad: link to the latest is good, confirmed

phildarnowsky-broad:
Got the new data along with some minor requests for tweaks (mostly just content).

phildarnowsky-broad:
One thing that seems worth mentioning too, is that you'll notice that the code to mock `Query` and `BaseQuery` is kind of clunky. It turns out that Jest does some things behind the scenes that make it necessary to duck and dive like that. Since I'm re-using this technique in another branch I'm working on in a number of places, I'm looking for ways to neaten that and eliminate boilerplate.

Longer-term, I'd like to refactor a lot of that querying code itself: a lot of it was apparently written by straight copy-and-paste, so we have if nothing else the opportunity to do some simple but worthwhile refactorings.

phildarnowsky-broad:
@rileyhgrant @mattsolo1 this should bring us up to date with the earlier round of comments, as well as the changes requested by Sarah (see https://github.com/orgs/broadinstitute/projects/29/views/3)

phildarnowsky-broad:
FYI @mattsolo1 @rileyhgrant I just got feedback from Kaitlin and Sarah, it looks like all basic content changes, nothing that would change the logic in any meaningful way, so still OK to review just what's here now. Then there will be some trivial-to-small fixup commits in the near future for those QA changes.

browser/src/types.d.ts
```diff
@@ -14,6 +35,83 @@ export type GeneMetadata = {
...
+
+export type HeterozygousVariantCooccurrenceSeverity = typeof heterozygousVariantCooccurrenceSeverities[number]
+
+export type HeterozygousCountCellSchema = {
+  in_cis: number
+  in_trans: number
+  unphased: number
+  two_het_total: number
+}
+
```
rileyhgrant, line 89: Continued discussion from office hours, this may be a good time to start breaking these type definitions into slightly more scoped files. A file like this in the Gene page directory with only the types that are relevant there might help with potential name collisions, and slow the rapid growth of this global types file.

Continuing to add to `types.d.ts` seems like it's going to cause problems sooner rather than later. Though I know there are some types that are more widespread than just in Genes, some amount of tighter scoping may start to be useful.

data-pipeline/src/data_pipeline/pipelines/genes.py
```diff
@@ -309,13 +356,15 @@ def annotate_with_preferred_transcript(table_path):
...
-        "genes_grch38": "annotate_grch38_genes_step_3",
+        "genes_grch37": "annotate_grch37_genes_step_5",
+        "genes_grch38": "annotate_grch38_genes_step_4",
         "base_transcripts_grch37": "extract_grch37_transcripts",
         "base_transcripts_grch38": "extract_grch38_transcripts",
         "transcripts_grch37": "annotate_grch37_transcripts",
         "transcripts_grch38": "extract_grch38_transcripts",
         "mane_select_transcripts": "import_mane_select_transcripts",
+        "heterozygous_variant_cooccurrence_counts": "prepare_heterozygous_variant_cooccurrence_counts",
+        "homozygous_variant_cooccurrence_counts": "prepare_homozygous_variant_cooccurrence_counts",
```
rileyhgrant, line 367: Are these anticipated to be referenced elsewhere? If they're only going to be used to annotate the final `genes` hailtables, I believe these aren't needed as pipeline outputs.

Setting pipeline outputs is used so that `export_to_elasticsearch` can find the location of the hailtables to export.

data-pipeline/src/data_pipeline/pipelines/variant_cooccurrence_counts.py
```diff
@@ -0,0 +1,152 @@
+import hail as hl
+
+TWO_HET_DATA_PATH = (
+    "gs://gnomadev-data-pipeline-output/pwd-2022-12-06/external_sources/chet_unphased_same_hap_per_gene.tsv"
+)
+HOMOZYGOUS_DATA_PATH = "gs://gnomadev-data-pipeline-output/pwd-2022-12-06/external_sources/het_hom_per_gene.tsv"
```
rileyhgrant, line 6: If this is close to going to prod, maybe now is the time to discuss where we want to keep additional data sources that we may or may want to have in a bucket that we can maintain ownership of.

Currently, `/external_sources/` houses data like gencode, gtex, and clinvar. Those types of data feel like a different definition of external sources as opposed to things like variant cooccurence, genomic constraint, or other 'secondary analyses' type datasets.

Is it worth it to make a bucket to house secondary analyses type datasets so that we always at least have one (potentially outdated copy) of the input data? @mattsolo1, thoughts?

browser/src/TeamPage/__snapshots__/TeamPage.spec.tsx.snap
```diff
@@ -920,7 +920,7 @@ exports[`Team Page has no unexpected changes 1`] = `
               aria-labelledby="website-staff"
               className="c15"
               dangerouslySetInnerHTML={
-                Object {
+                {
                   "__html": "- **Matthew Solomonson**
 - Phil Darnowski
 - Riley Grant
```
rileyhgrant, line 926: More general, but there will be some merge conflicts when rebasing this off main. 

I have that dang Downloads page PR still open (I was hoping to wait for the additional dataset to be ready to add that in) that may also cause merge conflicts with this updating `jest` and its snapshots commit here. I could wait on that if this is close to going to prod, otherwise, there may be a bit of rebasing in your future, just fyi.

browser/src/GenePage/GenePageContainer.tsx
```diff
@@ -8,8 +8,31 @@ import StatusMessage from '../StatusMessage'
 
 import GeneNotFound from './GeneNotFound'
 import GenePage from './GenePage'
+import {
+  heterozygousVariantCooccurrenceSeverities,
+  heterozygousVariantCooccurrenceAfCutoffs,
+  homozygousVariantCooccurrenceSeverities,
+  homozygousVariantCooccurrenceAfCutoffs,
+} from './VariantCooccurrenceCountsTable'
```
rileyhgrant, line 16: Nothing necessarily actionable on this comment, so really just a comment, but having to import things from the particular component used within the gene page feels a bit odd.

I get why it's happening, that the container is where the query happens and the query is so unreasonably large that you map over those values to construct it instead of typing out however many lines, but it is a bit odd. 

I haven't looked as closely at the data, so I could very easily be missing why this doesn't work, but it seems as though there's an implementation where instead of however many distinctly named fields you manually ask for from GraphQL, arrays could be used with another field or two representing the cutoff and/or severity. That way the query could be comprehensible and wouldn't require programmatic creation.

rileyhgrant (changes requested):
Overall lookin' pretty good.

I added some comments, some actionable and specific, some more high level (google bucket storage for secondary analyses datasets, thoughts on graphql data model).

The actual small requested changes are pretty straightforward, but the more open ended questions might warrant discussion with the team.

Feel free to ping me if discussion in a more high bandwidth way is desired!

graphql-api/src/graphql/types/gene.graphql
```diff
@@ -18,6 +18,59 @@ type ManeSelectTranscript {
...
+
+type HomozygousVariantCooccurrenceCountsPerSeverityAndAf {
+  lof: HomozygousVariantCooccurrenceCountsPerAf!
+  strong_revel_missense_or_worse: HomozygousVariantCooccurrenceCountsPerAf!
+  moderate_revel_missense_or_worse: HomozygousVariantCooccurrenceCountsPerAf!
+  supporting_revel_missense_or_worse: HomozygousVariantCooccurrenceCountsPerAf!
+  missense_or_worse: HomozygousVariantCooccurrenceCountsPerAf!
+  synonymous_or_worse: HomozygousVariantCooccurrenceCountsPerAf!
+}
+
```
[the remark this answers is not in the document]
phildarnowsky-broad: Yeah this is a good call, and also a good use of code review.

data-pipeline/src/data_pipeline/pipelines/variant_cooccurrence_counts.py
phildarnowsky-broad: I see what you're saying but in that case, seems like it would make a difference here that I got this data from Sarah and a pipeline that she maintains, rather than one under our purview. So to me that makes the TSV files feel like "primary" data, from our point of view at least. But I don't have a strong opinion on this point and am interested in @mattsolo1 's point of view.

browser/src/TeamPage/__snapshots__/TeamPage.spec.tsx.snap
phildarnowsky-broad: Thanks for the heads-up but the chance (or even likelihood) of a merge conflict isn't a big problem here, definitely not grounds to hold back your branch from `main` if it's ready. In a simple case like this, when you rebase and get the merge conflict, it's generally pretty clear what happened and how to fix it, so small conflicts like this are just speedbumps.

That just goes for small merge conflicts, though. Before git caught on, probably the most popular version control system was Subversion, and its manual had something about this that has stuck with me. They said something to the effect that a good merge algorithm is not as important as communication. That is, if you're frequently having trouble resolving large merge conflicts, you don't have a code problem that you fix with software, you have a team and process problem that you fix with communication.

browser/src/GenePage/GenePageContainer.tsx
phildarnowsky-broad: I'm thinking about this question more, and I don't really see what's so bad about big GraphQL queries. @mattsolo1 what was your concern about them?

data-pipeline/src/data_pipeline/pipelines/variant_cooccurrence_counts.py
rileyhgrant: Ah, ok I think I wasn't very articulate. I bring this up 'cause it relates to the distinction both between Core Dataset and Secondary Analyses, and on our having a bucket to maintain control of data. So I think there are two points here to be addressed:

1. Do we want to have this data in a bucket that we control. Should we copy it to be also located somewhere in `gnomad-browser-data-pipeline`, in addition to the original place where we got it, or should we not copy it to our bucket at all and only reference the original source. I bring this up because I believe you had expressed interest in having our own copies of all input data.

2. If we want to copy it to be also in a bucket we control, which bucket is that. In this instance, going along with the downloads reformatting, there is a distinction between core dataset and secondary analyses, that distinction being who produces/maintains the dataset. If it is the methods team and thus the data is a part of a gnomad release, then it is considered core dataset. If not, it is a secondary analyses. Genomic Constraint, for example, is a dataset from Siwei. While its under some elses purview and produced from a pipeline they maintain, it is not part of the core gnomad dataset the methods team released, and considered 'secondary analyses' on the downloads page. I was curious if mirroring this structure could help keep the GCS bucket neat and tidy, as the current `external_sources` bucket seems to be distinct from this "datasets created using gnomad data", and this dataset also doesn't fit neatly in `gnomad_v2` or some bucket like that.

browser/src/GenePage/GenePageContainer.tsx
mattsolo1: The debate here is whether this data is better represented in a "long" or "wide" format.

See: https://www.statology.org/long-vs-wide-data/

The wide format is easier to understand from a human-readable standpoint but more complex from a computer-readable standpoint because now we need to keep track of so many column names in the code.

Phil has chosen a very wide format, which means we need to name every column we want in the query. In fact, there are so many columns that we now need to programmatically generate the query.

So what is the actual use case for needing to query columns in this way? I think Phil previously said that the data is already "baked" by the pipeline, so it's ready to use in the app & and perhaps by API users who are interested in this data. This likely simplifies client-side code, since the work of shaping the data into the format you're looking for has already been done.  

However, since you're baking the data in way that is ready for UI display, you're in a sense of moving display logic into the pipeline step, which actually reduces the utility of the data for API users who may want to use it for some other purpose.

The other use case for Phil's implementation is maybe users want to query specific severity or cutoffs. GraphQL has a better solution for that: arguments (https://graphql.org/learn/queries/#arguments)

Riley said:

> there's an implementation where instead of however many distinctly named fields you manually ask for from GraphQL, arrays could be used with another field or two representing the cutoff and/or severity

Consider:

image

Riley is suggesting a long table approach (i.e. moving the theoretically infinite number of "year" columns under one column called "year"). 

I agree, and I strongly suggest that we revise this PR as such. I know it's a lot of work, but I think it's worth it.

In the gnomAD codebase, the only other part of the GraphQL query that takes the long table approach is the GTEx query, which I think was a mistake, and now we're stuck with this long list of tissue names-as-fields that would be better represented under a variable called "tissue" 

https://github.com/broadinstitute/gnomad-browser/blob/a4c5891399036cf906269356fb96652fc8a891b4/browser/src/GenePage/GenePageContainer.tsx#L85

I'd rather we not repeat this mistake with the dual rare variants feature. I don't think it's purely aesthetic --  it will be much easier to understand and maintain this feature in the long term if we design it right. If we really need to get this feature out quickly for stakeholders let's go ahead and merge but I'd like to revise it as soon as we can.

browser/src/TeamPage/__snapshots__/TeamPage.spec.tsx.snap
mattsolo1: > I was hoping to wait for the additional dataset to be ready to add that in

Why do we need to wait? Couldn't you add it after the fact?

browser/src/types.d.ts
```diff
@@ -74,3 +177,178 @@ export type Transcript = {
 }
 
 export type Strand = '+' | '-'
+
+export type ClinvarVariant = {
```
mattsolo1, line 181: Another case where there are many things here not obviously fitting under the title of this PR that would be better as it's own smaller PR.

data-pipeline/src/data_pipeline/pipelines/variant_cooccurrence_counts.py
mattsolo1: > Is it worth it to make a bucket to house secondary analyses type datasets so that we always at least have one (potentially outdated copy) of the input data?

Yes I think so... and with some sort of version specifier -- maybe this is the date, would be ideal.  

 I'd be happy if we have separate folders that match the naming schema of downloads page.

So would the bucket schema is like `gs://gnomad-browser-data-pipeline/inputs/secondary-analyses/variant-coocurrence/<date>/<file>`?

data-pipeline/src/data_pipeline/pipelines/variant_cooccurrence_counts.py
```diff
@@ -0,0 +1,152 @@
...
+    heterozygous_variant_cooccurrence_counts_path=None,
+    homozygous_variant_cooccurrence_counts_path=None,
+):
+    genes = hl.read_table(genes_path)
+    heterozygous_variant_cooccurrence_counts = hl.read_table(heterozygous_variant_cooccurrence_counts_path)
+    homozygous_variant_cooccurrence_counts = hl.read_table(homozygous_variant_cooccurrence_counts_path)
+
+    # Note that we JSON-encode the counts before adding them to genes here.
+    # This is a workaround for a limitation of export_to_elasticsearch, which
+    # crashes if given a table with too many fields per row. These count
```
mattsolo1, line 138: Your many column implementation required JSON serialization... which to me is another indicator that this is not the best approach.

tests/__helpers__/datasets.tsx
```diff
@@ -8,6 +8,27 @@ export const forAllDatasets = (
   describe.each(allDatasetIds)(contextDescription, tests)
 }
 
+export const forDatasetsMatching = (
```
mattsolo1, line 11: Should be in separate PR

mattsolo1 (changes requested):

browser/help/topics/variant-cooccurrence.md
```diff
@@ -0,0 +1,12 @@
+---
+id: variant-cooccurrence
+title: 'Variant cooccurrence'
+---
+
```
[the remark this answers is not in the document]
phildarnowsky-broad: Makes sense to me, I'll ask Sarah for content on that.

browser/src/GenePage/VariantCooccurrenceCountsTable.tsx
```diff
@@ -0,0 +1,424 @@
...
+  variant_cooccurrence_counts: VariantCooccurrenceCountsPerSeverityAndAf<
+    Severity,
+    AfCutoff,
+    CountCellSchema
+  >
+  severity: Severity
+  afCutoffs: readonly AfCutoff[]
+  dataCellComponent: DataCell<Severity, AfCutoff, CountCellSchema>
+}) => (
+  <>
```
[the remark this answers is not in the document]
phildarnowsky-broad: yes fixed in the stuff I'll be pushing shortly

browser/src/GenePage/GenePage.spec.tsx
```diff
@@ -1,30 +1,142 @@
...
-import { apiCallsMatching } from '../../../tests/__helpers__/apiCall'
-import { forAllDatasets } from '../../../tests/__helpers__/datasets'
+import { forDatasetsMatching, forDatasetsNotMatching } from '../../../tests/__helpers__/datasets'
+import { mockQueries } from '../../../tests/__helpers__/queries'
 import { withDummyRouter } from '../../../tests/__helpers__/router'
 
-forAllDatasets('GenePage with dataset "%s"', (datasetId) => {
-  const fetch = jest.fn(() => {
-    return new Promise(() => {})
+import Query, { BaseQuery } from '../Query'
```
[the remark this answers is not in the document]
phildarnowsky-broad: Sure--if it would help clarify this PR, I could rebase things around so that at least this helper (and the other new one) are in separate commits. Then you could look at each of those commits in isolation, at least.

browser/src/GenePage/GenePageContainer.tsx
phildarnowsky-broad: Yeah, I think I see where you're coming from now. My main concern with the long representation is that (for reasons that take a little explaining, I can if you really want the details) it ends up subverting type safety when you reshape it on client-side into the data structure we'll actually be using there.

Thing about that is, we didn't actually have type safety there anymore to speak of: the JSON encoding trick already blew that for us. While I never really like having to give that up, it's not as big a potential booby trap here as it might be otherwise, since we--this team specifically--control both ends of the transaction, the frontend browser code and the GraphQL server.

Thinking a little more, I think I do perhaps see a way to have both a long format and type safety on the frontend, will investigate.

Also more generally, as I think I remember you saying in the recent past @mattsolo1 , this does make me wonder if GraphQL is the best choice for our API--that's a conversation for another time but one worth having I'd say.

data-pipeline/src/data_pipeline/pipelines/variant_cooccurrence_counts.py
phildarnowsky-broad: > So would the bucket schema is like gs://gnomad-browser-data-pipeline/inputs/secondary-analyses/variant-coocurrence/<date>/<file>?

Works for me--I like the idea of being more explicit about this now that I get what Riley meant.

graphql-api/src/graphql/types/gene.graphql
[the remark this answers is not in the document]
phildarnowsky-broad: Most of this disappeared in recent commits.

browser/src/types.d.ts
phildarnowsky-broad: @mattsolo1 FYI what Riley refers to here is that I said to him earlier that in retrospect a centralized `types.d` was probably a mistake, and it would be better to generally have types live in the code that they're most relevant to.

phildarnowsky-broad:
@rileyhgrant @mattsolo1 there's more to come, but I've pushed enough changes now that they're worth looking at if you're looking for something to do

phildarnowsky-broad:
@rileyhgrant @mattsolo1 also the new changes will most definitely be easier to review if you go through them commit by commit as opposed to trying to take all the changes in at once

browser/src/GenePage/GenePage.tsx
```diff
@@ -117,6 +126,35 @@ const LegendSwatch = styled.span`
...
+const BaseTableSelector = styled.div<TableSelectorProps>
+const TableSelector = BaseTableSelector.attrs(
+  ({ setSelectedTableName, ownTableName }: TableSelectorProps) => ({
+    onClick: () => setSelectedTableName(ownTableName),
+  })
+)`
+  border: 1px solid black;
+  border-radius: 0.5em;
+  cursor: pointer;
+  font-weight: bold;
```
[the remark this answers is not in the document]
phildarnowsky-broad: It looks like `SegmentedControl` can't in fact contain an `InfoButton` (the component that corresponds to those question-mark icons), and it would need significant changes to make it able to. Given that it makes more sense to me to just match the font weight itself for now and revisit `SegmentedControl` later.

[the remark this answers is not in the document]
phildarnowsky-broad: Also the gray background color was slightly off from the dataset button, changed that as well.

browser/src/GenePage/GenePageContainer.tsx
```diff
@@ -250,6 +256,53 @@ type Props = {
   geneIdOrSymbol: string
 }
 
+interface UnrolledVariantCooccurrenceCounts {
```
mattsolo1, line 259: I think you'll want to move this shaping logic elsewhere. Notice no other data is shaped in GenePageContainer. Looks like its main job is to fetch the data. 

To me it makes sense to do this shaping closer to where the data is used, like https://github.com/broadinstitute/gnomad-browser/blob/bfb9489fb10dfd065169090bc2d4c44c845b1ed8/browser/src/GenePage/VariantCooccurrenceCountsTable.tsx#L316

graphql-api/src/graphql/types/gene.graphql
```diff
@@ -18,57 +18,27 @@ type ManeSelectTranscript {
...
-  lof_moderate_revel_missense_or_worse: HeterozygousVariantCooccurrenceCountsPerAf!
-  lof_supporting_revel_missense_or_worse: HeterozygousVariantCooccurrenceCountsPerAf!
-  strong_revel_missense_or_worse_strong_revel_missense_or_worse: HeterozygousVariantCooccurrenceCountsPerAf!
-  strong_revel_missense_or_worse_moderate_revel_missense_or_worse: HeterozygousVariantCooccurrenceCountsPerAf!
-  strong_revel_missense_or_worse_supporting_revel_missense_or_worse: HeterozygousVariantCooccurrenceCountsPerAf!
-  moderate_revel_missense_or_worse_moderate_revel_missense_or_worse: HeterozygousVariantCooccurrenceCountsPerAf!
-  missense_or_worse_missense_or_worse: HeterozygousVariantCooccurrenceCountsPerAf!
-  synonymous_or_worse_synonymous_or_worse: HeterozygousVariantCooccurrenceCountsPerAf!
-  supporting_revel_missense_or_worse_supporting_revel_missense_or_worse: HeterozygousVariantCooccurrenceCountsPerAf!
+type HeterozygousVariantCooccurrenceCounts {
```
mattsolo1, line 32: Much cleaner imho

mattsolo1 (changes requested):
Looking great! Just one request about moving the data shaping logic into the component where it is used.

browser/help/topics/variant-cooccurrence.md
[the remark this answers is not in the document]
phildarnowsky-broad: Still waiting on this, pinged Sarah on it.

data-pipeline/src/data_pipeline/pipelines/genes.py
phildarnowsky-broad: Confirmed, don't need to set these.

browser/src/GenePage/GenePageContainer.tsx
phildarnowsky-broad: @mattsolo1 I want to think about this a little more, but it seems like it might be more advantageous to reshape data closer to the query than the point of use, as a general rule. If I still think the same way on Monday I'll write a fuller explanation.

phildarnowsky-broad:
@rileyhgrant @mattsolo1 the latest commits, none of them large, should cover all the feedback I have to date, except for the question of data shaping as noted.

browser/src/GenePage/GenePageContainer.tsx
phildarnowsky-broad: @mattsolo1 I recommend that we start reshaping our data, when necessary, at the first opportunity once we have it--in other words, right after getting it back from the API. The essential reason is that that lets us keep the representation of that data, to the maximum degree possible, consistent throughout the frontend tier.

rileyhgrant (approved):
Code lookin' good to me!

Just one more question on process for my curiosity/best practices:

Anything marked `DONTMERGE` is clearly not to be merged. In terms of a PR like this where those are still included in the commits that are to be approved, is it fair to just assume those will be dropped in an interactive rebase by the PR opener before the commits are squashed into a concise amount, them finally merged into main?

I guess in a more general sense, is it better to rebase and drop those commits before asking for review on a PR (I know in this case the PR was open then got a round of revisions which necessitated further commits) or to include all of it and flag certain commits with `DONTMERGE` for reviewers' and submitters' sake down the line.

phildarnowsky-broad:
> In terms of a PR like this where those are still included in the commits that are to be approved, is it fair to just assume those will be dropped in an interactive rebase by the PR opener before the commits are squashed into a concise amount, them finally merged into main?

Exactly. As a reviewer you can usually assume that a `DONTMERGE` contains some kind of temporary change needed to make the code work in dev specifically as opposed to prod, so it's generally safe to give them only the quickest skim necessary to make sure you understand the intent of that commit.

browser/src/GenePage/GenePageContainer.tsx
phildarnowsky-broad: @mattsolo1 are you cool with that? I can explain my reasoning in more detail if you want.

mattsolo1: That's fine. I'm not aware of too many instances when data must be shaped high in the component tree and then is used in several different places that would benefit from that shaping. I don't know how much it matters to have strict policies about this, I just thought the previous shaping logic looked out of place where it was in the gene page container. It may be a case-by-case basis, where sometimes it makes sense to shape right when we get it, and in other cases shape it in the component if the shape is strongly coupled to a particular view or something.

PR #1072, opened 2023-01-31
Fix both LC OS flags displaying on the Variant tables

### LINKED ISSUE #1070: Bug: Variant table incorrectly also displays LC when there is an OS consequence
Related slack thread/discussion: https://atgu.slack.com/archives/CNL7NA6H2/p1668446829531889

Currently, if a variant has an Other Splice Loss-of-Function (`OS pLoF`) flag as a consequence, in the table view (Gene or Region page) it displays both a Low Confidence Predicted Loss-of-Function (`LC pLoF`) flag as well as an `OS pLoF` flag. On the page for one such Variant, it correctly only shows `OS pLoF` as the the consequence. 

Other Splice is rare enough, and the confidence that annotation is low enough (even lower than Low Confidence Predicted Loss-of-Function, `LC pLoF`) that the browser should not flag these as `LC pLoF` in the table. The Variant page displays the correct behavior.

Konrad said in the Slack Thread:

> Other Splice is rare enough and our confidence in it is low enough that we haven't had a colored circle there historically

A "colored circle" on a Variant Page's VEP Consequences section is indicative of the confidence in the annotation, thus I believe this quote is more generally interpreted as:

> Other Splice is rare enough and our confidence in it is low enough that we haven't [even] had a [low confidence flag associated with it] historically

The correct behavior is to only show `OS pLoF` in the Variant table. The logic for displaying flags in the table view needs to be fixed.

---

Screenshot of current behavior:
Screenshot 2023-01-31 at 11 19 15

rileyhgrant:
Resolves #1070 

Adds an additional branch of to the flag display logic to correctly show only the `OS pLoF` on the Variant table of the Gene or Region page when the variant's most severe consequence is `OS pLoF`.

Screenshot:
Screenshot 2023-01-31 at 11 24 10

Localhost link to Region that shows the change: http://localhost:8008/region/1-63868015-63868025
(The fix is in the API, so to confirm locally you should modify `Query.tsx` to point to localhost for the API, and start a local API)

### DIFF
--- a/browser/src/VariantList/VariantFlag.tsx
+++ b/browser/src/VariantList/VariantFlag.tsx
@@ -35,7 +35,7 @@ export const FLAGS_CONFIG: Record<string, Flag> = {
     label: 'OS pLoF',
     level: 'info',
     formatTooltip: () =>
-      'Variant predicted to create or disrupt a splice site outside the canonical splice site (beta)',
+      'Other Splice Predicted Loss-of-Function: this variant is predicted to create or disrupt a splice site outside the canonical splice site (beta)',
   },
   mnv: {
     label: 'MNV',
--- a/graphql-api/src/queries/variant-datasets/shared/flags.js
+++ b/graphql-api/src/queries/variant-datasets/shared/flags.js
@@ -22,7 +22,11 @@ const getFlagsForGeneContext = (variant, geneId) => {
     // than a non-coding transcript with a pLoF VEP annotation. LOFTEE does not annotate
     // non-coding transcripts. Check for mostSevereConsequenceInGene.lof to avoid showing an
     // LC pLoF flag next to a non-pLoF VEP consequence.
-    if (!lofConsequencesInGene.some((csq) => csq.lof === 'HC') && mostSevereConsequenceInGene.lof) {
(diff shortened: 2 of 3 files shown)


rileyhgrant:
On the Gene page, the variant table results are stored in the Redis cache for a week. I tested the fix by removing the caching step in a local environment, and the expected behavior happened (only `OS pLoF`, no `LC pLoF`). In addition, the `Jest` test for `flags.js` has been updated to confirm the behavior. 

Is this acceptable? To have this PR be reviewed, and possibly merged, and just let the Redis cache expire within the week to fix the final bug that's currently on live? I could clear the Redis cache if needed, but it seems unnecessary.

mattsolo1 (changes requested):
The tooltip currently says:

> Variant predicted to create or disrupt a splice site outside the canonical splice site (beta)

To match the flag, "OS pLoF", the phrase "Other Splice Predicted Loss-of-Function" should be added to the beginning to the tooltip. At first I thought it meant "outside splice", until I read the issue.

Otherwise lgtm.

mattsolo1 (approved):

PR #1073, opened 2023-02-02
Try esbuild-loader to reduce Webpack compilation time

mattsolo1:
Adding this Webpack loader ([esbuild-loader](https://github.com/privatenumber/esbuild-loader)) significantly reduces Webpack compilation time. For me, running `yarn start` in the `browser` directory now takes 15s to compile rather than 1 min.

### DIFF
--- a/browser/package.json
+++ b/browser/package.json
@@ -50,6 +50,7 @@
   },
   "devDependencies": {
     "cypress": "^10.4.0",
+    "esbuild-loader": "^2.21.0",
     "jest-styled-components": "^6.0.0",
     "ts-migrate": "^0.1.30",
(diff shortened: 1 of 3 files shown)


rileyhgrant:
Holy moly this is super great.

I experienced similarly huge reductions in compilation time to Matt. On my machine:
- `yarn start` went from ~45s -> ~12s
- `yarn build` went from ~60s -> ~30s

Crazy that a single plugin can give 4x faster compilation! Very nice find.

rileyhgrant (approved):
LGTM!

I checked out the commit and poked around pretty much every combination of page and dataset I could think of. The dynamic pages and the static pages all seemed stable.

PR #1074, opened 2023-02-02
Allow Gene pipeline to output a smaller dataset

### LINKED ISSUE #1064: Allow Gene pipeline to produce smaller test datasets
Related: #1042 

In order to more easily create dev environments to allow for testing of changes in a safer env, we should work towards modifying all the pipelines to add the option of outputting a smaller but replicable dataset.

The genes pipeline is a good place to start, as it is required to have the output of this pipeline to run other pipelines. We should add functionality to create a smaller resultant dataset including some pre-specified genes, as well as a seeded random subset of the total set of genes.

Having Genes as an example will set up the changes needed to the pipeline, as well as provide a reference for how to modify all further pipelines.

rileyhgrant:
Will eventually resolve #1064 

This is very in progress, it is not yet functional. I think.

---

Extends the Pipeline to allow it to receive an additional flag per Matt's suggestion. If given this flag, any task that is flagged as subset-able (happy to change terminology) will be passed a boolean value that changes the Tasks resultant data.

Modified the Genes pipeline and tasks within the Genes pipeline to create a smaller dataset.

WIP. Need to have all annotation and steps that start from their own tables (Genes, Canonical Transcripts, Pext, etc) trim down to the same smaller set of data.

### DIFF
--- a/data-pipeline/src/data_pipeline/data_types/seeds.py
+++ b/data-pipeline/src/data_pipeline/data_types/seeds.py
@@ -8,6 +8,6 @@ INTEGER = 1337
 #             PCSK9,             BRCA2,             DMD,               TTN
 GENES_SET = {"ENSG00000169174", "ENSG00000139618", "ENSG00000198947", "ENSG00000155657"}
 
-#TODO: There could be some sort of manual, and large (100+) set of genes such that
+# TODO: There could be some sort of manual, and large (100+) set of genes such that
 #   you never have to rely on ht.sample(n, m) - sample results in a random set of genes
 #   and you need to have annotations in the tables for the same set of genes across everything
--- a/data-pipeline/src/data_pipeline/data_types/seeds.py
+++ b/data-pipeline/src/data_pipeline/data_types/seeds.py
@@ -1,0 +1,13 @@
+# This file consists of shared variables to be used by various pipeline
+#   functions if given the '--create-test-datasets' flag.
+
+SUBSAMPLE_FRACTION = 0.01
+INTEGER = 1337
+
+# The list of genes that should be included overall
+#             PCSK9,             BRCA2,             DMD,               TTN
+GENES_SET = {"ENSG00000169174", "ENSG00000139618", "ENSG00000198947", "ENSG00000155657"}
+
+#TODO: There could be some sort of manual, and large (100+) set of genes such that
+#   you never have to rely on ht.sample(n, m) - sample results in a random set of genes
+#   and you need to have annotations in the tables for the same set of genes across everything
--- a/data-pipeline/src/data_pipeline/pipelines/genes.py
+++ b/data-pipeline/src/data_pipeline/pipelines/genes.py
@@ -158,15 +158,25 @@ pipeline.add_task(
 #     {"tmp_path": "/tmp"},
 # )
 
-pipeline.add_task(
-    "prepare_pext",
-    prepare_pext_data,
-    "/pext_grch37.ht",
-    {
-        "base_level_pext_path": "gs://gcp-public-data--gnomad/papers/2019-tx-annotation/gnomad_browser/all.baselevel.021620.ht",
-        "low_max_pext_genes_path": "gs://gcp-public-data--gnomad/papers/2019-tx-annotation/data/GRCH37_hg19/max_pext_low_genes.021520.tsv",
-    },
-)
+# ====================================================================================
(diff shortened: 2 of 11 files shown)


rileyhgrant:
The `canonical_transcripts` step for v3 continues to take quite a long time. This will still output a smaller resultant table, in theory, but 2 hours for that step seems to be against at least part of what this trying to accomplish. I believe this is because both `sample` and `filter` still have to force computation on an entire table to filter out rows that don't pass a given predicate. Filtering over 300 million rows of v3 canonical transcript genomes is takes time no matter how you slice it.

To get both a faster run, and a smaller dataset, is it feasible to move instead towards taking the resultant hailtable of the last previous full pipeline run, and sample that? So instead of adding a flag to a pipeline when running it to create a smaller dataset, the full pipeline gets run once on prod or something and maybe we copy data over, then every time thereafter to get a smaller dataset we run a seperate pipeline to downsample the previous final output?

Just a thought.

data-pipeline/src/data_pipeline/data_types/seeds.py
```diff
@@ -0,0 +1,13 @@
+# This file consists of shared variables to be used by various pipeline
+#   functions if given the '--create-test-datasets' flag.
+
+SUBSAMPLE_FRACTION = 0.01
+INTEGER = 1337
```
mattsolo1, line 5: Nice. 

However, a bit confusing to name this `INTEGER` because someone could think this is a type and not a hardcoded value. Maybe something like `SEED_NUMBER`.

data-pipeline/src/data_pipeline/pipelines/genes.py
```diff
@@ -62,6 +65,10 @@
     "/genes/genes_grch37_base.ht",
     {"gencode_path": pipeline.get_task("download_gencode_v19_gtf"), "hgnc_path": pipeline.get_task("download_hgnc")},
     {"reference_genome": "GRCh37"},
+    # TODO:FIXME:
+    # 6th positional argument is subsettable
+    subsettable=True,
+    # True,
 )
```
mattsolo1, line 73: Starting to wonder if instead of passing subset flags into each pipeline step, we could conditionally insert a new task in front of main task that does the subsetting before passing it on to the main task.

mattsolo1 (commented):
I took a quick look but still need to try running this. We need to should think carefully about how to approach testing pipelines... it's really tricky.

It will probably be helpful for me to show you how I approached things with Genebass, so lets put this on hold till then. But this is a good start -- thank you!

data-pipeline/src/data_pipeline/data_types/seeds.py
phildarnowsky-broad: of all the random things from my youth to survive to this point..."1337" and "leet".

rileyhgrant:
> I took a quick look but still need to try running this. We need to should think carefully about how to approach testing pipelines... it's really tricky.
> 
> It will probably be helpful for me to show you how I approached things with Genebass, so lets put this on hold till then. But this is a good start -- thank you!

Closing this for now. This issue will need to be circled around to later.

Overall, the method in this draft PR takes an additional flag, and uses that flag to filter the input data as part of the pipeline, as early in the process as possible, to try and speed up computation by way of having less data at later steps in the pipeline.

The issues with this are:
- Performing this step on every run of the pipeline when given the new flag is still time consuming
- There are many discrete input data sets, each of which needs to be filtered. Filtering over transcripts, for example, is going to take a long time no matter what.
  - The Genes pipeline, for example, needs to filter the given version hailtable, Gencode, GTEx, MANE transcripts, pext, constraint, etc. All of these need to be kept in sync (filter to the same set of Genes) so that the pipeline works successfully

When circling back to this, a possible solution is as Matt has done with GeneBass
- Create a discrete step that you can perform once (or at the very least, infrequently), this step takes all the input datasets and returns smaller datasets
- Modify the test dataset flag to use these smaller datasets instead

This method would move the computation of reducing the size of the input datasets into a step that is run infrequently, and the result would be much the same as if the pipeline performs those steps every time.

The goal with creating a test datasets flag is twofold:
1. Create a smaller resultant set of data for the browser, allowing a demo instance to be run with a much smaller set of data loaded into the database, while still allowing for good testing (manual, and hopefully automated)
2. Allow the pipelines to be run to completion with far less computational power. Allowing the data pipeline to be debugged with less expense, and a tighter feedback loop

PR #1079, opened 2023-02-15
Add test helpers to test datasets based on name matching regex

phildarnowsky-broad:
This adds two more wrappers for `describe` that streamline running a set of tests over all datasets whose names do/don't match a given regex.


phildarnowsky-broad:
FYI @mattsolo1

rileyhgrant (approved):
LGTM!

PR #1080, opened 2023-02-15
Graphql query helper

phildarnowsky-broad:
This is a query helper to facilitate mocking out GraphQL calls through the `Query`/`BaseQuery` components in our code.

As noted in the code, some things that Jest does behind the scenes mean that we have to do more manual setup than you might want, or in a way that looks odd, and there might be room for further simplification. Feel free to ask if you see anything that looks weird, it's probably that way for a reason.

Also I think the actual usage of this, once you get past the setup, is simple enough, but let me know if you think it might want more documentation.


phildarnowsky-broad:
FYI @rileyhgrant . I asked Matt to review this, but you'll probably want to have a look too, since even if you don't have feedback you may want this helper soon.

browser/src/queryHelperExample.spec.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+// Mock query calls are not cleaned up automatically
+afterEach(() => {
+  resetMockApiCalls()
+  resetMockApiResponses()
+})
+
+// Now for the example code proper.
+
+// This is a simplified version of the typical Query/BaseQuery boilerplate
```
phildarnowsky-broad, line 61: note to self: duplicate comment

mattsolo1 (approved):

PR #1082, opened 2023-02-22
Elasticsearch 7 Compatibility

sjahl:
This includes changes to the elasticsearch deployment, data-pipeline, and graphQL API (thanks @rileyhgrant ) to support elasticsearch 7. The big issue that this fixes is that due to [elastic's deprecation of index document types](https://www.elastic.co/guide/en/elasticsearch/reference/7.17/removal-of-types.html), our elasticsearch_export and api queries no longer needed to nest their idea of the document structure within a type. I've also updated the `node.roles` config here to match the [most recent config syntax](https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-node-configuration.html), since the one we were using previously was deprecated starting with ES7.

There's a few other changes here that aren't strictly related to the v7 upgrade, but I've included them here, since they'll matter for when we move our prod deployment over to an upgraded cluster:

- The disk watermark settings are now applied to all nodeSets. For whatever reason, only specifying the watermarks on the master nodeset wasn't having any effect.
- Various deployments that were pinned to a 'default-pool' GKE pool are now pinned to 'main-pool', since that will be the name in the new deployment
- The non-data elasticsearch pods are now pinned to 'main-pool' instead of only being told _not_ to deploy onto the 'es-data' pool. This will prevent the ES master and coordinating pods from deploying onto the redis pool.
- Disabled the geoip downloader service in elasticsearch. I don't think we were using the geoip database, and it was creating an unneeded index
- Added a service account association to the elasticsearch nodeSets. This will associate them with the service account that has access to write ES snapshots in the new deployment.
- Various formatting fixes applied by the YAML formatter in my text editor

### DIFF
--- a/data-pipeline/requirements.txt
+++ b/data-pipeline/requirements.txt
@@ -1,3 +1,3 @@
-elasticsearch~=6.8
+elasticsearch~=7.17
 hail
 tqdm
--- a/data-pipeline/src/data_pipeline/helpers/elasticsearch_export.py
+++ b/data-pipeline/src/data_pipeline/helpers/elasticsearch_export.py
@@ -128,7 +128,7 @@ def export_table_to_elasticsearch(
 
     # https://www.elastic.co/guide/en/elasticsearch/reference/current/index-modules.html#index-modules-settings
     request_body = {
-        "mappings": {type_name: mapping},
+        "mappings": mapping,
         "settings": {
             "index.codec": "best_compression",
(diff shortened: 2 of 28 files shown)


mattsolo1 (approved):
LGTM! 

(caveat: not testing, but I don't see anything strange).

Excited to finally be on lucky ES version 7.

PR #1085, opened 2023-03-01
Update location of genomic constraint raw data

rileyhgrant:
Updates the location of Siwei's Genomic Constraint Data to `gs://gcp-public-data--gnomad`, where it is now permanently stored.

Previously, this input data was stored in a bucket owned by Siwei. This data was then copied into a bucket in `exac-gnomad` to ensure that the browser would always have access to this data. 

Now, after the discussion with the Production team regarding the Core Dataset / Secondary Analyses dichotomy, the production team is copying Secondary Analyses data into the `gs://gcp-public-data--gnomad/`, where all the Core Dataset files are already stored. 

This PR changes that location on the downloads page, and in the data pipeline. It also removes one field that was previously included in the Genomic Constraint data release. The pipeline was run successfully locally with the new location, and a dev environment was used to confirm that since the frontend never requests the removed field (`coding_prop`), removing this from the API and the pipeline is safe to do.

### DIFF
--- a/browser/src/DownloadsPage/GnomadV3Downloads.tsx
+++ b/browser/src/DownloadsPage/GnomadV3Downloads.tsx
@@ -359,40 +359,28 @@ const GnomadV3Downloads = () => (
         <ListItem>
           <GenericDownloadLinks
             label="README"
-            gcsBucket="gnomad-nc-constraint-v31-paper"
-            path="/download_files/nc_constraint_gnomad_v31_README.docx"
-            includeAWS={false}
-            includeAzure={false}
+            path="/release/3.1/secondary_analyses/genomic_constraint/nc_constraint_gnomad_v31_README.docx"
           />
         </ListItem>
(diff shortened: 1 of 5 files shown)


rileyhgrant:
I didn't check on this as soon as it was in my court,, and the hail team helps the methods team get data synced to Amazon and Microsoft. The data has been synced to the google bucket.

As such, this is going to wait a bit to be merged to until those are synced up so we can include all the links.

rileyhgrant:
The Hail team synced the data to AWS for the methods team, but it is currently not synced to Azure. My understanding is that that step is automatic once the Hail team syncs it to AWS. 

If it's not synced tomorrow morning I'll ping someone and ask about that.

rileyhgrant:
Updated now that I'm back, double checked and the Azure buckets have synced.

mattsolo1 (approved):

PR #1087, opened 2023-03-08
Fuller type checking

phildarnowsky-broad:
It turns out that even with full Typescript support turned on, build tools such as `babel` silently discard TS type errors, and there's no way to get those tools to do otherwise (I can't explain why they do this: it seems like that immediately takes away most of the value of supporting TS in the first place). The recommendation, if you really want to make sure your types are checked in full, is to run a separate pass with `tsc` to explicitly do so.

Here we add some scripts to allow a dev to run those checks easily, and a corresponding CI step. We also fix a handful of type errors that this process turned up: there might have been more of them, but there are several other junctures in our dev workflow where types are checked to some partial extent, so many possible errors were indeed caught at some point.

### DIFF
--- a/browser/src/Query.tsx
+++ b/browser/src/Query.tsx
@@ -37,16 +37,15 @@ const cancelable = (promise: any) => {
   }
 }
 
-type OwnBaseQueryProps = {
-  operationName: string | null
-  query: string
-  url?: string
-  variables?: any
-}
-
 type BaseQueryState = any
 
-type BaseQueryProps = OwnBaseQueryProps & typeof BaseQuery.defaultProps
+type BaseQueryProps = {
+  operationName: string | null
+  query: string
+  url: string
+  variables?: any
+  children: (state: BaseQueryState) => JSX.Element
+}
 
 export class BaseQuery extends Component<BaseQueryProps, BaseQueryState> {
   static defaultProps = {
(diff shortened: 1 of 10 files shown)


browser/src/Query.tsx
```diff
@@ -37,17 +37,16 @@ const cancelable = (promise: any) => {
   }
 }
 
-type OwnBaseQueryProps = {
+type BaseQueryState = any
```
rileyhgrant, line 40: Just for my curiosity and TS education, is the state of a generic `Component<Props, State>` often hard to concretely type, thus `any` is an acceptable type for it? 

I'm curious as to the reasoning behind the `any` here. I have no clue what it would be otherwise, and wondered if you could enlighten me a bit.

rileyhgrant (approved):
LGTM.

Very interesting about `babel` and other tools happening to mitigate some of the usefulness of TS's static typing. Nice catch.

browser/src/Query.tsx
phildarnowsky-broad: In this case, the `any` was inserted automatically by `ts-migrate` a while back, presumably because it wasn't able to figure out a more specific type. States usually are just a typical JS/TS object, nothing exotic, so not especially difficult to write types for. Just haven't touched this one in particular yet.

rileyhgrant: Oh oh right oops. I didn't realize this was Git's diff looking slightly funky (removing the line in one place, and adding it in another place). I see, that makes sense.

PR #1090, opened 2023-03-10
Publications revision

phildarnowsky-broad:
Redirect old citations page to Publications, and overhaul the content for same.


rileyhgrant (approved):
LGTM with one caveat:

The current rather smol Cypress test for the publications page looks for the paragraph text that used to be there, and is now causing CI to fail on that step.

https://github.com/broadinstitute/gnomad-browser/blob/5d00c477e3175f98ff1c5ed2ce052af60d27dec4/browser/cypress/e2e/nav_bar_pages.cy.ts#L55

Could be changed to any text in the opening paragraph (this line changed from "publications by the gnomad group" -> "publications from the gnomad group", causing this test to now fail.)

PR #1105, opened 2023-04-11
Fix v3 subsets on variants page

phildarnowsky-broad:
The variants page was mistakenly trying to load local ancestry populations for all v3 datasets, not just the main v3 dataset. Corrected.


rileyhgrant (approved):
Confirmed this fixed the crash on my localhost. Nice sleuthing. 

LGTM!

PR #1107, opened 2023-04-21
Migrate GraphQL API to TS

phildarnowsky-broad:
Most of the changes fall into two categories:

* Changing from CommonJS-style `require/module.exports` exports to ES6-style `import/export`.
* Adding types, almost all of which are just `any` at this point.

We also now emit CommonJS module code, which both node and the browser can deal with, rather than ES6. In the future, it might be worth configuring things so that the API uses CommonJS and the browser uses ES6.


graphql-api/src/lodash.d.ts
```diff
@@ -0,0 +1,3 @@
+// There's a @types package for lodash but adding it as a dependency
```
mattsolo1, line 1: Strange

mattsolo1 (approved):
Looks great to me! I poked around through various datasets in the online demo and saw no issues.

graphql-api/src/lodash.d.ts
phildarnowsky-broad: could very well be, will try

phildarnowsky-broad: @mattsolo1 ...except that even though it passes CI, now trying to build an image for deployment crashes. Putting this back like it was.

PR #1108, opened 2023-04-24
Set up eslint for better TS linting

phildarnowsky-broad:
Here we add TS plugins for eslint, and lint existing code. In this first lint pass, we mainly just turn off eslint rules, making only those changes to the code that are simplest or most important.

### DIFF
--- a/.github/workflows/graphql-api-ci.yml
+++ b/.github/workflows/graphql-api-ci.yml
@@ -18,7 +18,7 @@ jobs:
       - name: Setup Node
         uses: actions/setup-node@v1
         with:
-          node-version: '14.15.2'
+          node-version: '^14.17.0'
       - name: Get yarn cache
         id: yarn-cache
         run: echo "::set-output name=dir::$(yarn cache dir)"
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
     "typecheck:watch": "yarn run tsc -w --noEmit"
   },
   "engines": {
-    "node": ">=14.15.2 <15.0.0"
+    "node": "^14.17.0 <15.0.0"
   },
   "workspaces": [
     "browser",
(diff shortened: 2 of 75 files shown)


.github/workflows/graphql-api-ci.yml
```diff
@@ -18,7 +18,7 @@ jobs:
       - name: Setup Node
         uses: actions/setup-node@v1
         with:
-          node-version: '14.15.2'
+          node-version: '^14.17.0'
```
rileyhgrant, line 21: If we're updating this, should we update the version of node used in the dockerfile? (`deploy/dockerfiles/browser/api.dockerfile`)

From my perspective, yes. But also, it currently looks like our frontend CI and frontend dockerfile have gotten out of sync, so I'm unsure if this was intentional or it happened at some point.

I would think it's useful to keep these in sync, so suggest updating the API dockerfile in this PR.

package.json
```diff
@@ -18,7 +18,7 @@
     "typecheck:watch": "yarn run tsc -w --noEmit"
   },
   "engines": {
-    "node": ">=14.15.2 <15.0.0"
+    "node": "^14.17.0 <15.0.0"
```
rileyhgrant, line 21: Ah, going along with the previous comment about node versions, perhaps if this is happening in the `package.json`, this is also the time to reconcile the difference in the browser, the dockerfile (`browser.dockerfile`  on `14.15.2`) and the browser CI (`browser-ci.yml`, on `14.19.3`).

Or at the very least, I think, we should file an issue to come and look at keeping these node versions in sync.

rileyhgrant (approved):
LGTM

Woohoo for better linting!

Approved with the stipulation of at least filing an issue to look at the discrepancy in node versions between `package.json`, the dockerfiles for the browser and API, and the CI yml files for the browser and API.

.github/workflows/graphql-api-ci.yml
phildarnowsky-broad: Actually it turns out that the docker image won't build if they're out of sync, which, fair enough. Pushing changes shortly.

package.json
phildarnowsky-broad: I'll change it by hand for now, but what we should really be doing is figuring out how to eliminate the need to keep them in sync. We'd have to figure out if there's a reliable way to map from the version given in `engines` to the proper Dockerhub package.

phildarnowsky-broad: But at least we could probably sync the version up automatically between `package.json` and the two `ci.yml` files.

PR #1113, opened 2023-05-19
Restore stub lodash declaration file

phildarnowsky-broad:
Despite the existing `@types` declaration, building an image for deploy won't work without this stub. Unclear why.


rileyhgrant (approved):
LGTM!

Interesting that it needs the stub

Issue #1126, opened 2023-07-05
Gene search leads to wrong gene


mattsolo1:
Feedback from user:

> ...over 50% of the time I type in a gene name into the gnomAD browser, I end up on the wrong gene. I have to be incredibly careful about pausing as I type and deliberately choosing the right gene that comes up in the search-ahead function or most of the time I will land on the wrong gene. I realize this is partly user error (I assume I am somehow hitting return before I've quite finished typing the full name), but I've never understood why it is more likely to lead me astray than land on the right gene. But here is an example that may give some insight into the challenge. I was typing ABCD1 into the browser and accidentally landed on CX3CL1. I tried to recapitulate how I got there instead of ABCD1, and could see that typing in ABCD leads to CX3CL1 as the first option, not ABCD1. I assume this is because ABCD-3 is an alias of CX3CL1, but I wonder if we should remove the use of aliases in the type-ahead function and either not allow aliases, or only allow them if they fully map to the gene. So for example, only typing in the full ABCD-3 would get you to CX3CL1, not ABCD.

phildarnowsky-broad:
One alternative is that we could keep the aliases, but sort things in the search-ahead dropdown so that the "real" gene always appears first. 

Perhaps as a complement to that we could add, for each search-ahead entry, a brief explanation of why that entry matches the search term. For instance, in this case, the option that was clicked in error might read "CX3CL1 (alias: ABCD-3)". But since that requires the user to read and interpret, rather than just knowing that they can count on a click in a certain spot to link to the gene itself, this approach probably wouldn't suffice by itself.

mattsolo1:
I like the idea of adding the alias info to the search results

mattsolo1:
@phildarnowsky-broad is this something we can QC using our test suite?

phildarnowsky-broad:
@mattsolo1 unless there's something odd about the components in question, should be.

mattsolo1:
Related: https://github.com/broadinstitute/gnomad-browser/issues/1057

phildarnowsky-broad:
It looks like the best way to implement this entails a small extension to GBTK, so I'm going to put a pin in this issue, go write that small extension, and return to this issue once GBTK is updated.

nawatts:
For what it's worth, this extra term query in the ES query for gene search was intended to prioritize genes where the search term matched the primary symbol by increasing that gene's score.
[...]

PR #1127, opened 2023-07-07
Sort gene search results in a more intuitive way

phildarnowsky-broad:
This updates search on a gene symbol such that first come all prefix matches for the query string, and then all others, both groups alphabetized.

For example, suppose we got the gene search query `QWE`, and the genes that came back from the API had these symbols respectively:
* `ABCD2`
* `QWE2`
* `CDE1`
* `ABCD1`
* `QWE1`

The order in which these would appear in the dropdown would then be:
* `QWE1`
* `QWE2`
* `ABCD1`
* `ABCD2`
* `CDE1`

### DIFF
--- a/browser/src/Searchbox.tsx
+++ b/browser/src/Searchbox.tsx
@@ -6,6 +6,7 @@ import styled from 'styled-components'
 import { Searchbox, Select } from '@gnomad/ui'
 
 import { fetchSearchResults } from './search'
+import { DatasetId } from '@gnomad/dataset-metadata/metadata'
 
 const Wrapper = styled.div`
   display: flex;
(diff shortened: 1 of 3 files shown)


phildarnowsky-broad:
@mattsolo1 one thing that occurred to me is that we may want to do this for variant search as well, which I believe is the only other free-text type search we have, WDYT?

mattsolo1 (approved):
> we may want to do this for variant search as well, which I believe is the only other free-text type search we have, WDYT

I think that makes sense, yes.

I tested this PR locally and indeed appears to fix #1126. Note we also had a comment on this issue from Nick this afternoon: https://github.com/broadinstitute/gnomad-browser/issues/1126#issuecomment-1625817877

phildarnowsky-broad:
@mattsolo1 actually, taking a closer look at how the variant search is implemented, it looks like there wouldn't be a corresponding issue after all, so I'm going to call this done and merge it.

PR #1142, opened 2023-07-28
Modify coocurrence table to only render on v2 data

### LINKED ISSUE #1140: Don't show variant cooccurrence table or constraint if not using a v2 dataset
per Anne, "The variant co-occurrence data should not be shown on gnomAD v3 or SV v3 (same as constraint data) as it was generated on v2 (but we will eventually generate it on v3)."

(I assume that applies to Exac as well).

Should be a simple matter of adding another `dataset-metadata` flag and checking that in the appropriate place.

rileyhgrant:
Resolves #1140 

Adds a metadata predicate to only render the variant co-ocurrence table on v2 datasets.

### DIFF
--- a/browser/src/GenePage/VariantCooccurrenceCountsTable.spec.tsx
+++ b/browser/src/GenePage/VariantCooccurrenceCountsTable.spec.tsx
@@ -7,6 +7,50 @@ import {
 } from '../__factories__/VariantCooccurrenceCountsPerSeverityAndAf'
 
 import VariantCooccurrenceCountsTable from './VariantCooccurrenceCountsTable'
+import { forDatasetsMatching, forDatasetsNotMatching } from '../../../tests/__helpers__/datasets'
+
+const v2Regexp = /_r2/
+
+forDatasetsNotMatching(
+  v2Regexp,
+  'VariantCoocurrenceCountsTable with non v2 dataset "%s"',
+  (datasetId) => {
+    test('has no unexpected changes and renders as placeholder text', () => {
+      const heterozygousCounts =
+        HeterozygousVariantCooccurrenceCountsPerSeverityAndAfFactory.build()
+      const homozygousCounts = HomozygousVariantCooccurrenceCountsPerSeverityAndAfFactory.build()
+      const tableContent = render(
+        <VariantCooccurrenceCountsTable
+          datasetId={datasetId}
+          heterozygous_variant_cooccurrence_counts={heterozygousCounts}
+          homozygous_variant_cooccurrence_counts={homozygousCounts}
+        />
+      )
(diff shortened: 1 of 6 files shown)


browser/src/GenePage/VariantCooccurrenceCountsTable.spec.tsx
```diff
@@ -7,13 +7,58 @@ import {
 } from '../__factories__/VariantCooccurrenceCountsPerSeverityAndAf'
 
 import VariantCooccurrenceCountsTable from './VariantCooccurrenceCountsTable'
+import { forDatasetsMatching, forDatasetsNotMatching } from '../../../tests/__helpers__/datasets'
+
+const v2Regexp = /_r2/
+
+forDatasetsNotMatching(
```
phildarnowsky-broad, line 14: Not worth delaying this change for, but at some convenient point we should define a couple of constants like this in a test helper library.

browser/src/GenePage/VariantCooccurrenceCountsTable.spec.tsx
```diff
@@ -7,13 +7,58 @@ import {
...
+      const tableContent = render(
+        <VariantCooccurrenceCountsTable
+          datasetId={datasetId}
+          heterozygous_variant_cooccurrence_counts={heterozygousCounts}
+          homozygous_variant_cooccurrence_counts={homozygousCounts}
+        />
+      )
+      const normalContentFragment = tableContent.asFragment()
+      expect(normalContentFragment.querySelectorAll('p').length).toEqual(1)
+      expect(normalContentFragment.querySelectorAll('table').length).toEqual(0)
```
phildarnowsky-broad, line 31: I like this combo of DOM queries and a snapshot, I had always thought of them as mutually exclusive but thinking about it they have complementary strengths.

phildarnowsky-broad (changes requested):
Looks good far as it goes, but we are also supposed to omit the constraint table for non-v2 datasets. That's on me for not capturing the requirement properly.

rileyhgrant:
> Looks good far as it goes, but we are also supposed to omit the constraint table for non-v2 datasets. That's on me for not capturing the requirement properly.

Unless I'm misunderstanding, I think the constraint table already behaves like that and doesn't require any changes.

https://gnomad.broadinstitute.org/gene/ENSG00000169174?dataset=gnomad_r3

At least, in the sense that it doesn't render any data. The header and ability to swap between the two (which now will both say something to the tune of "sorry this isn't available for this version") are still present on non-v2 datasets, but the actual table and data don't render.

phildarnowsky-broad:
@rileyhgrant right you are, NVM.

phildarnowsky-broad (approved):

PR #1147, opened 2023-08-03
Don't crash Clinvar plot when VEP and Gencode disagree on variant length

phildarnowsky-broad:
In some cases, such as one variant on BRCA1, the data from VEP gives us an HGVSP for a variant that runs past the end of the 3' UTR as specified by Gencode. These were causing the plot of Clinvar variants to crash.

Here we fix that, along with an assortment of off-by-one errors and edge cases.


mattsolo1:
I think there is something wrong with how these frameshift variants are being plotted in BRCA1. Can you take another look?

image

phildarnowsky-broad:
@mattsolo1 I think those variants are OK. Note how the HGVSP ends with `fsTer?`. As a nearby comment puts it:

> Termination site position may be "?" if the new reading frame does not encounter a stop codon. In this case, place the termination site at the end of the transcript

...which doesn't mean we _have_ to display it this exact way, but that's what the code was originally meant to do.

mattsolo1:
> place the termination site at the end of the transcript

Because BRCA1 is on the reverse strand, the "end" is actually at the beginning. So for a variant with HGVSp `p.Tyr1874ArgfsTer`, shouldn't the line should be drawn from amino acid 1874 to the end of the transcript? The plot above shows the line being drawn from amino acid 1874 to the beginning of the transcript.

phildarnowsky-broad:
@mattsolo1 wait, I see the problem now. The direction of the line is actually fine, since the "X" marks the end of a frameshift variant (rather than the beginning like you might assume), and this view uses global rather than strand-relative coordinates, meaning on a negative-strand variant like this the "X" should always be on the left side of the view. However, the _right_ endpoint of the line is in the wrong place, it's at the start of the coding region when it should be 1874 codons into it, i.e. much further to the left.

So, to sum up, left endpoint is fine, direction of the line is fine, but the right endpoint needs to move left a large distance. Should be a quick fix.

phildarnowsky-broad:
Here we are, notice how the variant in question is tiny now. Push to come shortly.
Screenshot 2023-08-14 at 3 42 29 PM

phildarnowsky-broad:
@rileyhgrant neat--I don't think I was even aware of that other bug. Might be a good idea then to have @msingerb verify that this fixes said bug. @msingerb you might be able to verify that from Riley's screenshot above but if you'd like a demo set up to verify it live please let me know.

msingerb:
It looks like its fixed from the screenshot. Thanks for bringing this up @rileyhgrant , I never would have remembered this issue.

phildarnowsky-broad:
@msingerb nice, glad to hear. @rileyhgrant well spotted sir.

mattsolo1 (approved):

phildarnowsky-broad:
@mattsolo1 @rileyhgrant I'd like to merge and deploy this ASAP but unfortunately it looks like I'm likely not going to be online next until Thursday morning at earliest, if one of you could do so before then that would be aces, if not no big deal, I'll take care of it when I get back.

rileyhgrant:
@phildarnowsky-broad I can merge and deploy this tomorrow, and make a corresponding changelog post.

mattsolo1:
Thank you Riley! Indeed let's try to get this merged as part of the next app rollout.

rileyhgrant:
Just to leave a paper trail, I just force pushed to Phil's branch to squash a fixup commit into the previous one. 

I'll shortly be merging and then deploying this fix. Thanks Phil for your work!

PR #1148, opened 2023-08-03
Update bug report template

rileyhgrant:
Adds an open comment (`<!--`) to the bug template to match the close comment at the end of the file. This prevents the comment text from erroneously being present in user submitted bug reports.


elissa-alarmani (approved):
Looks great!

PR #1152, opened 2023-08-09
Add genome build to dropdowns

### LINKED ISSUE #1010: Add genome build in dropdown menu
A user thought it would be useful to display the genome build next to the gnomAD version in the search bar drop down:

image

Modify dropdown to be:

gnomAD v3.1.2 (GRCh38)
gnomAD v2.1.1 (GRCh37)
gnomAD SVs v2.1 (GRCh37)
ExAC (GRCh37)

mattsolo1:
See comment below 👇 https://github.com/broadinstitute/gnomad-browser/pull/1152#issuecomment-1672063250

fixes #1010

### DIFF
--- a/browser/src/DatasetSelector.tsx
+++ b/browser/src/DatasetSelector.tsx
@@ -13,6 +13,7 @@ import {
   labelForDataset,
   hasShortVariants,
   hasStructuralVariants,
+  referenceGenome,
 } from '@gnomad/dataset-metadata/metadata'
 
 const NavigationMenuWrapper = styled.ul`
@@ -123,6 +124,12 @@ const SubNavigationLink = styled.a`
   }
 `.withComponent(Link)
 
+const ItemDescriptions = styled.div`
+  display: flex;
+  flex-direction: row;
+  justify-content: space-between;
+`
+
 const ItemDescription = styled.div`
   margin-top: 0.125em;
   margin-left: 5px;
@@ -130,19 +137,19 @@ const ItemDescription = styled.div`
   opacity: 0.6;
 `
 
+type ChildDataset = {
+  id: string; label: string; url: string; description: string; referenceGenome?: string
+}
+
 type Props = {
   items: (
     | {
-        id: string
-        isActive?: boolean
-        label: string
-        url: string | any
-      }
-    | {
-        id: string
-        isActive?: boolean
-        label: string
-      }
+      id: string
+      isActive?: boolean
+      label: string
+      url: string | any
+      children: ChildDataset[]
+    }
   )[]
 }
 
@@ -378,7 +385,7 @@ class NavigationMenu extends Component<Props, State> {
                   </TopLevelNavigationLink>
                   {/* @ts-expect-error TS(2769) FIXME: No overload matches this call. */}
                   <SubNavigationMenu isExpanded={isExpanded}>
-                    {(item as any).children.map((childItem: any) => (
+                    {(item).children.map((childItem) => (
                       <li key={childItem.id}>
                         <SubNavigationLink
                           data-item={childItem.id}
@@ -391,9 +398,14 @@ class NavigationMenu extends Component<Props, State> {
                           onKeyDown={this.onKeyDownSubMenuItem}
                         >
                           {childItem.label}
-                          {childItem.description && (
-                            <ItemDescription>{childItem.description}</ItemDescription>
-                          )}
+                          <ItemDescriptions>
+                            {childItem.description && (
+                              <ItemDescription>{childItem.description}</ItemDescription>
+                            )}
+                            {childItem.referenceGenome && (
+                              <ItemDescription>Build {childItem.referenceGenome}</ItemDescription>
+                            )}
+                          </ItemDescriptions>
                         </SubNavigationLink>
                       </li>
                     ))}
@@ -438,12 +450,13 @@ const DatasetSelector = withRouter(({ datasetOptions, history, selectedDataset }
         isActive: hasShortVariants(selectedDataset),
         label: labelForDataset(topLevelShortVariantDataset),
         url: datasetLink(topLevelShortVariantDataset),
+        referenceGenome: referenceGenome(topLevelShortVariantDataset),
       },
       {
         id: 'other_short_variant',
         isActive: hasShortVariants(selectedDataset),
         label: 'More datasets',
-        children: [] as { id: string; label: string; url: string; description: string }[],
+        children: [] as ChildDataset[],
       },
     ]
 
@@ -454,6 +467,7 @@ const DatasetSelector = withRouter(({ datasetOptions, history, selectedDataset }
         label: labelForDataset('gnomad_r3'),
         url: datasetLink('gnomad_r3'),
         description: `${sampleCounts.gnomad_r3.total.toLocaleString()} samples`,
+        referenceGenome: referenceGenome('gnomad_r3'),
       })
     }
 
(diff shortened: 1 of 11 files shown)


rileyhgrant:
Hiya, snooping on the PR while I'm looking through issues and PRs.

I will probably bring this up at browser meeting, but the Cypress tests I think are doing more harm than good (my fault for wanting to get those in when I started, without really knowing how to do make them effective). I don't think they're catching anything that the Jest tests aren't already, and right now they just add another step to the CI.

At this point, I'd be in favor of removing the Cypress tests step from the CI, and revisiting some sort of integration test for a happy path with Cypress at a later date. I just figured I'd comment here since that is what's causing this PR to fail currently.

elissa-alarmani:
Currently, users are unaware of the Genome Reference Build they are searching on – whether it's the GRCh37 build or GRCh38 build. For example, gnomAD v2.1.1 uses the GRCh37 build, this goes the same for all datasets before v3.

This pull request aims to providing users with visibility into the reference genome build for each dataset they select. This is added in both the `Searchbox` and the `DatasetSelector` component as seen below.

Screenshot 2023-08-09 at 3 49 30 PM

Screenshot 2023-08-09 at 3 49 19 PM

mattsolo1:
@ch-kr hoping for your feedback on the UI changes specifically

ch-kr:
This is fantastic -- thank you for working on this! A lot of the emails coming into the inbox are users asking questions about why they can't find a variant in one gnomAD version (or why a variant seems to have "disappeared"), and the answer is often because they're searching with coordinates from the wrong genome build.

I really like this addition:
image

I have a quick question on this view:
image

Is "Build" necessary? It might be nice to have the reference build text be consistent between both drop downs (and just be "GRCh37" or "GRCh38")

mattsolo1:
For some reason it looked aesthetically pleasing to have "Build" there because it made the text width more symmetrical with the sample number. We can remove it though and show you how it looks, no problem.

phildarnowsky-broad:
Personally I like the "Build XXXX" as shown, but that's just one man's
opinion.

ch-kr:
ooh, you're right, it does look a bit strange when the build is right aligned and just GRCh37/8. I like the last option (with parentheses). Maybe it's worth posting in gnomad_browser to see if there is a consensus?

mattsolo1:
Sure @ch-kr. 

@elissa-alarmani can you please write a message in the #gnomad_browser channel on ATGU slack summarizing the proposed change, along with a link to this issue?

mattsolo1:
The people have spoken: the one with the parentheses.

browser/src/DatasetSelector.tsx
```diff
@@ -378,7 +385,7 @@ class NavigationMenu extends Component<Props, State> {
                   </TopLevelNavigationLink>
                   {/* @ts-expect-error TS(2769) FIXME: No overload matches this call. */}
                   <SubNavigationMenu isExpanded={isExpanded}>
-                    {(item as any).children.map((childItem: any) => (
+                    {(item).children.map((childItem) => (
```
rileyhgrant, line 388: Nice, nickel and dime-ing our way through getting rid of the `any`s introduced with `ts-migrate` by way of stricter more specific types is always a good thing.

browser/src/Searchbox.tsx
```diff
@@ -76,14 +76,14 @@ export default withRouter((props: any) => {
...
         }}
       >
-        <option value="gnomad_r3">gnomAD v3.1.2</option>
-        <option value="gnomad_r2_1">gnomAD v2.1.1</option>
-        <option value="gnomad_sv_r2_1">gnomAD SVs v2.1</option>
-        <option value="exac">ExAC</option>
+        <option value="gnomad_r3">gnomAD v3.1.2 (GRCh38)</option>
+        <option value="gnomad_r2_1">gnomAD v2.1.1 (GRCh37)</option>
+        <option value="gnomad_sv_r2_1">gnomAD SVs v2.1 (GRCh37)</option>
+        <option value="exac">ExAC (GRCh37)</option>
```
rileyhgrant, line 86: I wonder if having these present in the Searchbox is a bit redundant and too much information.

I am completely on board with the genome build in the `DatasetSelector`, since that is a custom component and can be styled. I think the UI choices for that component (smaller font, grey font color) do a good job at conveying the information while keeping things legible.

Here, since it's just the HTML element, the genome build is way more prominent, and on a few pages, fairly redundant (see screenshots at the end of this comment)

I think in an ideal world, mimicking the `DatasetSelector` component to make a custom component for the `Searchbox` to both make the UI more cohesive, and allow for the genome build to show up only on dropdown, would be ideal. Though, I know that's a bigger lift than this ticket initially suggested. I'm curious as to other opinions here, as I'm a bit biased since I've had a few rounds of comments from people that have made me try to steer clear of redundancy.

---

- Variant pages have the build in parentheses in the header
  - Screenshot 2023-08-10 at 17 59 02


- Gene pages have the genome build present in the top left
	- Screenshot 2023-08-10 at 17 58 52

rileyhgrant (commented):
Nice, looking good code wise! Excellent work Elissa and Matt.

My main comment is about the content itself, I think it may be one worth at least running by KC. The tl;dr is I personally think the build in the `Searchbox` component is a bit large and redundant (the `DatasetSelector` components way of handling it I think is perfect). Unforunately, changing that is a bigger lift as it involves overhauling the `Searchbox` into a styled React component, and handling a whole slew of events à la the `DatasetSelector`.

Since I'm only one opinion of many, I'm just leaving this review as commenting on the PR, and neither approving nor requesting changes on it.

Nicely done jumping into things and getting PRs opened and ready for review, Elissa :)

mattsolo1:
Thanks for the feedback!

> I personally think the build in the Searchbox component is a bit large and redundant

The original request was specifically to have build in the search bar (https://github.com/broadinstitute/gnomad-browser/issues/1010), so we need to keep this. Regarding redundancy (that the info is already visible on the gene/variant pages), the goals are (1) to inform user what dataset they _will_ be searching as opposed to the build of the dataset they have _already_ searched for because it could be completely different and (2) when users start from the home page they have no info about genome build on screen anywhere.

> the UI choices for that component (smaller font, grey font color) do a good job at conveying the information while keeping things legible

This is a good suggestion and what indeed what I would have done for the Search if the component could support it, but decided it wasn't worth the effort. Elissa and I can take another look to see if we want to do this, otherwise maybe it warrants a new ticket.

mattsolo1:
Perhaps we can look at `react-select` library for inspiration: https://react-select.com/home

I see they have a "grouped" option. One idea could be to group datasets by GRCh37 or GRCh38. This would allow us to preferentially show GRCh38 datasets first (which would encourage users to use the latest build, which is what we want). And would address your concern that parentheses add clutter. 

image

@rileyhgrant what do you think about using `react-select` directly in gnomAD browser?

rileyhgrant:
> I see they have a "grouped" option. One idea could be to group datasets by GRCh37 or GRCh38. This would allow us to preferentially show GRCh38 datasets first (which would encourage users to use the latest build, which is what we want). And would address your concern that parentheses add clutter.

Ah neat! This seems like it would be a good middle ground. 

I see now the sticking point of wanting to know the build before searching the first time. I definitely agree that the information should be there, and see how having it only appear in the dropdown does not necessarily address that goal, I am still in favor of that option, whether it be our own component or using some other functionality (`react-select` or just html tags).

I like the idea of grouping them in the dropdown, and I believe the HTML select [has an `optgroup` tag](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/optgroup) that does this as well, so it might not even require adding `react-select`.

mattsolo1:
Oooh amazing idea @elissa-alarmani!

mattsolo1:
@elissa-alarmani can you push this change so we can try locally?

mattsolo1:
OK yeah I love this and think it resolves @rileyhgrant's concerns. Are we OK to merge (once tests pass?)

rileyhgrant (approved):

rileyhgrant:
LGTM! Since there are quite a few individual commits, this PR would be a good fit for the "Squash and merge" option from the Github UI, Elissa.

PR #1157, opened 2023-08-14
Viewing context when searching variant table

### LINKED ISSUE #312: Searching variant table removes context
Searching the variant table filters out variants that don't match the search query. However, once a variant of interest has been located, it's useful to see variants in the surrounding region.

Using the browser's find on page feature in the old browser allowed you to jump to a variant of interest without losing its context. But that's not an option in the new browser because it does not render all variants at once.

elissa-alarmani:
Currently, when a user searches in the Variant table all neighboring rows are removed from the view and only exact search matches are displayed. This removes the context of a variant's position in the table and the surrounding variants.

This pull request creates an option where the user can check off if they would like to preserve the context of the table while searching. When "Preserves context search" is checked off, as the user searches the table jumps to the first result (which is highlighted) of their matched search without removing the rows that aren't a search match. 
 
fixes #312

### DIFF
--- a/browser/src/VariantList/Variants.tsx
+++ b/browser/src/VariantList/Variants.tsx
@@ -60,6 +60,31 @@ const variantsDefaultProps = {
 
 type VariantsProps = OwnVariantsProps & typeof variantsDefaultProps
 
+export function getFirstIndexFromSearchText(
+  searchFilter: VariantFilterState,
+  variantSearched: Variant[],
+  variantsTableColumns: any,
+  variantWindow: number[],
+) {
+  const searchedVariants = getFilteredVariants(
+    searchFilter,
+    variantSearched,
+    variantsTableColumns
+  )
+
+  if (searchedVariants.length > 0) {
+    const firstVariant = searchedVariants[0]
+    const firstIndex = variantSearched.findIndex(
+      (variant: Variant) => variant.pos === firstVariant.pos
+    )
+    if (variantWindow[0] !== null && firstIndex < variantWindow[0]) {
+      return firstIndex - 10
+    }
+    return firstIndex + 10
+  }
+  return 0 
+}
+
 const Variants = ({
   children,
   clinvarReleaseDate,
@@ -211,41 +236,16 @@ const Variants = ({
     table.current.scrollToDataRow(index)
   }, [positionLastClicked]) // eslint-disable-line react-hooks/exhaustive-deps
 
-  function getFirstIndexFromSearchText(
-    searchFilter: VariantFilterState,
-    variantSearched: Variant[],
-    variantsTableColumns: any
-  ) {
-    const searchedVariants = getFilteredVariants(
-      searchFilter,
-      variantSearched,
-      variantsTableColumns
-    )
-
-    if (searchedVariants.length > 0) {
-      const firstVariant = searchedVariants[0]
-      const firstIndex = renderedVariants.findIndex(
-        (variant: Variant) => variant.pos === firstVariant.pos
-      )
-      if (visibleVariantWindow[0] !== null && firstIndex < visibleVariantWindow[0]) {
-        return firstIndex - 10
-      }
-
-      return firstIndex + 10
-    }
-    return 0
-  }
 
   useEffect(() => {
-    const searchIndex = getFirstIndexFromSearchText(filter, renderedVariants, renderedTableColumns)
+    const searchIndex = getFirstIndexFromSearchText(filter, renderedVariants, renderedTableColumns, visibleVariantWindow)
     setCurrentSearchIndex(searchIndex)
-    
+
     if (termLastSearched === null) {
       return
     }
     // @ts-expect-error TS(2531) FIXME: Object is possibly 'null'.
     table.current.scrollToDataRow(searchIndex)
-
   }, [termLastSearched]) // eslint-disable-line react-hooks/exhaustive-deps
 
   const datasetLabel = labelForDataset(datasetId)
--- a/browser/src/VariantList/Variants.tsx
+++ b/browser/src/VariantList/Variants.tsx
@@ -108,7 +108,9 @@ const Variants = ({
     includeIndels: true,
     includeExomes: true,
     includeGenomes: true,
+    includeContext: true,
     searchText: '',
+    
   })
 
   const [sortState, setSortState] = useState({
@@ -164,6 +166,8 @@ const Variants = ({
   )
 
   const [positionLastClicked, setPositionLastClicked] = useState(null)
+  const [termLastSearched, setTermLastSearched] = useState(null)
+
   // @ts-expect-error TS(7006) FIXME: Parameter 'position' implicitly has an 'any' type.
   const onNavigatorClick = useCallback((position) => {
     setSortState({
@@ -172,6 +176,18 @@ const Variants = ({
     })
     setPositionLastClicked(position)
   }, [])
+  
+    // @ts-expect-error TS(7006) FIXME: Parameter 'position' implicitly has an 'any' type.
+  const onSearchResult = useCallback((position) => {
+    console.log("Searched row:", position);
+  
+    setSortState({
+      sortKey: 'variant_id',
+      sortOrder: 'ascending',
+    });
+    setTermLastSearched(position);
+  }, []);
+
 
   useEffect(() => {
     if (positionLastClicked === null) {
--- a/browser/src/VariantList/Variants.tsx
+++ b/browser/src/VariantList/Variants.tsx
@@ -232,20 +232,27 @@ const Variants = ({
     if (index === -1) {
       index = renderedVariants.length - 1
     }
+
     // @ts-expect-error TS(2531) FIXME: Object is possibly 'null'.
     table.current.scrollToDataRow(index)
   }, [positionLastClicked]) // eslint-disable-line react-hooks/exhaustive-deps
 
 
   useEffect(() => {
-    const searchIndex = getFirstIndexFromSearchText(filter, renderedVariants, renderedTableColumns, visibleVariantWindow)
-    setCurrentSearchIndex(searchIndex)
-
     if (termLastSearched === null) {
+      setCurrentSearchIndex(-1)
       return
     }
+
+    const searchIndex = getFirstIndexFromSearchText(filter, renderedVariants, renderedTableColumns, visibleVariantWindow)
+
+    if (searchIndex !== -1) {
+      setCurrentSearchIndex(searchIndex)
+    }
+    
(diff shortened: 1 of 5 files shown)


browser/src/VariantList/Variants.tsx
```diff
@@ -164,15 +168,19 @@ const Variants = ({
...
 
   const [positionLastClicked, setPositionLastClicked] = useState(null)
-  // @ts-expect-error TS(7006) FIXME: Parameter 'position' implicitly has an 'any' type.
-  const onNavigatorClick = useCallback((position) => {
+  const [termLastSearched, setTermLastSearched] = useState(null)
+
+  const createCallback = useCallback((sortByKey: string, stateSetter: any) => (position: number) => {
     setSortState({
-      sortKey: 'variant_id',
+      sortKey: sortByKey,
```
rileyhgrant, line 175: It seems like this could be narrowed down by using the existing `Variant` type and getting all of its possible properties. I might ping Phil on this to see if he knows a way to do this, some brief searching didn't yield any good results.

Edit:
Then again, since users are sorting based on clicking the rendered column headers, they will by definition always be sorting by a valid `sortKey`, so maybe stronger typing here doesn't actually aid in safety of the code.

browser/src/VariantList/Variants.tsx
```diff
@@ -184,7 +192,7 @@ const Variants = ({
     }
 
     index = renderedVariants.findIndex(
-      (variant: any, i: any) =>
+      (variant: Variant, i: any) =>
```
rileyhgrant, line 195: Lots of type comments on this one, apologies. I think we can narrow down the type of `i` to `number`.

browser/src/VariantList/Variants.tsx
```diff
@@ -198,6 +206,38 @@ const Variants = ({
...
   }, [positionLastClicked]) // eslint-disable-line react-hooks/exhaustive-deps
 
+  function getFirstIndexFromSearchText(
+    searchFilter: VariantFilterState,
+    variantSearched: Variant[],
+    variantsTableColumns: any
+  ) {
+    const searchedVariants = getFilteredVariants(searchFilter, variantSearched, variantsTableColumns)
+  
+    if (searchedVariants.length > 0) {
```
rileyhgrant, line 216: I like this additional logic to get fix those edge cases, thanks for looking into that. When I locally try these changes, I'm now getting the search going far past the matched searched (I checked on Firefox and Chrome). Both if I'm before the search term, or after the search term, it scrolls far past the the highlighted variant (~4 full displayed sizes worth of variants). 

Any ideas as to what might be causing this? I haven't looked too far into it, and am happy to pair program and debug if that's helpful.

browser/src/VariantList/filterVariants.ts
```diff
@@ -1,6 +1,51 @@
...
+  includeContext: boolean,
+  searchText: string,
+}
+
+export function getFilteredVariants(
+  filter: VariantFilterState,
+  variants: Variant[],
+  variantTableColumns: any
+){
+  const searchColumns = variantTableColumns.filter((column: any) => !!column.getSearchTerms)
```
rileyhgrant, line 27: Another type comment, there are a couple `any`s here. If the `any` on line 27 gets narrowed down, the rest of those `any`s and variables in general in that chain should be able to be un-typed, since TypeScript will know what the type is at the beginning.

browser/src/VariantList/filterVariants.ts
```diff
@@ -12,57 +57,40 @@ const filterVariants = (variants: any, filter: any, selectedColumns: any) => {
   if (!isEveryConsequenceCategorySelected) {
     filteredVariants = variants.filter((variant: any) => {
       const category = getCategoryFromConsequence(variant.consequence) || 'other'
-      return filter.includeCategories[category]
+      return (filter.includeCategories as any)[category] // TODO: fix this
```
rileyhgrant, line 60: Was this TODO meant as a signal to come back to this as part of this PR?

rileyhgrant (changes requested):

rileyhgrant:
Oops, didn't include a comment with the request-changes button press.

Overall very nicely done Elissa! It may seem like there are a couple comments, but many of them are more minor comments about narrowing down Types, always good practice and something that we are actively focusing on at the moment (after migrating this codebase to TS, the tool used introduced a lot of `any`s throughout the codebase, so part of the working plan is to progressively take out these `any`s when any person make a PR that touches files that haven't already been cleaned up).

The major comment I have is in regards to the functionality, locally on my machine the search functionality seemed to always scroll further down the table than the matched variant. Feel free to message me if you want to do a code walk or demo, if either of those and/or pair programming would be helpful in debugging this. Thanks for taking the time to add more commits to work towards fixing the case you noticed of the unintended behavior if the variant is above the position the user is currently looking at.

Most of the other minor comments are around typing. Additionally, if you find you have a lack of things on your plate, you could go through these files and look at the lines of code you didn't touch to clean up any remaining `any` types that exist in these 3 files. This last point is not blocking, and you don't need to do that for approval of this PR though.

On the whole, nicely done on tackling this issue, it's shaping up to a nice quality of life improvement for users, that has been requested by some key stakeholders in the past!

browser/src/VariantList/Variants.tsx
```diff
@@ -193,11 +206,43 @@ const Variants = ({
...
+      const firstIndex = renderedVariants.findIndex((variant: Variant) => variant.pos === firstVariant.pos)
+      if (positionLastClicked !== null && firstVariant.pos < positionLastClicked) {
+        return firstIndex - 10
+      }
+      return firstIndex + 10
+    }
+    return 0
+  }
+
+  const searchIndex = getFirstIndexFromSearchText(filter, renderedVariants, renderedTableColumns)
```
rileyhgrant, line 235: Upon re-review, I noticed that due to logic that exists to re-render the variant bubble type track and the grid based on the state of the mouse in the other (e.g. highlight bubbles in the above track when a user hovers a variant in the table below), since this computation is happening outside of a scoped `useEffect` hook, this function is called anytime the mouse moves inside the variant table (potentially dozens of times per second). This was unexpected, as most components don't re-render based solely on mouse position.

While this isn't a particularly heavyweight function, I believe it is generally better practice to keep some piece of computation like this, that is only needed in a particular circumstance, happening close to where it is needed (in this case, in the `useEffect` hook below), particularly since this is computation happening in the users' browsers.

browser/src/VariantList/Variants.tsx
```diff
@@ -211,15 +215,19 @@ const Variants = ({
...
+      searchFilter,
+      variantSearched,
+      variantsTableColumns
+    )
+
     if (searchedVariants.length > 0) {
       const firstVariant = searchedVariants[0]
-      const firstIndex = variants.findIndex((variant: Variant) => variant.pos === firstVariant.pos)
+      const firstIndex = renderedVariants.findIndex((variant: Variant) => variant.pos === firstVariant.pos)
       if (positionLastClicked !== null && firstVariant.pos < positionLastClicked) {
```
rileyhgrant, line 227: I believe there still exists a case in which a user scrolls past the variant in the table below, without ever clicking the bubble track (I'm not entirely sure what to call that track). In that case, if the user searches for a variant that is above the currently displayed window in the table, this `if` statement will not be triggered, and the table will jump to `firstIndex + 10`, rather than `firstIndex - 10`.

I'd suggest comparing the value of `firstIndex` to the first value of `visibleVariantWindow[0]` (an array containing the first and last indices currently visible in the table and bubble track) piece of state that is defined on line `154`. Locally, when making this change, this resolved the inconsistent behavior I was observing.

browser/src/VariantList/variantTableColumns.tsx
```diff
@@ -45,8 +46,22 @@ const getConsequenceDescription = (contextType: any) => {
       return ' for consequence in this transcript'
   }
 }
-
-const variantTableColumns = [
+export type VariantTableColumn = {
```
rileyhgrant, line 49: Nice! I know some of these larger types can be a bit of a chore, particularly depending on your familiarity with the code base, but I do think it both helps to build understanding of the codebase, and gives all the juicy benefits of static typing :)

rileyhgrant (changes requested):
Nice job on sticking this through, and digging into defining some of those larger types. That's the kind of work that can be annoying in the short term, but proves useful and saves time in the long term.

I have 2 main comments this time around:
1. Due to the `Variants` component's rendering depending on mouse position, we should move the function call that checks the index of the first variant located when searching into the `useEffect` hook that handles the logic of moving the context to this variant. This will help keep the logic localized and in my humble opinion a bit easier to follow, but more importantly, avoids re-running this function on every frame that the user moves their mouse within the variant table
2. To handle the case where users scroll lower in the variant table without ever clicking on the bubble track, I think we should compare the visible variant window (which is already stored in a `useState` hook) to the indice of the first located variant, rather than using the last position clicked. Using the visible variant window should handle all the cases, and make the the component behave as intended (as far as I can tell).

browser/src/VariantList/Variants.tsx
elissa-alarmani: That's a great point and thanks for suggesting using that array of the visible variant window. Logically this feature is connected with the table view rather than the feature on the page about clicking the variants, so this makes more sense to use as a comparer for the edge cases.

rileyhgrant (changes requested):
Apologies for taking so long to submit a formal review this time around -- definitely my bad.

Unfortunately, I'm still encountering a bug where if I 
1. Search for a variant
2. Scroll very, very far down the table
3. Clear the search with the "x" button, 
4. Re-search with any term the table does not jump.

Once in this state, the table appears to not jump to any search term, however, it does highlight terms.

I'm not sure how often this bug would arise in production, but I believe it would be worth figuring out the root cause. I think the first step would be confirming that you can replicate this bug on your machine. Can you confirm that this happens for you too?

elissa-alarmani:
Thanks for the review Riley! I'm weirdly not able to recreate this bug, what are the specific variants you're searching so I could try to replicate it.

browser/src/VariantList/filterVariants.ts
```diff
@@ -12,57 +62,40 @@ const filterVariants = (variants: any, filter: any, selectedColumns: any) => {
...
```
rileyhgrant, line 65: Was this `// TODO: fix this` meant to be circled back to?

rileyhgrant (commented):
Pretty much there with this PR to my eyes, nicely done! I just have one remaining comment about a `TODO:` that was left in the code base.

elissa-alarmani:
My bad! I missed that TODO we left early on in this issue, but just fixed the type error

rileyhgrant (approved):
Awesome, nicely done on this one, Elissa.

PR #1162, opened 2023-08-17
Update AF display when AN is 0

### LINKED ISSUE #1161: Update how AF is displayed when there's no AN for a population
Per discussion in [this slack thread](https://atgu.slack.com/archives/CRA2TKTV0/p1692216171078009), the current way that the browser handles the display of Allele Frequency (AF) when Allele Count (AC) and Allele Number (AN) are both 0, is misleading. 

Currently, the browser handles this by calling the AF 0 for that row. This implies that there is information about the variant.

Instead, some non-number symbol should be used in this instance, 

> "it's really important that users understand that we have no information on the variant vs getting the impression that we're confident that there's no variant present at that sight"

rileyhgrant:
Resolves #1161 
Relevant [Slack thread](https://atgu.slack.com/archives/CRA2TKTV0/p1692216171078009)

Currently, when the Allele Number (AN) and Allele Count (AC) for a variant are 0, the browser displays the Allele Frequency (AF) in the PopulationTable on the Variant Page, for a given population or subpopulation, as 0. This is misleading, as it implies confidence that there is no variant there, rather than signaling that there is no information on the variant for that population.

This PR updates the `PopulationsTable` component to display a '`-`', as suggested in the Slack thread.

Links for convenience: 
- [Prod link](https://gnomad.broadinstitute.org/variant/1-55505681-G-T) to a Variant with AC and AN of 0
- [Localhost link](http://localhost:8008/variant/1-55505681-G-T) to the same variant (to see the change)

---

Screenshot 2023-08-17 at 16 34 09

### DIFF
--- a/browser/src/VariantPage/PopulationsTable.tsx
+++ b/browser/src/VariantPage/PopulationsTable.tsx
@@ -40,6 +40,23 @@ const SEX_IDENTIFIERS = ['XX', 'XY']
 const isSexSpecificPopulation = (pop: any) =>
   SEX_IDENTIFIERS.includes(pop.id) || SEX_IDENTIFIERS.some((id) => pop.id.endsWith(`_${id}`))
 
+// if the allele number (denominator) is 0, return a non-number
+//   to signal to users that there is no information to be displayed about
+//   the frequency, rather than artificially calling it 0
+const calculatePopAF = (ac: number, an: number) => {
+  if (an === 0) {
+    return '-'
+  }
+  return ac / an
+}
+
+const renderPopAF = (af: number | string) => {
+  if (typeof af === 'number') {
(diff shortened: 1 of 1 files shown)


mattsolo1 (approved):

MaryGoAround:
Sorry, what is the difference between population frequency and gnomAD frequency

I have this from UK biobank

image

rileyhgrant:
Hey @MaryGoAround 

Thanks for filing this issue! As this appears to be something relating more to the gnomAD dataset itself, and less the browser, is it possible to redirect your question to our forum? https://discuss.gnomad.broadinstitute.org

The forum includes members of the gnomAD methods team that are closer to the creation of the data, and are better equipped to answer your question. Thanks!

PR #1169, opened 2023-08-30
Docker browser build edits

ignatiusm:
I had difficulty building the browser docker image.

This PR contains some tiny edits to `deploy/dockerfiles/browser/browser.dockerfile`, `development/env.sh` and `CONTRIBUTING.md` files.

I'm not sure if you want to include my dockerfile changes, but the `development/env.sh` and `CONTRIBUTING.md` changes will be helpful for future collaborators.

Summary of changes:
- advise people to create a blank `browser/build.env` file before running `development/env.sh`
- fix the call to `docker-compose` (adds a hypen)
- change location of change of `user` in dockerfile so that `yarn install` can complete with correct permissions.

### DIFF
--- a/deploy/dockerfiles/browser/browser.dockerfile
+++ b/deploy/dockerfiles/browser/browser.dockerfile
@@ -3,8 +3,6 @@ FROM --platform=linux/amd64 node:14.17-alpine as build
 RUN mkdir -p /home/node/app && chown -R node:node /home/node/app
 WORKDIR /home/node/app
 
-USER node
-
 ENV NODE_ENV=production
 
 # Install dependencies
@@ -35,6 +33,8 @@ FROM --platform=linux/amd64 nginx:stable-alpine
 COPY --from=build /home/node/app/browser/dist/public /usr/share/nginx/html
 
 COPY deploy/dockerfiles/browser/browser.nginx.conf /etc/nginx/browser.nginx.conf.template
+
+USER node
 
 CMD REAL_IP_CONFIG=$([ -z "${PROXY_IPS:-}" ] || echo "$PROXY_IPS" | awk 'BEGIN { RS="," } { print "set_real_ip_from " $1 ";" }') \
   envsubst "\$API_URL \$REAL_IP_CONFIG" < /etc/nginx/browser.nginx.conf.template > /etc/nginx/conf.d/default.conf && \
--- a/deploy/dockerfiles/browser/browser.dockerfile
+++ b/deploy/dockerfiles/browser/browser.dockerfile
@@ -3,6 +3,8 @@ FROM --platform=linux/amd64 node:14.17-alpine as build
 RUN mkdir -p /home/node/app && chown -R node:node /home/node/app
 WORKDIR /home/node/app
 
+USER node
+
 ENV NODE_ENV=production
 
 # Install dependencies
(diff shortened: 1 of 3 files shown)


sjahl:
Hi @ignatiusm -- thanks for the contributions, much appreciated. The documentation fixes look fine to me. Could you describe the issues/errors you were running into that prompted moving the `USER` directive in the dockerfile? My recollection is that we're setting the user before the `yarn install` specifically to avoid running the install as a root user, so I'm curious to see what the issues were there.

Thanks again!

ignatiusm:
Hi @sjahl,

Here are the errors I get when running `./development/env.sh browser build`:

```
error Could not write file "/home/node/app/yarn-error.log": "EACCES: permission denied, open '/home/node/app/yarn-error.log'"
error An unexpected error occurred: "EACCES: permission denied, mkdir '/home/node/app/node_modules'".
```

I'm pretty puzzled, hopefully it isn't a noob error on my part. Here are a few of the things I have checked:
- [correct version of docker for using copy with chown](https://stackoverflow.com/questions/44766665/how-do-i-docker-copy-as-non-root?answertab=votes#tab-top)
- comment out the `volumes` block in `development/browser.docker-compose.yaml` in case they inherited permissions from my filesystem

deploy/dockerfiles/browser/browser.dockerfile
```diff
@@ -36,6 +34,8 @@ COPY --from=build /home/node/app/browser/dist/public /usr/share/nginx/html
 
 COPY deploy/dockerfiles/browser/browser.nginx.conf /etc/nginx/browser.nginx.conf.template
 
+USER node
```
sjahl, line 37: This should be reverted -- since it comes after the `FROM nginx:stable-alpine` line, it'll result in running processes as a user that doesn't exist in the `nginx` image, rather than what the image intends, which runs things as the `nginx` user.

sjahl (changes requested):
@ignatiusm thanks, interesting!

I can't reproduce that failure mode on the `HEAD` of the main branch. I tested with both the latest release of Docker for Mac, and on a fresh ubuntu 22.04LTS VM with docker v24.0.5 to rule out any magic that I've got in my local build cache.

In any case, I'm happy accept the documentation and env.sh fix. I've left a code review comment inline requesting that moving the `USER` line be reverted. It's hard to tell why you might be hitting the build issues you are, but we won't actually want that `USER` change in the `FROM nginx` part of the dockerfile, since that is a nonexistent user in the nginx image.

ignatiusm:
@sjahl - thank you for taking the time to try to reproduce the error on a few different systems. I was troubled that you couldn't recreate it, so have done some digging.

It turns out it was due to me having an outdated version of docker compose (v1.29.2!) :facepalm: . I tried with a current version of docker compose plugin, and the build ran smoothly.

I'll edit the PR, which will result in only a tiny docs change (adding the creation of the `build.env` file).

ignatiusm:
Ready for your review @sjahl . I reverted the unnecessary changes (see comment above). It ended up being a _very_ small PR in the end :smile:

sjahl (approved):
@ignatiusm great! thanks for following up, it's helpful to know about cases like this, just to be aware that docker/compose versions are things that we should have people check when they run into issues.

PR #1174, opened 2023-09-06
Add Filtering by Number of Stars for Clinvar Variants

### LINKED ISSUE #734: Allow filtering ClinVar variants by number of stars

elissa-alarmani:
Currently, there is no option to filter the Clinvar Variant Track on each Gene page by the number of Clinvar stars a variant has.

This PR creates a dropdown to allow the user to filter variants by their Clinvar star value based on the number of stars. This filter is implemented as a "n+ stars" as mentioned in the original issue, "there is likely not a use case for a user wanting to filter to only 4 star variants and 1 star variants".

Any UI suggestions would be appreciated!
ClinvarStarFilter1
ClinvarStarFilter2

Fixes #734


browser/src/ClinvarVariantsTrack/ClinvarVariantTrack.tsx
```diff
@@ -211,14 +220,28 @@ const ClinvarVariantTrack = ({ referenceGenome, transcripts, variants }: Props)
...
+          <FilterRow>
             <Checkbox
               id="clinvar-track-in-gnomad"
               label="Only show ClinVar variants that are in gnomAD"
               checked={showOnlyGnomad}
               onChange={setShowOnlyGnomad}
             />
-          </ControlRow>
+            <label htmlFor="star-filtering">
+              Filter by Stars: &nbsp;
```
mattsolo1, line 231: Maybe this should be "Filter by review status"?

See: https://www.ncbi.nlm.nih.gov/clinvar/docs/review_status/

image

Should we also link to this page in the label so that users can understand what review status means? I.e. the label becomes "Filter by <a href="https://www.ncbi.nlm.nih.gov/clinvar/docs/review_status/">review status</a>" or whatever the appropriate link component is.

mattsolo1 (commented):
The alignment is a bit off in my browser
image

We want the label text perfectly aligned with "Only show ClinVar..."

browser/src/ClinvarVariantsTrack/ClinvarVariantTrack.tsx
```diff
@@ -211,14 +229,32 @@ const ClinvarVariantTrack = ({ referenceGenome, transcripts, variants }: Props)
...
             <Checkbox
               id="clinvar-track-in-gnomad"
               label="Only show ClinVar variants that are in gnomAD"
               checked={showOnlyGnomad}
               onChange={setShowOnlyGnomad}
             />
-          </ControlRow>
+            <label htmlFor="star-filtering">
+              Filter by{' '}
+              <Link to="https://www.ncbi.nlm.nih.gov/clinvar/docs/review_status/">
```
mattsolo1, line 241: When I click on this it takes me to http://localhost:8008/gene/https://www.ncbi.nlm.nih.gov/clinvar/docs/review_status/?dataset=gnomad_r2_1

There is a special link component for external links, for example: https://github.com/broadinstitute/gnomad-browser/blob/70d1916bb76998c9d815c33547d81a8579891c68/browser/src/GenePage/GeneReferences.tsx#L73

mattsolo1 (changes requested):
In Safari this looks properly aligned

image

In Chrome it looks properly aligned

image

In Firefox it still looks a bit off

image

browser/src/ClinvarVariantsTrack/ClinvarVariantTrack.tsx
```diff
@@ -87,6 +87,22 @@ const SelectCategoryButton = styled(Button)`
   line-height: 18px;
 `
 
+const FilterRow = styled.div`
+  display: flex;
+  align-items: center;
```
mattsolo1, line 92: Does it look a bit cleaner to use `justify-content: space-between;` here? This puts the review status filter justified to the right. Otherwise I feel like there's not enough space between this and "Only show Clinvar..."

Thoughts @ch-kr @rileyhgrant @phildarnowsky?

image

elissa-alarmani: I actually like this layout better I feel like it similar to the Coverage track with the right side dropdown

ch-kr: I prefer this as well. also, it'll be cool to have this feature on the browser!

browser/src/ClinvarVariantsTrack/ClinvarVariantTrack.tsx
```diff
@@ -119,14 +135,16 @@ const ClinvarVariantTrack = ({ referenceGenome, transcripts, variants }: Props)
...
 
   const filteredVariants = variants.filter(
     (v) =>
       // @ts-expect-error TS(7053) FIXME: Element implicitly has an 'any' type because expre... Remove this comment to see the full error message
       includedClinicalSignificanceCategories[clinvarVariantClinicalSignificanceCategory(v)] &&
       // @ts-expect-error TS(7053) FIXME: Element implicitly has an 'any' type because expre... Remove this comment to see the full error message
       includedConsequenceCategories[getCategoryFromConsequence(v.major_consequence)] &&
-      (!showOnlyGnomad || v.in_gnomad)
+      (!showOnlyGnomad || v.in_gnomad) &&
+      (starFilter === 0 || v.gold_stars >= starFilter)
```
phildarnowsky-broad, line 147: The value of `v.gold_stars` should always be non-negative, and hence if `starFilter` is 0, `v.gold_stars >= starFilter` will always be true. Hence you can get rid of the explicit `starFilter === 0` check here.

phildarnowsky-broad (changes requested):
I'd like to see some tests added, specifically these two:

1. A snapshot test, to help guard against unexpected changes in the rendered output
2. A test that changing the selected value changes the content appropriately

One general long-term goal we have here is to build up the test suite for the browser, to make it more useful for catching bugs before they even get merged to `main`. Snapshot tests are limited in the kinds of bugs they can pick up, but are generally simple and easy to do, and give you broad test coverage, so they're a good starting point. Since we're adding new interactive functionality here, we also want to test that, but there's no need at this point to necessarily go back and add tests for the other filters--a good rule of thumb is "test it when you touch it".

The test for this module should go in `browser/src/ClinvarVariantsTrack/ClinvarVariantsTrack.spec.tsx`.

For good examples of typical snapshot test, see https://github.com/broadinstitute/gnomad-browser/blob/main/browser/src/VariantList/VariantFlag.spec.tsx.

This is a typical interactive test: https://github.com/broadinstitute/gnomad-browser/blob/7983136ec596ce120f9c4be92f1b2cacf26b5389/browser/src/GenePage/GenePage.spec.tsx#L83 

I'm not clear on what your experience with testing is like so I'm not sure how much detail to go into here, if you have questions or would like me to explain more explicitly what I'm suggesting please feel free to ask--Riley can also probably clarify a lot as well.

mattsolo1 (approved):
Hurray for tests! I'm sure the Clinvar users will be excited about this feature. Also remember to update the changelog when going live.

phildarnowsky-broad (changes requested):
I flagged one small change to the interaction test.

Beyond that, I'd suggest that, once everything in this branch is working and approved, but before you merge to `main`, that you `git rebase` to squish the whole string of commits in this PR down to one single commit. The ideal for commits in a PR merged to `main` is that:

- Each commit merged to `main` corresponds to a single coherent task, analogous to a  module in code or a paragraph in an essay
- Each commit merged should look like you got it right the first time--in other words, if you edit some code in file A within commit ABC, and then go back and fix the formatting in file A in commit QRS, you should `rebase` so that ABC and QRS are combined into one commit
- If you check out any arbitrary commit and run the test suite, the suite should pass

Those first two points make following the logic of your PR much easier for someone who's coming along later and trying to understand what you did. The third point makes `git bisect`, which is a very useful tool, work much better.

Rebasing is a thing that people's familiarity with varies widely, please let me know if you'd like me to explain in more detail and/or help walk you through the process.

phildarnowsky-broad (approved):
Looking good, please remember to use `git rebase` to combine these all into a single commit before merging to `main` as discussed.

PR #1193, opened 2023-09-26
Fix gene page crashes when filtering variants

### LINKED ISSUE #1191: issue with "display nearby variants" search
### What you did: I tried searching for a variant, using the p. nomenclature (e.g. 
p.Arg8Leu in DNAH11), in the variant table on the gene page and deselected the "display nearby variants" box. I was looking for a variant that was at the same position but has a different amino acid change (e.g. p.Arg8Tyr). I just deleted some of the original search criteria in the search bar and the browser changed to display the "something went wrong" page. The same thing happened when i redid the first step and just started typing something different into the search bar that didn't match anything that existed in the table. This could just be when I typed an error into the search bar as well.

 
Trying to find a variant that did not exist using the gnomAD gene page search in the variant table.
https://gnomad.broadinstitute.org/gene/ENSG00000105877?dataset=gnomad_r2_1

I expected to see no matching variant in the variant table but not the error page.

The gene page would be there with a note of no matching variants in the variant table.

I saw the something went wrong page (screenshot attached). 
image

rileyhgrant:
Resolves #1191 

Fixes two errors that cause crashes that are currently present in the Variant tracks.

1. #1191, when keep search context is toggled off, an unmatched search term no longer tries to scroll the table, which prevents the crash observed in the issue
2. When filtering the variants to only those that match the search term, if a user clicks on the bubble type track above when the table is filtered to no results, the page currently crashes as it tries to scroll the null table. A null check in this PR prevents that crash

### DIFF
--- a/browser/src/VariantList/Variants.tsx
+++ b/browser/src/VariantList/Variants.tsx
@@ -208,7 +208,7 @@ const Variants = ({
 
   // When a user clicks on the bubble track, update the position in the variant table
   useEffect(() => {
-    if (positionLastClicked === null) {
+    if (positionLastClicked === null || table.current === null) {
       return
     }
 
@@ -228,7 +228,7 @@ const Variants = ({
       index = renderedVariants.length - 1
     }
 
-    // @ts-expect-error TS(2531) FIXME: Object is possibly 'null'.
+    // @ts-expect-error TS(2339) FIXME: 'scrollToDataRow' does not exist on type 'never'.
     table.current.scrollToDataRow(index)
   }, [positionLastClicked]) // eslint-disable-line react-hooks/exhaustive-deps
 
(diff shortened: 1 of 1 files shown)


rileyhgrant:
Hey @mattsolo1,

This is a very small PR, but it fixes 2 bugs that are currently in production and that cause crashes on the Gene Page (one that is new, one that appears to be quite old that was happened upon when fixing the first one). So if possible, review on this sooner rather than later would be appreciated!

mattsolo1 (approved):

rileyhgrant:
> I think this is still not working. We just tried it again and got the same error.

@msingerb could you try again now? Based on the timestamp of your comment, when you checked PR had been merged into main, but not yet deployed to production.

PR #1198, opened 2023-10-02
Modernize toolchain

phildarnowsky-broad:


reads/start.sh
```diff
@@ -7,4 +7,4 @@ export NODE_ENV="development"
 DEFAULT_PORT="8000"
 export PORT=${PORT:-$DEFAULT_PORT}
 
-yarn run nodemon ./src/server.js
+pnpm nodemon ./src/server.js
```
sjahl, line 10: The reads api calls `node src/server.js` (as well as `yarn install`) in deploy/dockerfiles/reads/reads-api.dockerfile to startup. Do we need to update the build in that dockerfiles/reads service as well? or will updating the reads service be a subsequent PR?

sjahl (commented):
One build question inline regarding the reads API -- just wanted to make sure we weren't overlooking that particular component.

The current set of dockerfile and Actions build step changes all look good to me. As far as the code goes, I tried to focus more on the changes that didn't seem strictly formatting related, and not seeing anything unexpected there. This all looks good to me, pending that question about the reads API build.

.tool-versions
```diff
@@ -1,2 +1,2 @@
-nodejs 14.19.3
-yarn 1.22.19
+nodejs 18.17.1
+pnpm 8.8.0
```
mattsolo1, line 2: Does it make sense to use latest pnpm version (` 8.14.3`)?

reads/start.sh
mattsolo1: Noting that the variant page crashes on the current demo.  Could be related to this?

stub_declarations.d.ts
```diff
@@ -0,0 +1,4 @@
+declare module 'query-string'
```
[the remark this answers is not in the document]
phildarnowsky-broad: Yes, tl;dr they weren't working but I see what the problem was now, updated.

browser/package.json
```diff
@@ -37,28 +36,52 @@
     "memoize-one": "^5.0.2",
     "polished": "^3.2.0",
     "prop-types": "^15.7.2",
-    "query-string": "5",
-    "react": "^16.13.1",
+    "query-string": "~5",
+    "react": "^17.0.0",
```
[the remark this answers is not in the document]
phildarnowsky-broad: We have a lot of dependencies that aren't yet compatible with it, uptake of React 18 seems to have been very slow generally.

.tool-versions
phildarnowsky-broad: Probably, I'll make sure it works OK.

package.json
```diff
@@ -34,65 +30,45 @@
...
     "eslint-plugin-react": "^7.20.3",
     "eslint-plugin-react-hooks": "^4.0.5",
-    "file-loader": "^6.2.0",
-    "html-webpack-plugin": "^5.2.0",
     "jest": "^29.0.0",
     "jest-environment-jsdom": "^29.0.1",
+    "jest-mock": "^29.7.0",
     "lodash-es": "^4.17.21",
     "nodemon": "^2.0.4",
+    "pnpm": "^8.8.0",
```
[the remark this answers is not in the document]
phildarnowsky-broad: I don't remember, will see if we can do without it.

phildarnowsky-broad:
@sjahl @mattsolo1 found the problem, the whole question of the reads deployment turned out to be a red herring. ready for re-review.

mattsolo1:
I clicked through the demo a bit more and couldn't find any issues, so I think this is good to go!

mattsolo1 (approved):

PR #1204, opened 2023-10-10
Reject genes on Y chromosome pseudoautosomal region

phildarnowsky-broad:


mattsolo1 (commented):
Just curious why this needs a distinct new step for itself? For simplicity would it fit into a prior step?

phildarnowsky-broad:
@mattsolo1 we definitely could put it in an existing step. I seperated it out here to save us some time when we use this shortly to rebuild the data: this way, we can use the old overall output as the input to this new step, rather than back up one or more steps and rebuild from that point on.

However I don't feel very strongly about this point so if you would rather integrate this into an existing step I'm fine with that too.

mattsolo1 (approved):

PR #1210, opened 2023-10-16
Update input schema for new v4 mock variant dataset

mattsolo1:


data-pipeline/src/data_pipeline/datasets/gnomad_v4/types/initial_variant.py
```diff
@@ -162,7 +188,7 @@ class Rf:
 
 
 @attr.define
-class RegionFlag:
+class RegionFlags:
```
mike-w-wilson, line 191: There are new flags:
```
fail_interval_qc: bool
outside_ukb_capture_region: bool
outside_broad_capture_region: bool
```
Sorry if its already been discussed and they are being omitted but just wondering if adding them to the browser should be brought up at tomorrow's meeting,  @ch-kr ?

ch-kr: sounds good -- I'll let Sam know that we have a question for her `Hold for any final help/FAQ or Stats page questions` bullet

Issue #1183, opened 2023-09-15
Create Redis cluster ahead of ASHG


rileyhgrant:
Or otherwise find a way to tolerate the anticipated high load due to ASHG

Issue #1208, opened 2023-10-13
Certain pages fail when redis is configured but unavailable


sjahl:
When standing up a new demo that lacked redis, but had the redis URL defined in its pod environment variables, some pages seemed fine, but we found that other pages failed to load, and logged timeouts related to the redis cache.

Initial diagnosis here is that we have something (i.e. the GeneCoverage query, when loading the gene page) that is either:

- not failing gracefully when it has a timeout connecting to redis
- or, is waiting too long to try and connect to a dead/nonexistent redis and then timing out before it can retrieve the data from elasticsearch.

This would be a good case to try and resolve, so that the app is at least somewhat resilient to failures of the redis cache, and has the ability to run without it. Or, we could fail to start up if a cache is configured but cannot be used.

rileyhgrant:
From browser meeting:
- We should fail more gracefully when the cache is unavailable
- There's a real prod use case of when redis is down, we'd like to be able to return things
- Any query that uses `withCache()` will hang for 30 seconds then almost certainly fail if there's no redis

PR #1215, opened 2023-10-27
Support a clustered redis

sjahl:
works on #1183 -- probably need a developer to vet this a bit...


sjahl:
OK, testing this in a dev environment, and things seem to be working mostly as expected.

Given the time crunch, I don't know if this is a now thing, or a later thing... but right now this config ONLY supports a clustered redis with a failover sentinel. It would be nice for us to maybe handle the use case where someone can easily drop in a single redis node instead. Could also lob that into the project to fix how we handle not having a redis in #1208

deploy/manifests/browser/base/api.deployment.yaml
```diff
@@ -43,16 +43,16 @@ spec:
                   key: ips
             # Global rate limiting
             - name: MAX_CONCURRENT_ES_REQUESTS
-              value: '100'
+              value: '10'
```
phildarnowsky-broad: Since it looks like this commit just undoes changes made in the immediately previous commit, let's merge this one into the previous one. I made Elissa do it, fair's fair. 😅

deploy/manifests/browser/base/api.deployment.yaml
```diff
@@ -35,7 +35,7 @@ spec:
             - name: WRITE_CACHE_REDIS_URL
               value: redis
             - name: RATE_LIMITER_REDIS_URL
-              value: redis://redis:26379/2
+              value: redis
```
phildarnowsky-broad, line 38: Looks like you could merge this commit and the previous two all together.

phildarnowsky-broad (approved):
I pretty much have to take the k8s changes on faith but everything you did in the frontend code looks legit. I just have some small suggestions regarding squashing this down to a smaller number of commits, no notes otherwise.

phildarnowsky-broad (approved):

PR #1216, opened 2023-10-27
Add v2 RMC dataset

rileyhgrant:
Adds the data pipeline, graphql api, and frontend pieces required to load, serve, and render the v2 regional missense constraint (RMC) dataset.

### DIFF
--- a/browser/src/RegionalConstraintTrack.tsx
+++ b/browser/src/RegionalConstraintTrack.tsx
@@ -1,5 +1,4 @@
 import { transparentize } from 'polished'
-import PropTypes from 'prop-types'
 import React from 'react'
 import styled from 'styled-components'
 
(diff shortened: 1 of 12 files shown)


rileyhgrant:
Hiya again @phildarnowsky-broad.

Also requesting review on this one, with the hopes of getting it merged, with a submitted issue if there are design considerations that are not breaking, per standup this morning.

There's at least one `any`, and one `ts-ignore` comment remaining, these arise from a helper function that was created before my time (despite it being new to `main` here, it had lived in another branch before), I realize these are not ideal, I was having a bit of a time trying to figure out how to resolve these two remaining type errors, and am going to move onto other work for the time being.

Thanks as always for reviewing my PRs :)

data-pipeline/src/data_pipeline/pipelines/genes.py
```diff
@@ -199,6 +202,14 @@
     "/genes/homozygous_variant_cooccurrence_counts.ht",
 )
 
+pipeline.add_task(
+    "prepare_gnomad_v2_regional_missense_constraint",
+    prepare_gnomad_v2_regional_missense_constraint,
+    "/constraint/gnomad_v2_regional_missense_constraint.ht",
+    # TODO: before merging - update to a more permanent location for this data
```
phildarnowsky-broad, line 209: @rileyhgrant that reminds me, I actually did not do this for the SVs, I think if you have a current output Hail table in a dev environment, it'd be OK to take that and export it to the new ES rather than necessarily rerun the whole pipeline from scratch. So probably no need to worry about this now, we can move stuff around and fill in gaps in post.

browser/src/RegionalConstraintTrack.tsx
```diff
@@ -137,60 +147,70 @@ type OwnRegionalConstraintTrackProps = {
...
-                </TooltipAnchor>
-              )
-            })}
-          </svg>
-        </PlotWrapper>
-      )}
-    </Track>
-  </Wrapper>
-)
+// type RegionalConstraintTrackProps = OwnRegionalConstraintTrackProps &
```
phildarnowsky-broad: commented lines here can be deleted

phildarnowsky-broad (approved):
Flagged one extremely minor change, you are good to merge soon as that's done. No need to seek re-review.

Issue #858, opened 2022-01-10
Update version of GENCODE used in genes data pipeline


nawatts:
Update the version of GENCODE used in the genes data pipeline to match the version used in the version of VEP used to annotate v4.

https://github.com/broadinstitute/gnomad-browser/blob/b4e38686e23fe4a31ff7c9541fe78b82b4df286b/data-pipeline/src/data_pipeline/pipelines/genes.py#L36-L38

rileyhgrant:
Per [this slack thread](https://the-tgg.slack.com/archives/C018PB68VU5/p1677689628398959?thread_ts=1676408384.152109&cid=C018PB68VU5), the next GTEx release is going to be on Gencode 39.

rileyhgrant:
Closed by #1222

PR #1222, opened 2023-10-28
Update to GENCODE v39 for GRCh38 genes

phildarnowsky-broad:
@mattsolo1 This downloads v39 of the GENCODE GTF and uses that as the basis of the GRCh38 genes pipeline, in place of GENCODE v35. Please let me know if I've missed something as the ticket and related thread were a little terse.

I'm in the process at this moment of firing up a dataproc to try the pipeline, just to make sure there are no surprises.


phildarnowsky-broad:
closes #858

mattsolo1 (approved):

PR #1227, opened 2023-10-30
Update the term "Population" to "Genetic Ancestry Group", and "popmax" to "grpmax" on user facing portions of the browser

### LINKED ISSUE #1051: Update population/ancestry terms
Related to https://github.com/broadinstitute/gnomad_production/issues/707: we received feedback from external users that they prefer to use the term "genetic ancestry" rather than "population" or "ancestry" for clarity

rileyhgrant:
Resolves #1051 

Updates 

- the term "Population" to "Genetic ancestry group", 
- the term "popmax" to "grpmax" 

throughout the user facing text of browser.

---

Changing the term usage in the API, and possibly local variable names if we want, is a task for later.

### DIFF
--- a/browser/src/CopyNumberVariantPage/__snapshots__/CopyNumberVariantPage.spec.tsx.snap
+++ b/browser/src/CopyNumberVariantPage/__snapshots__/CopyNumberVariantPage.spec.tsx.snap
@@ -327,7 +327,7 @@ exports[`CopyNumberVariantPage with dataset gnomad_cnv_r4 with variant of type D
   </div>
   <section>
     <h2>
-      Genetic ancestry group Frequencies
+      Genetic Ancestry Group Frequencies
     </h2>
     <table
       className="Table__BaseTable-sc-7fgtt2-0 CNVPopulationsTable__Table-sc-1dnyyn-0 iypcwR"
@@ -779,7 +779,7 @@ exports[`CopyNumberVariantPage with dataset gnomad_cnv_r4 with variant of type D
   </div>
   <section>
     <h2>
-      Genetic ancestry group Frequencies
+      Genetic Ancestry Group Frequencies
     </h2>
     <table
       className="Table__BaseTable-sc-7fgtt2-0 CNVPopulationsTable__Table-sc-1dnyyn-0 iypcwR"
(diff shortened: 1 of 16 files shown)


browser/src/CopyNumberVariantPage/__snapshots__/CopyNumberVariantPage.spec.tsx.snap
```diff
@@ -779,7 +779,7 @@ exports[`CopyNumberVariantPage with dataset gnomad_cnv_r4 with variant of type D
   </div>
   <section>
     <h2>
-      Population Frequencies
+      Genetic ancestry group Frequencies
```
mattsolo1, line 782: All section titles (like on variant pages) should be in title case (Genetic Ancestry Group Frequencies)

browser/src/CopyNumberVariantPage/__snapshots__/CopyNumberVariantPage.spec.tsx.snap
```diff
@@ -795,7 +795,7 @@ exports[`CopyNumberVariantPage with dataset gnomad_cnv_r4 with variant of type D
               onClick={[Function]}
               type="button"
             >
-              Population
+              Genetic ancestry group
```
mattsolo1, line 798: Can probably keep lower case

mattsolo1 (changes requested):
You could also change all user-facing text (not the API query/code yet) of Popmax -> Grpmax as part of this PR.

rileyhgrant:
> You could also change all user-facing text (not the API query/code yet) of Popmax -> Grpmax as part of this PR.

Oop yeah, missed that, good call will do.

browser/src/CopyNumberVariantPage/__snapshots__/CopyNumberVariantPage.spec.tsx.snap
rileyhgrant: Seems like the other column headers are title case so going with that for now, happy to do sentence case instead if you prefer.

Thanks for catching the case consistency. 🙏

mattsolo1:
Actually I wonder if it should be Groupmax fully spelled out? @ch-kr?

mattsolo1 (approved):

PR #1228, opened 2023-10-30
Update Downloads page for v4

rileyhgrant:
Updates the downloads page to include v4 links

For launch, only GCP is guaranteed to have synced. AWS and Azure will be added to the downloads links after they have synced.


rileyhgrant:
Closing because I'm silly and bungled the naming on my branches, and I'd rather just open another PR than figure out how to shuffle commits around.

PR #1230, opened 2023-10-30
Update Downloads page for v4

rileyhgrant:
Updates the downloads page to include v4 links

For launch, only GCP is guaranteed to have synced. AWS and Azure will be added to the downloads links after they have synced.

Mostly waiting on this for the v4 release files to start syncing to GCP. Until that happens, there's not much of a point to updating the Downloads page

### DIFF
--- a/browser/src/DownloadsPage/DownloadsPage.tsx
+++ b/browser/src/DownloadsPage/DownloadsPage.tsx
@@ -1,7 +1,9 @@
 import React from 'react'
 import styled from 'styled-components'
 
-import { ExternalLink, PageHeading, Link } from '@gnomad/ui'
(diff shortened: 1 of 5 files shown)


mattsolo1:
Awesome, thank you Riley. Can you add a section for the v2 RMC downloads?

Also I think feel free to make this a real PR I can approve so we get into to main and part of the rolling demos.

mattsolo1 (approved):

rileyhgrant:
Going to clean this up (remove some todos) and add paths that have currently synced so this can get merged. 

This won't have all the proper paths for the moment, so we may get some comments about that on the internal demo in case some very curious party tries to download everything.

PR #1233, opened 2023-10-31
Update constraint tooltips for v4

rileyhgrant:
Updates tooltips for constraint on the gene page per a comment on the internal demo.


rileyhgrant:
This change is no longer relevant now that constraint has been removed from the MVP release. As such I'd guess that this PR should probably be closed instead of merged.

mattsolo1:
OK... this will still be useful to refer to in the future.

PR #1236, opened 2023-11-01
Remove v4 constraint for MVP

rileyhgrant:
Removes v4 constraint for the MVP launch, per a decision late on Tuesday (Oct 31, 2023).

- Removes v4 datasets from `hasConstraint` to have the frontend not render the table constraint table for v4
- Removes references to v4 constraint in help/faq text
- Removes v4 constraint section from downloads page

In addition, the genes pipeline has been re-run, and outputs were written to a new bucket (`genes_no_constraint_20231031T224855`, I tried to followed the convention as best I could figure). The indices have been loaded into the v4 elasticsearch, all that is left to do is change the aliases to point to the new indices after this PR is merged.

The new indices are (these are on the v4 elasticsearch cluster)
- `genes_grch38-2023-11-01--03-17`
- `transcripts_grch38-2023-11-01--03-19`

The old indices were (these still exist on the v4 elasticsearch cluster)
- `genes_grch38-2023-10-29--23-36`
- `transcripts_grch38-2023-10-29--23-38`

I'm happy to update these indices after these get merged. The only detrimental thing that will happen with the currently deployed demo is that (once the nginx cache has been cleared) the v4 gene and constraint pages will render the warning that reads something like "constraint not available for this gene" rather than "constraint not available for v4.0.0")


mattsolo1:
Sorry Riley I meant to open a new PR with my change but accidentally pushed the commit to your branch.

The snapshot tests are failing with even when I run `yarn jest -u`.

Message: `Invariant failed: You should not use <withRouter() /> outside a <Router>`

mattsolo1:
Changing the `<Link/>` to an `<a/>` enables tests to pass, so it's something to do with `<Link/>` being added to this component. Unfortunately then we lose link styles. There's something about the test config I'm not understanding.

rileyhgrant:
> Changing the `<Link/>` to an `<a/>` enables tests to pass, so it's something to do with `<Link/>` being added to this component. Unfortunately then we lose link styles. There's something about the test config I'm not understanding.

Interesting, yeah I'm not seeing where the router is being used in a router either, will poke around for a tiny bit and if it doesn't go just change it to an <a> tag and re-create the styles like you said?

mattsolo1 (approved):

PR #1237, opened 2023-11-01
Update v4 downloads again

rileyhgrant:
Update v4 downloads with checksums, sizes, and final synced paths.


mattsolo1 (changes requested):
Fantastic. The only change is constraint section needs removing.

rileyhgrant:
> Fantastic. The only change is constraint section needs removing.

Oop yeah I threw that commit in a diff PR, but it can be in here easily enough.

mattsolo1 (approved):
I manually checked all of the links... looks good!

PR #1238, opened 2023-11-01
use Persistent http cache

sjahl:
As an insurance policy to let us restart the browser during the launch without losing the cache


rileyhgrant (approved):
LGTM!

PR #1242, opened 2023-11-01
Don't allow TTN API lookups by gene or transcript

mattsolo1:


sjahl (approved):
This looks OK to me 👍

PR #1244, opened 2023-11-01
Update Total Values Genetic Ancestry Group Table (CNVs)

elissa-alarmani:
Currently, the total `Site Count` `Site Number`, and `Site Frequency` values in the Genetic Ancestry Group Frequency table do not match the totals provided. This PR updates the `total` row in the table with the `Site Count` and `Site Number` provided to derive the correct `Site Frequency` value.

### DIFF
--- a/browser/src/CopyNumberVariantPage/CNVPopulationsTable.tsx
+++ b/browser/src/CopyNumberVariantPage/CNVPopulationsTable.tsx
@@ -2,6 +2,7 @@ import React, { Component } from 'react'
 import styled from 'styled-components'
 
 import { BaseTable, TextButton, TooltipAnchor, TooltipHint } from '@gnomad/ui'
+import { CopyNumberVariant } from './CopyNumberVariantPage'
 
 const Table = styled(BaseTable)`
   min-width: 100%;
@@ -80,6 +81,7 @@ type OwnPopulationsTableProps = {
     }[]
   }[]
   initiallyExpandRows?: boolean
+  variant: CopyNumberVariant
 }
 
 type CNVPopulationsTableState = any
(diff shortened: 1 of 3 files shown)


browser/src/CopyNumberVariantPage/CNVPopulationsTable.tsx
```diff
@@ -80,22 +81,24 @@ type OwnPopulationsTableProps = {
...
+  variant: CopyNumberVariant
 }
 
 type CNVPopulationsTableState = any
 
 type CNVPopulationsTableProps = OwnPopulationsTableProps & typeof CNVPopulationsTable.defaultProps
 
 export class CNVPopulationsTable extends Component<
-  CNVPopulationsTableProps,
+  CNVPopulationsTableProps & { variant: CopyNumberVariant},
```
phildarnowsky-broad, line 92: It should be simple enough to fold all these various `WhateverProps` together to make things a little clearer, but let's save that for later. This whole business with `OwnWhateverProps & typeof Whatever.defaultProps` that you see here and there in the code is just an artifact of the automatic conversion of the codebase from JS to TS a long time ago, no need to look for any deep meaning in it. I clean these up when I am working near one but it's not a priority.

phildarnowsky-broad (approved):
Good to go

browser/src/CopyNumberVariantPage/CNVPopulationsTable.tsx
elissa-alarmani: Thanks for pointing this out, when I go back and go through the actual code review of the CNVs I'll make sure to come back to this.

PR #1245, opened 2023-11-01
Make better TTN message

mattsolo1:


elissa-alarmani (approved):
The new message looks good!

PR #1260, opened 2023-11-03
Update default structural dataset to SVs v4

### LINKED ISSUE #1252: Update default structural dataset for v4
The current default structural dataset in the `DatasetSelector` is the v4 CNVs. This should be changed to the v4 SVs.

Screenshot 2023-11-01 at 16 26 23

elissa-alarmani:
Currently, the default structural dataset in the `DatasetSelector` is the v4 CNVs. This PR changes that to the v4 SVs. 

Fixes #1252

### DIFF
--- a/browser/src/DatasetSelector.tsx
+++ b/browser/src/DatasetSelector.tsx
@@ -591,15 +591,15 @@ const DatasetSelector = withRouter(({ datasetOptions, history, selectedDataset }
   if (includeStructuralVariants || includeCopyNumberVariants) {
     const topLevelStructuralVariantDataset = hasStructuralVariants(selectedDataset)
       ? selectedDataset
-      : 'gnomad_sv_r2_1'
+      : 'gnomad_sv_r4'
 
     const topLevelCopyNumberVariantDataset = hasCopyNumberVariants(selectedDataset)
       ? selectedDataset
       : 'gnomad_cnv_r4'
 
-    const currentDataset = hasStructuralVariants(selectedDataset)
-      ? topLevelStructuralVariantDataset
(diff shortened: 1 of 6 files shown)


browser/src/DatasetSelector.tsx
```diff
@@ -591,15 +591,15 @@ const DatasetSelector = withRouter(({ datasetOptions, history, selectedDataset }
   if (includeStructuralVariants || includeCopyNumberVariants) {
     const topLevelStructuralVariantDataset = hasStructuralVariants(selectedDataset)
       ? selectedDataset
-      : 'gnomad_sv_r2_1'
+      : 'gnomad_sv_r4'
```
rileyhgrant, line 594: Now that there are two different datasets that have structural datasets, we'll need additional logic to determine the topLevelStructuralVariantDataset.

As it stands at the moment, this correctly updates the default SV dataset for v4 (and v3, which makes sense imo) to the v4 SNVs. It also has the unintended side effect of updating the default dataset for v2 to v4 SNVs. We probably want to keep v2's default SV dataset as `gnomad_sv_r2_1`

- Screenshot 2023-11-06 at 07 27 09

rileyhgrant (changes requested):
Thanks for taking this issue!

I just have one requested change to add some additional logic to keep the v2 page the way that it was.

rileyhgrant (approved):
LGTM! Thanks for doing this.

Since this is so tightly scoped, this seems like a good candidate for the "squash and merge" button, just so you don't have to deal with manually interactive rebasing this down to a single commit.

PR #1269, opened 2023-11-07
Use public HT paths in gene and v4 variant pipelines

mattsolo1:


rileyhgrant (commented):
LGTM, also going to request review from you since I added some commits.

Mostly just cleanup of the GCS buckets and few filepaths.

Most notably, there's a commit that removes the Variant Cooccurrence annotation steps on genes grch38, I believe those values were calculated only for v2, so grch38 should not be annotated with them.

I believe we'll need to re-run the genes pipeline and update the index, the frontend handles the table not rendering on v3 and v4 with a conditional, but the API responds with counts.

mattsolo1:
Oh didn't realize co-occurrence was being added to grch38... weird. Thanks for catching that.

lgtm

rileyhgrant (approved):
LGTM!

Issue #1246, opened 2023-11-01
Variant page site quality metrics histogram crash


rileyhgrant:
For certain (quite possibly all) v4 variants such as this: `/variant/1-55051215-G-GA?dataset=gnomad_r4`

Selecting certain metrics for the Site Quality Metric histogram cause the page to crash.

- `InbreedingCoeff`
- `AS_MS`
- `AS_MQRankSum`

iirc from briefly looking into this, the shape of the response for `AS_MS` and `AS_MQRankSum` are different than the rest, and `InbreedingCoeff` is not there at all.

There was not an immediately clear way to fix this on the frontend only in the time crunch we have, my hunch is it will require editing the variant pipeline to fix the shape of the data, then re-running and re-loading the v4 variant data -- but that remains to be seen.

PR #1270, opened 2023-11-07
Fix variant site quality metrics

mattsolo1:
Currently, the variant site quality plot breaks when metrics such as `AF_FS` or `InbreedingCoeff` are selected.

In the first case, this was because an a new error was added that breaks the page in the event of an undefined quality metric value. This PR allows undefined metrics, and if so, displays "No value for this metric" in the text.

image

Note other quality metrics might be null, so this will hopefully prevent errors with those as well.

See: https://github.com/broadinstitute/gnomad-browser/blob/5ef4ccd65b10afa741c425c779fcb0cbb5f6d214/data-pipeline/src/data_pipeline/datasets/gnomad_v4/types/initial_variant.py#L63

The issue with `InbreedingCoeff` is separate in that this field was renamed to `inbreeding_coeff` by methods team. This PR also includes changes to support the new field.

### DIFF
--- a/browser/src/VariantPage/VariantSiteQualityMetrics.tsx
+++ b/browser/src/VariantPage/VariantSiteQualityMetrics.tsx
@@ -31,6 +31,8 @@ const qualityMetricDescriptions = {
   FS: "Phred-scaled p-value of Fisher's exact test for strand bias.",
   InbreedingCoeff:
     'Inbreeding coefficient as estimated from the genotype likelihoods per-sample when compared against the Hardy-Weinberg expectation.',
+  inbreeding_coeff:
+    'Inbreeding coefficient as estimated from the genotype likelihoods per-sample when compared against the Hardy-Weinberg expectation.',
   MQ: 'Root mean square of the mapping quality of reads across all samples.',
   MQRankSum:
     'Z-score from Wilcoxon rank sum test of alternate vs. reference read mapping qualities.',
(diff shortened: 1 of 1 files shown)


mattsolo1:
Sorry I forget to commit before running formatter so it's tough to see the relevant changes

mattsolo1:
Ah I missed that issue... maybe this won't resolve everything there.

browser/src/VariantPage/VariantSiteQualityMetrics.tsx
```diff
@@ -467,7 +470,8 @@ const getAvailableMetrics = (datasetId: any) => {
   if (isV4(datasetId)) {
     return [
       'SiteQuality',
-      'InbreedingCoeff',
+      'inbreeding_coeff',
+      // 'InbreedingCoeff',
```
rileyhgrant, line 474: Is this commented out metric needed for some reason? If not, it should be removed.

rileyhgrant: Also, it's a bit odd that when selecting the metrics, this metric is the only one in lower case. 

Screenshot 2023-11-07 at 14 22 10

rileyhgrant (changes requested):
Just two nitpicky comments regarding a commented out line, and an observation on inconsistent case in the various quality metrics.

The inconsistent case is maybe better handled at a seperate time, to get this crash fixed sooner.

browser/src/VariantPage/VariantSiteQualityMetrics.tsx
mattsolo1: Yeah... we could keep it in camel case but this what it actually is in the hail table/vcf.

mattsolo1: The casing is all over the map with these metrics

mattsolo1:
Double checking that the specific metrics mentioned in #1246 are fixed in this PR

Using the variant in that PR as an example (`/variant/1-55051215-G-GA?dataset=gnomad_r4`)

`InbreedingCoeff` is fixed

`AS_MQRankSum` looks good. I'm guessing the 0 value in exomes was causing the break.

image

`AS_MS` I'm guessing you meant `AS_FS`... this looks good now too

browser/src/VariantPage/VariantSiteQualityMetrics.tsx
mattsolo1: I'm going to leave it as `inbreeding_coeff` because 

* The browser is setup to display the QC metric names as they are in the data. Changing this would require adding renaming logic
* The other metrics are a mix of camel case, upper snake case, and upper/lower mixed snake case. Since there is  no clear existing convention, it doesn't seem important to enforce one now
* People who see this field in the browser will have an easier time finding it in the hail table if they care to

rileyhgrant (approved):

PR #1274, opened 2023-11-08
Fix export variants to csv button

rileyhgrant:
Resolves https://atgu.slack.com/archives/CNL7NA6H2/p1699389268749339

Fixes a bug in which certain variants that didn't have counts for all populations would cause the "export to csv" button for the variant list to silently fail - printing an error to the console but nothing else.

This adds a simple check to see if the value exists, and if not enters a blank value in the csv.


mattsolo1 (approved):
I tested the export and it works for me.

PR #1279, opened 2023-11-09
Correct UCSC links for v4 SVs

### LINKED ISSUE #1277: UCSC links not working in gnomAD v4 SV's
### What you did:

I was trying to see the UCSC page for the gnomAD SV I was looking at by clicking on the UCSC link. https://gnomad.broadinstitute.org/variant/DEL_CHR5_4DD240ED?dataset=gnomad_sv_r4

I wanted to see the UCSC genome browser page for the appropriate SV in HG38.

I was brought to the correct coordinates in UCSC HG19 instead of HG38. This wasn't even the correct gene.

image
image

elissa-alarmani:
Currently, for the v4 SVs, the UCSC link redirects to the variant's coordinates in UCSC HG19 instead of HG38. This PR implements the inclusion of the two reference genome builds of v2 and v4 to correctly redirect the v4 SVs on the reference genome build of GRCh38 / HG38. 

Fixes #1277

### DIFF
--- a/browser/src/StructuralVariantPage/SVReferenceList.spec.tsx
+++ b/browser/src/StructuralVariantPage/SVReferenceList.spec.tsx
@@ -1,0 +1,20 @@
+import { describe, expect, test } from '@jest/globals'
+import 'jest-styled-components'
+
+import React from 'react'
+import renderer from 'react-test-renderer'
+
+import { forAllDatasets } from '../../../tests/__helpers__/datasets'
+import SVReferenceList from './SVReferenceList'
+import structuralVariantFactory from '../__factories__/StructuralVariant'
+import { svTypes } from '../StructuralVariantList/structuralVariantTypes'
+
+forAllDatasets('SVReferenceList with dataset %s selected', (datasetId) => {
+  describe.each(svTypes)('for SV of type %s', (variantType: string) => {
+    test('has no unexpected changes', () => {
+      const sv = structuralVariantFactory.build({ type: variantType })
+      const tree = renderer.create(<SVReferenceList variant={sv} datasetId={datasetId} />)
+      expect(tree).toMatchSnapshot()
+    })
+  })
+})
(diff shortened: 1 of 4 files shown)


phildarnowsky-broad (changes requested):
Code looks good but I would like to see at least a snapshot test added here since we don't yet have any testing for this component.

When you're writing a test, especially for brand-new code, one big question is what sets of inputs you want to test the code on. The usual ideal is that you'd like to hit each major code path once. 

Looking at this component, we see it has a number of different parameters that might change its behavior: `chrom`, `datasetId`, `pos`, `variant.type`, and so on. But only two of these actually control what path execution takes: specifically, `datasetID` and `variant.type`. That means we would like one test per combination of dataset ID and variant type, but we can use the same arbitrary constant `chrom`, `pos`, etc. in each of those tests.

If you do the arithmetic, you see that that comes to quite a few tests, but luckily, you don't have to write each step of each one individually. We have a test helper you may have encountered called `forAllDatasets`that runs the same set of tests on each dataset in turn, and since we only have a limited number of possible variant types, we can iterate in an inner loop over those.

So to sum up, the new test file I'd like to see here, `SVReferenceList.spec.tsx`, leaving out all the usual boilerplate, should read something like this:

```
forAllDatasets("SVReferenceList with dataset %s selected", () => {
  describe.each(LIST_OF_ALL_POSSIBLE_VARIANT_TYPES)("for SV of type %s", (variantType: string) => {
    test("has no unexpected changes", () => {
      const sv = StructuralVariantFactory.build({type: variantType})
      const tree = renderer.create(<SVReferenceList variant={variant} datasetId={datasetId}
      expect(tree).toMatchSnapshot()
    })
  })
})
```

One last thing: don't forget to actually look over the snapshots to ensure that they're correct, it's surprisingly easy to forget that.

phildarnowsky-broad (approved):

PR #1281, opened 2023-11-09
Correct default coverage metric for v4 CNV and SV pages

### LINKED ISSUE #1280: Incorrect coverage metric on CNV pages
The exome coverage track on CNV pages is displaying the wrong default metric. It should be Over 20 instead of Mean.

image

This is because `isV4` is returning false when checking the CNV dataset id

https://github.com/broadinstitute/gnomad-browser/blob/8ab617ee75c2ea5de466fd1521473b061662c00e/dataset-metadata/metadata.ts#L235

I think it (and the SV dataset) should be added here?

elissa-alarmani:
This PR corrects the default metric issue with the coverage tracks on CNV and SV pages, changing it from `mean` to `Over 20`. The correction involves incorporating the CNV and SV v4 datasetId in the `isV4` logic.

Fixes #1280

### DIFF
--- a/dataset-metadata/metadata.ts
+++ b/dataset-metadata/metadata.ts
@@ -232,7 +232,7 @@ const metadataForDataset = (datasetId: DatasetId): DatasetMetadata => ({
     datasetId === 'gnomad_sv_r4',
   isV2: datasetId.startsWith('gnomad_r2'),
   isV3: datasetId.startsWith('gnomad_r3'),
-  isV4: datasetId.startsWith('gnomad_r4'),
+  isV4: datasetId.startsWith('gnomad_r4') || datasetId.includes('r4'),
   isSVs: datasetId.startsWith('gnomad_sv'),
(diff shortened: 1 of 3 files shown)


dataset-metadata/metadata.ts
```diff
@@ -232,7 +232,7 @@ const metadataForDataset = (datasetId: DatasetId): DatasetMetadata => ({
     datasetId === 'gnomad_sv_r4',
   isV2: datasetId.startsWith('gnomad_r2'),
   isV3: datasetId.startsWith('gnomad_r3'),
-  isV4: datasetId.startsWith('gnomad_r4'),
+  isV4: datasetId.startsWith('gnomad_r4') || datasetId.includes('r4'),
```
mattsolo1, line 235: The left operand can be removed (`startsWith`...) now since it is made redundant by the right one.

mattsolo1 (approved):
Just that one change then looks good

PR #1282, opened 2023-11-09
Support viewing mitochondrial variants when gnomad v4 is selected

### LINKED ISSUE #1278: mtDNA variants should be part of v4

mattsolo1:
Currently if a users queries a variant, gene, or region in the mitochondrial dataset (which was originally part of v3 genomes), they get an error saying mito variants not available for v4. Since v3 is now part of v4, this should not happen and they should see mito variants on the v4 dataset.

Furthermore the coverage track component required modification b/c when v4 is the current dataset, the `over_20` metric is default, which mito coverage does not have. This PR adds a new prop so that the metric can specified as needed; doing this will override the metric determined by the `datasetId` and the default metric (`mean`).

Examples:

Variant: http://localhost:8008/variant/M-4491-G-A?dataset=gnomad_r4
Gene: http://localhost:8008/gene/ENSG00000198763?dataset=gnomad_r4
Region: http://localhost:8008/region/M-4470-5511?dataset=gnomad_r4

Fixes #1278

### DIFF
--- a/graphql-api/src/queries/mitochondrial-coverage-queries.ts
+++ b/graphql-api/src/queries/mitochondrial-coverage-queries.ts
@@ -7,6 +7,7 @@ import { extendRegions, mergeOverlappingRegions, totalRegionSize } from './helpe
 import { assertDatasetAndReferenceGenomeMatch } from './helpers/validation-helpers'
 
 const COVERAGE_INDICES = {
+  gnomad_r4: 'gnomad_v3_mitochondrial_coverage',
   gnomad_r3: 'gnomad_v3_mitochondrial_coverage',
 }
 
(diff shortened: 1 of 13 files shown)


graphql-api/src/queries/mitochondrial-coverage-queries.ts
```diff
@@ -7,6 +7,7 @@ import { extendRegions, mergeOverlappingRegions, totalRegionSize } from './helpe
 import { assertDatasetAndReferenceGenomeMatch } from './helpers/validation-helpers'
 
 const COVERAGE_INDICES = {
+  gnomad_r4: 'gnomad_v3_mitochondrial_coverage',
```
rileyhgrant, line 10: This makes complete sense, especially right now -- just commenting to say this is maybe something to mentally flag for the future. Coming back to this without context could be confusing.

rileyhgrant (approved):
LGTM!

Nicely done on getting this fixed so quick.

PR #1291, opened 2023-11-14
Update v4 Download checksums

rileyhgrant:
Resolves https://atgu.slack.com/archives/CNL7NA6H2/p1699945990840919

- Updates a checksum for v4 genomes.

- Adds md5 checksums to v4 SVs that have it (I don't know enough about checksums to know why some, but not all of the v4 SV VCFs have md5 checksums, since not all have it, crc32c checksums are used instead)


rileyhgrant:
Closed in favor of #1296

PR #1292, opened 2023-11-14
Handle large genes

### LINKED ISSUE #1289: Determine how to effectively serve variant data for huge genes (TTN, etc)
Currently, genes with huge amounts of variants like TTN are problematic - due to the increase in the number of total variants in v4, there are now too many variants to return before it hits our query timeout. Thus, the results never get cached, and that query can block other users queries from running while it is working.

The current stopgap solution for TTN, a common large gene, is to not load the variant table at all. There are likely other, similarly large genes that cause problems that we don't single out like TTN, and not serving the data for these is not ideal in the first place.

Some potential avenues discussed:
- As a better temporary solution, alter the variant query to filter down to loss of function and missense variants if we can tell that the list is going to be too large to effectively return in a reasonable amount of time
- Region pages handle this by running a smaller query to see how many variants exist in the region, and if it is >30,000, display a message instead (as a side note, this functionality may be broken, last I checked)
- As a long term, look at how the data is structured to try and address the underlying slowness of these large queries. Caching is treating us well, but it is likely only one part of the solution.

mattsolo1:
Adds an optional JSON cache client that can read/write to local storage or google cloud storage.

Can be configured to cache all database lookups using the `withCache` function or just specific genes in the large genes list.

The proposed configuration for production is to only use the client for TTN lookup from GCS (see: deploy/manifests/browser/base/api.deployment.yaml).

Demo: http://34.117.225.187/gene/ENSG00000155657?dataset=gnomad_r4

Resolves #1289


sjahl:
@mattsolo1 Just to make sure I'm understanding correctly, we:

- pre-compute and store the JSON in a GCS bucket
- at runtime, we check if we have the file on disk, and download if not

So, the local storage does not need to be shared between pods, and ephemeral is OK?

graphql-api/src/queries/helpers/json-cache.ts
```diff
@@ -0,0 +1,190 @@
...
+  private useCompression = false
+
+  filePathFromCacheKey(key: string): string {
+    const gzipExt = this.useCompression ? ".gz" : ""
+    const fileName = `${key.replace(/:/g, "-")}.json${gzipExt}`
+    return join(this.json_cache_path, fileName)
+  }
+
+  constructor(json_cache_path: string, compression = false) {
+    this.json_cache_path = json_cache_path
```
sjahl, line 132: This is my limited typescript experience speaking -- but I think I'm reading that we fall back to a local filepath if we're _not_ using a bucket? What happens if the json cache path variable is empty? Is it writing files in the root (`/`) directory of the container?

sjahl (commented):
Just the one inline question on what filepath we default to when not using a bucket (or, simply just wondering what exactly happens when nothing is specified).

It might be helpful to have some docs written down somewhere with information on what needs to happen when we need to invalidate the cache, unless we don't need to worry about that.

graphql-api/src/queries/helpers/json-cache.ts
mattsolo1: If `gs://` is in the file path, it will use a bucket. If not, it will use a local file path. That could probably be more clear.

Since the constructor only accepts a string (and not null/undefined), it will never be empty.

mattsolo1: I agree about the need for docs of some kind.

sjahl: OK, yeah, i'd say some docs on how to determine if the JSON cache needs invalidation and the preferred procedure to clear it would be good, but i don't feel like that needs to block merging the code

https://github.com/broadinstitute/gnomad-browser/issues/1331

Issue #1259, opened 2023-11-03
Switching the dropdown from v2.1/GRCh37 to v4/GRCh38 should link you to the liftover coordinates


sjahl:
Example: https://gnomad.broadinstitute.org/variant/11-118344839-C-T?dataset=gnomad_r2_1

Selecting v4 from the dropdown on the top-right, results in a "variant not found", but it should probably link to the lifover coordinates, which are here: https://gnomad.broadinstitute.org/variant/11-118474124-C-T?dataset=gnomad_r4

when looking at the v2 variant, we know what the v4 variant is:

<img width="523" alt="Screenshot 2023-11-03 at 11 24 50 AM" src="https://github.com/broadinstitute/gnomad-browser/assets/636687/a9b6438b-7c4f-4df3-b7a7-b70ce23654a0">

Could we make the dropdown in the top right bring you to the same variant we link under Related Variants, Liftover?

<img width="523" alt="Screenshot 2023-11-03 at 11 25 33 AM" src="https://github.com/broadinstitute/gnomad-browser/assets/636687/9ede0459-d09a-4572-8eaa-a2e189dabda3">

PR #1293, opened 2023-11-14
Take liftover into account when switching from v2 to v4 variants

phildarnowsky-broad:
This adds a disambiguation page for liftover, similar to the existing one for RSIDs, since as with RSIDs, a v2 variant may lift over to multiple v4 variants, and vice versa. Switching to v4 via the dataset selector on the variants page will link to this new disambiguation page.

### DIFF
--- a/browser/src/VariantPage/LiftoverDisambiguationPage.tsx
+++ b/browser/src/VariantPage/LiftoverDisambiguationPage.tsx
@@ -1,0 +1,140 @@
+import React from 'react'
+import {
+  DatasetId,
+  labelForDataset,
+  referenceGenome,
+  isLiftoverSource,
+  baseDatasetForReferenceGenome,
+} from '@gnomad/dataset-metadata/metadata'
+import { TrackPage, TrackPageSection } from '../TrackPage'
+import DocumentTitle from '../DocumentTitle'
+import { BaseQuery } from '../Query'
+import Delayed from '../Delayed'
+import StatusMessage from '../StatusMessage'
+import Link from '../Link'
+import { Redirect } from 'react-router-dom'
+
+type LiftoverDisambiguationPageProps = {
+  fromVariantId: string
+  fromDatasetId: DatasetId
+  toDatasetId: DatasetId
+}
+
+type CorrespondingVariantField = 'liftover' | 'source'
+
+const LiftoverDisambiguationPage = ({
+  fromVariantId,
+  fromDatasetId,
+  toDatasetId,
+}: LiftoverDisambiguationPageProps) => {
+  const fromLabel = labelForDataset(fromDatasetId)
+  const fromReferenceGenome = referenceGenome(fromDatasetId)
+  const toReferenceGenome = referenceGenome(toDatasetId)
+
+  const operationName = 'LiftoverDisambiguation'
+  const correspondingVariantField: CorrespondingVariantField = isLiftoverSource(fromDatasetId)
+    ? 'liftover'
+    : 'source'
+  const disambiguationQuery = `
+query ${operationName}($source_variant_id: String, $liftover_variant_id: String, $reference_genome: ReferenceGenomeId!) {
+  liftover(source_variant_id: $source_variant_id, liftover_variant_id: $liftover_variant_id, reference_genome: $reference_genome) {
+	${correspondingVariantField} {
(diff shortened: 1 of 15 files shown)


phildarnowsky-broad:
closes #1259

mattsolo1:
@phildarnowsky-broad do you have an example variant that shows the disambiguation page?

mattsolo1 (commented):
The selector works great. I'm having trouble finding a variant where I encounter the disambiguation page.

I'll note the dataset selector is now inconsistent with other parts of the site. Namely it doesn't show any v3 datasets and if you are v4 you can't see any GRCh37 datasets. I'd be in favor of adding the v2 -> v4 lift over in the selection menu and then making the dataset selector consistent; otherwise it's unclear why some datasets are randomly missing when you're on the variant page.

image

phildarnowsky-broad:
@mattsolo1 the missing datasets are a bug: updated code to follow

phildarnowsky-broad:
@mattsolo1 @ch-kr per conversation in Slack I'm going to give interested parties a chance to argue for or against changing the dataset selector as you describe.

mattsolo1:
> the missing datasets are also missing in prod, this is from a v4 variant

The screenshot you show is on a GRCh38 variant (so it shoes v4 and v3); my screenshot is from a GRCh37 variant of your new branch (so it shows V4 and V2, but not V3).

Currently the browser, when on the variant page, limits what datasets you can select (presumably because lift over was not used by the dataset selector). Now that you've added liftover support, my point is we can now feel free to add all the datasets as options to the selector.

phildarnowsky-broad:
@mattsolo1 ah, I think I was misunderstanding you--I thought you were talking about this as something that had worked and was now broken, but this would be new behavior actually? if so, makes sense to me: I'll add in the reverse liftover and remove it if anyone really doesn't like it.

ch-kr:
I slightly prefer the lifting in a single direction only to push users to move from b37 to b38, though I guess we should support the other direction if not having liftunder from 38 back to 37 is really impacting clinical pipelines

mattsolo1:
@ch-kr I don't think we should choose not add low hanging fruit functionality for that reason. We don't know why they may want to lift back and forth between new and old variants. Perhaps they are trying to cross reference new literature with old literature, etc.

ch-kr:
that's a good point! works for me

browser/src/VariantPage/LiftoverDisambiguationPage.tsx
```diff
@@ -0,0 +1,140 @@
...
+                        >
+                      ) => {
+                        const variantId =
+                          correspondingVariant[correspondingVariantField]!.variant_id
+                        const datasetToLink = baseDatasetForReferenceGenome(toReferenceGenome)
+                        return (
+                          <li key={variantId}>
+                            <Link to={`/variant/${variantId}?dataset=${datasetToLink}`}>
+                              {variantId}
+                            </Link>
```
rileyhgrant, line 122: When I was manually checking here (set the conditionals above to always be false so this branch triggers) this link is broken for me, it appends the new dataset on top of the v2 dataset from where it came from (i.e. directed to "http://localhost:8008/variant/1-55039779-T-G?dataset=gnomad_r4?dataset=gnomad_r2_1"), this double query param results in an error page.

I believe the `Link` component has a default value for a prop called `preserveSelectedDataset` as true, to keep the dataset query param when navigating between pages. 

https://github.com/broadinstitute/gnomad-browser/blob/d9085aa14f846705598d80cabfcae813ee7e0cbb/browser/src/Link.tsx#L22-L45

As such, just passing false to this will fix the link.

```
<Link
  to={`/variant/${variantId}?dataset=${datasetToLink}`}
  preserveSelectedDataset={false}
>
  {variantId}
</Link>

rileyhgrant (changes requested):
Very clean and nice code quality. Loving all the tests.

Just one requested change to pass a param that should fix broken links observed when the variant disambiguation page gets used.

phildarnowsky-broad:
@mattsolo1 added a small commit to switch on the reverse liftover, I'm looking currently for a multiple-variant liftover and will post a link when I find one

rileyhgrant (approved):
LGTM, with squashes of fixup commits of course.

PR #1294, opened 2023-11-14
restore rate limits to pre-release values

sjahl:
As discussed today, this restores the ES concurrency/rate/query cost limits to the pre-release values.


rileyhgrant (approved):
LGTM

PR #1301, opened 2023-11-16
Update Stats page to allow easy download of graphs

### LINKED ISSUE #1265: Make Stats page bar charts easily downloadable
Similar to the tables on the Stats page, it has been suggested to make the bar charts easily downloadable. This is very easy to do with the package that was added for the tables. 

Filing this issue so it doesn't get lost.

rileyhgrant:
Resolves #1265 

Adds a button to allow easy downloads of graphs on the stats page

### DIFF
--- a/browser/src/StatsPage/DownloadFigure.tsx
+++ b/browser/src/StatsPage/DownloadFigure.tsx
@@ -1,0 +1,50 @@
+import html2canvas from 'html2canvas'
+import React from 'react'
+// @ts-expect-error
+import DownloadIcon from '@fortawesome/fontawesome-free/svgs/solid/file-download.svg'
+import styled from 'styled-components'
+
+export const downloadElementAsPNG = (elementID: string): void => {
+  const figure = document.getElementById(elementID)
+
+  if (figure) {
+    html2canvas(figure).then((canvas) => {
+      const link = document.createElement('a')
+      link.download = `gnomad_${elementID}.png`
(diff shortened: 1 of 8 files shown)


mattsolo1:
Works as expected!

The download button is huge, and most visitors will not want to download anything. What do you think about adding a more subtle icon? We already have an icon package installed (font awesome). Here are some options: https://fontawesome.com/search?q=download&o=r&m=free. I was thinking maybe it would fit in the legend or next to the title (see below)?

A few other general points about the plot (not related to downloading). @sambaxter do you have feedback?

- Normally people expect graphs to have titles. Perhaps something like "Sample size across major ExAC/gnomAD releases"
- Would be cool if the year of each was release was included below the release title to emphasize the long history of the project (2014, 2016, 2019, 2023)
- There is no Y axis label... should it be "Number of samples"?
- There is a point where the text overlaps:

image

Could set the minimum width of this plot to be larger?

rileyhgrant:
Thanks for the feedback, the button is pretty huge and imposing. I'll make it more subtle and incorporate your other suggestions.

rileyhgrant:
Updated with:
- Smaller download button with icon (next to title)
- Added prop for title so all bar graphs have a title
  - id of element and download are now generated by slugifying the title of the graph
- Cleaning up: defined y axis for graphs, used same small download icon on tables, added titles to all graphs

Screenshot 2024-01-02 at 15 06 07

PR #1303, opened 2023-11-17
Zoom in/out on-click to enable faster review of STR readviz images

bw2:
This type of UI has been working well in https://github.com/broadinstitute/flipbook

### DIFF
--- a/browser/src/ShortTandemRepeatPage/ShortTandemRepeatReads.tsx
+++ b/browser/src/ShortTandemRepeatPage/ShortTandemRepeatReads.tsx
@@ -26,9 +26,10 @@ const ShortTandemRepeatReadImage = styled.img`
 
   &.zoomedIn {
     position: absolute;
-    left: 10px;
+    left: 0;
     display: block;
     cursor: zoom-out;
+    padding: 10px;
   }
 `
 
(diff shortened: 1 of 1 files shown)


mattsolo1:
Basically the reads are shown in a zoomed out fashion first making it easier to see? Makes sense.

However when the user zooms in I found it a bit nicer to have the zoomed in reads contained in a scrollable container with `overflow-x`. Why not stick with that?

Aside: I found it a bit janky when changing the min/max repeats options, the reads disappear and the user would be sent back up the page b/c there is transiently no content where the reads should be. Is it possible display a placeholder box that would prevent this jarring reflow from happening?

bw2:
I think the pros of the proposed implementation are that it prevents the side margins from eating a lot of useful screen space in the zoomed-in use case. Also, on mobile, horizontal scroll bars aren't as user friendly. Finally, it allows people to optionally use the browser's zoom settings to set a custom zoom level. 

I'm not seeing the reflow or jumpiness in the browsers I've tested on 
https://gnomad.broadinstitute.org/short-tandem-repeat/HTT?dataset=gnomad_r3
How can I reproduce this?

mattsolo1 (approved):

mattsolo1:
@bw2 feel free to squash/merge when you think this is ready.

PR #1311, opened 2023-11-20
update genes pipeline - garch38 extract transcript genes_path

ignatiusm:
It is really cool to see all the GnomAD v4 development! There was a commit last week that removed a stage from the grch38 part of the `genes` pipeline. I think there was a remnant of the old numbering scheme left in the code (it was looking for `step_6` when it had been removed). This prevents the pipeline from running.

I have changed the `genes_path` to point to `annotate_grch38_genes_step_4` (pattern matching from the grch37 approach) however this may be incorrect. You may be wanting to extract transcripts from the outputs of a different step (e.g. after reject_par_y_genes / step_5)! Please review this as I have very little understanding of the pipeline steps at the moment!

### DIFF
--- a/data-pipeline/src/data_pipeline/pipelines/genes.py
+++ b/data-pipeline/src/data_pipeline/pipelines/genes.py
@@ -377,7 +377,7 @@ pipeline.add_task(
     "extract_grch38_transcripts",
     extract_transcripts,
     f"/{genes_subdir}/transcripts_grch38_base.ht",
-    {"genes_path": pipeline.get_task("annotate_grch38_genes_step_6")},
+    {"genes_path": pipeline.get_task("annotate_grch38_genes_step_4")},
 )
 
 ###############################################


ignatiusm:
@rileyhgrant @mattsolo1

rileyhgrant (changes requested):
Aha! Nice catch on us slipping up on this -- thanks. You're right on us forgetting to update the referenced step when extracting transcripts.

As you mentioned, I think in this step we will indeed want to use `step_5` instead of `step_4` so that we extract transcripts from the genes that have been filtered by that last step. So I'm requesting just that minor change.

Thanks for catching this :)

ignatiusm:
> Aha! Nice catch on us slipping up on this -- thanks. You're right on us forgetting to update the referenced step when extracting transcripts.
> 
> As you mentioned, I think in this step we will indeed want to use `step_5` instead of `step_4` so that we extract transcripts from the genes that have been filtered by that last step. So I'm requesting just that minor change.
> 
> Thanks for catching this :)

Thanks for the review @rileyhgrant . I have addressed your request in 09e8042. Ready for re-review.

rileyhgrant (approved):
LGTM, thanks again!

PR #1318, opened 2023-11-21
Fix Gene page "export to CSV" genetic ancestry group data mismatch

### LINKED ISSUE #1307: mismatched v4 variant data between downloaded file and web-displayed info
</style>
</head>

<body link="#0563C1" vlink="#954F72">

16 | 2089662
-- | --

</body>

</html>

* For concerns about data, check the frequently asked questions
  (https://gnomad.broadinstitute.org/help#frequently-asked-questions) or
  send an email to gnomad@broadinstitute.org.

-->

rileyhgrant:
Resolves #1307 

On the gene page there is a button that allows users to export the variant list to a `.csv`. As of v4, the variants in the list can have a differing set of genetic ancestry groups (e.g. some variants may have `AC`, `AN`, etc, for `mid` while others may not, this is because of v4 using v3 for its genomes, and its own values for its genomes).

The gene page combines the genome and exome level data variants in an early processing step, presumably for ease of exporting. It accomplishes this by accessing values based on their position in a list. Now that the exomes and genomes populations can differ in size and order, this export produces a list with incorrect values for v4 variants that occur in both exomes and genomes (e.g. taking `[4]` from genomes is Amish, but `[4]` in exomes is Middle Eastern).

This PR changes the logic to make it more general to fix this bug.

- Removes the assumption that the 0th element of the array's genetic ancestry groups are representative of every element in the array (which is a faulty assumption for v4 variants that are in both exomes and genomes)
- Combines genetic ancestry group values using the `id` field, rather than combining based purely on index

It also adds tests to confirm this behavior.

### DIFF
--- a/browser/src/VariantList/ExportVariantsButton.tsx
+++ b/browser/src/VariantList/ExportVariantsButton.tsx
@@ -1,10 +1,80 @@
 import React from 'react'
 
 import { Button } from '@gnomad/ui'
+import { Population } from '../VariantPage/VariantPage'
 
-import { GNOMAD_POPULATION_NAMES, PopulationId } from '@gnomad/dataset-metadata/gnomadPopulations'
+import {
+  GNOMAD_POPULATION_NAMES,
+  PopulationId,
+  populationsInDataset,
+} from '@gnomad/dataset-metadata/gnomadPopulations'
+import { DatasetId, isExac, isV2, isV3, isV4 } from '@gnomad/dataset-metadata/metadata'
 
-const exportVariantsToCsv = (variants: Variant[], datasetId: any, baseFileName: any) => {
+type Column = {
+  label: string
+  getValue: (variant: VariantTableVariant) => string
+}
+
+type Property = 'ac' | 'an' | 'ac_hemi' | 'ac_hom'
+
+const getValueGivenProperty = (
+  popId: PopulationId,
+  variant: VariantTableVariant,
+  property: Property
+) => {
+  return variant.populations.filter((v) => v.id === popId)[0]
+    ? JSON.stringify(variant.populations.filter((v) => v.id === popId)[0][property])
+    : ''
+}
+
+export const createPopulationColumns = (datasetId: DatasetId) => {
+  /* eslint-disable no-nested-ternary */
+  const topLevelDataset = isV4(datasetId)
+    ? 'v4'
+    : isV3(datasetId)
+    ? 'v3'
+    : isV2(datasetId)
+    ? 'v2'
+    : isExac(datasetId)
+    ? 'ExAC'
+    : 'default'
+  /* eslint-enable no-nested-ternary */
+
+  const datasetPopulations: PopulationId[] = populationsInDataset[topLevelDataset]
+
+  let populationColumns: Column[] = []
+  datasetPopulations.forEach((popId: PopulationId) => {
+    const popName = GNOMAD_POPULATION_NAMES[popId]
+
+    populationColumns = populationColumns.concat([
+      {
+        label: `Allele Count ${popName}`,
+        getValue: (variant: VariantTableVariant) => getValueGivenProperty(popId, variant, 'ac'),
+      },
+      {
+        label: `Allele Number ${popName}`,
+        getValue: (variant: VariantTableVariant) => getValueGivenProperty(popId, variant, 'an'),
+      },
+      {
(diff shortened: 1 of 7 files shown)


browser/src/VariantList/ExportVariantsButton.tsx
```diff
@@ -1,10 +1,79 @@
...
+
+  let populationColumns: Column[] = []
+  datasetPopulations.forEach((popId: PopulationId) => {
+    const popName = GNOMAD_POPULATION_NAMES[popId]
+
+    populationColumns = populationColumns.concat([
+      {
+        label: `Allele Count ${popName}`,
+        getValue: (variant: VariantTableVariant) =>
+          variant.populations.filter((v) => v.id === popId)[0]
```
phildarnowsky-broad, line 41: Looks like the same logic apart from column name in a number of places, let's take this opportunity to extract a helper function to make that clearer.

phildarnowsky-broad (changes requested):

browser/src/VariantList/mergeExomeAndGenomeData.spec.ts
```diff
@@ -0,0 +1,148 @@
...
+
+describe('mergeExomeAndGenomePopulationData', () => {
+  it('returns expected values when exomes and genomes have the same populations', () => {
+    const geneticAncestryGroupObjects = createAncestryGroupObjects([
+      { id: 'afr', value: 1 },
+      { id: 'remaining', value: 2 },
+      { id: 'eur', value: 4 },
+    ])
+
+    const mockVariant = variantFactory.build({
```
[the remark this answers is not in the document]
rileyhgrant: Ah, interesting, That makes sense!

browser/src/VariantList/mergeExomeAndGenomeData.ts
```diff
@@ -1,6 +1,48 @@
...
+        id: genomePopulation.id,
+        ac: genomePopulation.ac,
+        an: genomePopulation.an,
+        ac_hemi: genomePopulation.ac_hemi,
+        ac_hom: genomePopulation.ac_hom,
+      }
+    }
+  })
+
+  return Object.values(populations).map((population: Population) => population)
```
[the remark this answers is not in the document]
rileyhgrant: Oops, yeah good call.

browser/src/VariantPage/VariantPage.tsx
```diff
@@ -148,7 +148,7 @@ export type Population = {
   ac_hemi: number | null
   ac_hom: number
   homozygote_count?: number
-  hemizygote_count?: number | null
+  hemizygote_count?: number
```
phildarnowsky-broad, line 151: Ah what I meant here was let's use the `| null` rather than `?` in both counts--that's assuming that we're getting those values directly from GraphQL, if not then never mind.

phildarnowsky-broad (changes requested):
One last small change due to me being unclear earlier

browser/src/VariantPage/VariantPage.tsx
rileyhgrant: Aha, I see, yeah my edit was unclear also. The query for the Variant page doesn't actually request the `hemizygote_count` and `homozygote_count` fields, so I added the `?` to let me Type the populations here without having to add a dummy value for those two. So maybe the true type should actually be `hemizygote_count?: number | null`?

```
populations {
  id
  ac
  an
  ac_hemi
  ac_hom
}
``` 

Digging a little deeper, looks like `ac_hemi` and `ac_hom` are aliased names of `hemizygote_count` and `homozygote_count` that you can query to get the same value. Looks like ClinVar names them `ac_hemi` etc, so it made sense to add the alias in our API at some point in the past to match.

Unsure of what the best thing to do right now is. It feels like it should be safe to remove `hemizygote_count` and `homozygote_count` from this type (and a brief run of tests and typecheck after doing that goes fine), but I also feel like maybe that should be a seperate PR? It feels different in scope than this one, but I can also see the argument that this PR concerns this type and these files.

rileyhgrant: Ok nice, that makes sense to me.

Added that commit and going to formally re-request review! I did a force push so the changes since last reviewed might not work, but all the PR review commits are tagged as fixup.

phildarnowsky-broad (approved):

PR #1325, opened 2023-11-28
adds a service account and deployctl bits for annotating it

sjahl:
Bits for managing a gnomad-api kubernetes service account and annotating it with the (expected) name of a GCP service account that is managed via terraform.

I had to exclude the k8s ServiceAccount object from being suffixed per deployment with kustomize, because GKE workload identity can only bind to an exact serviceaccount within a namespace. e.g. I can only bind to "namespace/service-acct-name", and not to "namespace/service-account-name-*". Which means it cannot handle the case where we have service accounts with a deployment suffix like "service-account-name-blue" or "service-account-name-green"

The result is that all deployments need to share the same service account, which i think is a reasonable workaround in this case.

The alternative is that we can create the kubernetes service account in terraform (like we do with the elasticsearch snapshots account), but I'm worried that creating an implicit dependency between a service account in terraform and the app deployment is going to create too many problems in our more dynamic browser deployments...

### DIFF
--- a/deploy/deployctl/subcommands/browser_deployments.py
+++ b/deploy/deployctl/subcommands/browser_deployments.py
@@ -25,6 +25,14 @@ images:
   - name: gnomad-browser
     newName: {browser_image_repository}
     newTag: '{browser_tag}'
+patches:
+  - patch: |-
+      - op: add
+        path: /metadata/annotations/iam.gke.io~1gcp-service-account
+        value: '{service_account}@{project}.iam.gserviceaccount.com'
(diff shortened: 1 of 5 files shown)


sjahl:
the terraform for adding the service account and bucket to the dev project are merged, so this should be good to use, when deploying to my `gnomad-sj` cluster. The bucket that the api service account has access to is `gs://gnomadev-test-gene-cache`

mattsolo1 (approved):

PR #1329, opened 2023-11-30
Improve error page

rileyhgrant:
Adds a text entry field to the error page, and includes the error message in the message body so that if someone changes the title  the original error will still be there.

Example error page for possible local testing: http://localhost:8008/gene/ENSG00000169174?dataset=gnomad_r3?dataset=gnomad_r4

---

- Text entry field on page (rather utilitarian right now, will style if this is a good solution)
  - 
- Text entry carries through to submitting an inssue
  - Screenshot 2023-11-30 at 13 49 07


rileyhgrant:
Direct away from email and to the browser section on the forum

rileyhgrant:
The added Discourse link creates a topic under the "Browser" category with a tag of "Bug". It auto populates the create topic popup with all the information that goes to Github issues or email.

Screenshot 2023-12-04 at 12 53 33

The wording of the error page has been changed in the hopes of directing users primarily to Github and Discourse.
Screenshot 2023-12-04 at 12 56 03

browser/src/ErrorBoundary.tsx
```diff
@@ -57,14 +63,45 @@ Browser: ${navigator.userAgent}
           <DocumentTitle title="Error" />
           <PageHeading>Something Went Wrong</PageHeading>
           <p>An error prevented this page from being displayed.</p>
+          <p>This is a bug.</p>
+          <p>
+            Please describe what you did to cause this page crash
```
phildarnowsky-broad, line 68: Let's soften this wording a bit, even sophisticated users like our audience can get defensive about their computer skills and to be fair, if the page is crashing, that's probably something _we_ did to cause it.

How about "Please describe what you were trying to do at the time the page crashed"?

phildarnowsky-broad (changes requested):
Codewise looks OK to me but I have a suggestion regarding some of the copy.

browser/src/ErrorBoundary.tsx
rileyhgrant: Aha, yep -- very good call.

phildarnowsky-broad (approved):

PR #1334, opened 2023-12-01
add a stategic patch merge to generate the cache bucket name

sjahl:
Allows for matching the cache bucket name with the cluster name / env tag

### DIFF
--- a/deploy/deployctl/subcommands/browser_deployments.py
+++ b/deploy/deployctl/subcommands/browser_deployments.py
@@ -29,10 +29,24 @@ patches:
   - patch: |-
       - op: add
         path: /metadata/annotations/iam.gke.io~1gcp-service-account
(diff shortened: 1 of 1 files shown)


sjahl:
Don't merge, I need to switch to kustomize syntax that isn't deprecated

sjahl:
@mattsolo1 OK, i think this is ready for review. Note that the following two kustomize deprecation warnings started to fire:
- https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/vars/
- https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/patchesstrategicmerge/

And these needed to be fixed, since they'll break pretty soon if anyone on our team happens to upgrade to a kubectl CLI version that includes a kustomize `>5.0` in it.

rileyhgrant (approved):

PR #1340, opened 2023-12-05
switch to a GKE initiated managedcertificate

sjahl:
We previously created a managedcertificate out of band, in setup.py. This switches to the GKE built-in managedcertificate object. This removes a chicken and egg scenario where we couldn't deploy the app until a cert was present, and couldn't issue the cert until there was an app present.


sjahl:
Sadly, I don't know if this is going to be able to be deployed without some kind of downtime. It's optimized for the first deployment, where you're probably not urgently waiting for the cert to be provisioned while your site is down.

Roughly, in my experience, it takes around 20 minutes from the time that the ManagedCertificate resource being applied, to the time that it's associated with the ingress. This only happens the first time, or after we delete the ManagedCertificate resource. Subsequent updates to the ingress don't trigger recreating the cert.

The alternative here is to analog the old setup script, and create a separate managed certificate with terraform: https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_managed_ssl_certificate

But that terraform provider seems to come with a strongly worded warning that highlights some of the chicken-and-egg issues that I mentioned in the PR description.

I'd personally opt to use the changes from this PR over that, since waiting 20 minutes in rare circumstances seems like a better option than having a deploy process that has a convoluted number of steps when we have to start from scratch, but i'm interested to hear what others think.

rileyhgrant (approved):
Looks good to my eyes!

PR #1347, opened 2023-12-06
support passing in an alternate es cluster name

### LINKED ISSUE #1346: Support providing an alternate name for elasticsearch when generating kustomize configs
Our `deployctl elasticsearch apply` subcommand supports providing a custom name for the elasticsearch cluster. However, when creating a deployment with Kustomize, we didn't have that option. That means if you wanted to use the custom ES name that you chose when creating the ES cluster, you have to manually update the deployment manifests to point the API pods to the right ELASTICSEARCH_HOST and _PASSWORD secret.

part of #1343

sjahl:
This adds a `--es-cluster-name` parameter to the deployments `create` command, which lets you specify which ES cluster you'd like to use when generating your deployment, and defaults to `gnomad` if that's unspecified.

closes #1346

### DIFF
--- a/deploy/deployctl/subcommands/browser_deployments.py
+++ b/deploy/deployctl/subcommands/browser_deployments.py
@@ -58,6 +58,37 @@ patches:
                     value: 'gs://{cluster_name}-gene-cache/2023-12-01'
 """
 
+ALTERNATE_ES_CLUSTER_NAME = """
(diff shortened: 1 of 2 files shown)


sjahl:
Obsolete now that we are managing the ES deployment entirely with Kustomize

PR #1348, opened 2023-12-06
don't define reads services if reads-deployment isn't specified

### LINKED ISSUE #1342: Allow for demo deployments to optionally include the reads deployment
It's now a common occurrence for us to be deploying browser demos on clusters that are _not_ the production cluster. In many of these cases, there is no reads deployment, unless reads is the component that is being demoed. This creates friction when trying to create the demo ingress, since it expects there to be a reads service. The workaround is simply to attempt to create the ingress again, which works a second time, because the first attempt creates the (bogus) reads Service objects before it ultimately fails.

We should have an option in our demo ingress / kustomize config to only create the reads service when desired.

sjahl:
closes #1342 

We don't always have reads deployments on demos which aren't on the production cluster. This changes the behavior of the demo ingress creation script to only create reads services and Ingress paths if you specify a reads deployment.

### DIFF
--- a/deploy/deployctl/subcommands/ingress_demo.py
+++ b/deploy/deployctl/subcommands/ingress_demo.py
@@ -21,6 +21,9 @@ spec:
   ports:
     - port: 80
       targetPort: 80
+"""
+
+READS_SERVICE_MANIFEST_TEMPLATE = """
 ---
 apiVersion: v1
 kind: Service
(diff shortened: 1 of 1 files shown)


sjahl:
Superseded by https://github.com/broadinstitute/gnomad-browser/pull/1634

PR #1349, opened 2023-12-06
Update variant page in silico predictor display

### LINKED ISSUE #1345: Add negative warning/danger thresholds for Pangolin
Kathryn Russell reported that we didn't give a warning/danger thresholds for negative Pangolin scores (for splice loss). 

Could we add a second set of thresholds for Pangolin? 

image
```
  pangolin_largest_ds: { label: 'Pangolin', warningThreshold: 0.2, dangerThreshold: 0.5 },

```
To add: 
```
  pangolin_largest_ds: { label: 'Pangolin', warningThreshold: -0.2, dangerThreshold: -0.5 },

```

rileyhgrant:
Resolves #1345

- Re-adds a note referring users to SpliceAI Lookup that was mistakenly dropped after being included in mockups
- Updates Pangolin scores warning and danger thresholds to be based off of the absolute value of the score so negative (splice loss) predictions get correctly colored

Example localhost link to variant [http://localhost:8008/variant/7-66991136-C-T?dataset=gnomad_r4](http://localhost:8008/variant/7-66991136-C-T?dataset=gnomad_r4)

### DIFF
--- a/browser/src/VariantPage/VariantInSilicoPredictors.tsx
+++ b/browser/src/VariantPage/VariantInSilicoPredictors.tsx
@@ -20,13 +20,43 @@ const FLAG_DESCRIPTIONS = {
 }
 
 const PREDICTORS_V4 = {
-  revel_max: { label: 'REVEL', warningThreshold: 0.644, dangerThreshold: 0.773 },
-  spliceai_ds_max: { label: 'SpliceAI', warningThreshold: 0.2, dangerThreshold: 0.5 },
-  pangolin_largest_ds: { label: 'Pangolin', warningThreshold: 0.2, dangerThreshold: 0.5 },
-  phylop: { label: 'phyloP', warningThreshold: 7.367, dangerThreshold: 9.741 },
-  sift_max: { label: 'SIFT (max)', warningThreshold: 0.001, dangerThreshold: 0 },
-  polyphen_max: { label: 'PolyPhen (max)', warningThreshold: 0.978, dangerThreshold: 0.999 },
-  cadd: { label: 'CADD', warningThreshold: 25.3, dangerThreshold: 28.1 },
+  revel_max: {
+    label: 'REVEL',
+    warningThreshold: 0.644,
+    dangerThreshold: 0.773,
+    absoluteValue: false,
+  },
+  spliceai_ds_max: {
+    label: 'SpliceAI',
+    warningThreshold: 0.2,
+    dangerThreshold: 0.5,
+    absoluteValue: false,
+  },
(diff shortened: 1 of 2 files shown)


browser/src/VariantPage/VariantInSilicoPredictors.tsx
```diff
@@ -93,7 +118,9 @@ const VariantInSilicoPredictors = ({ variant, datasetId }: Props) => {
               const predictor = predictors[id]
 
               let color = null
-              const parsedValue = parseFloat(value)
+              const parsedValue = predictor.absoluteValue
```
sjahl, line 121: i think using the absolute value here is probably reasonable here, as long as we know the thresholds are always going to be the same on either side of 0.

i.e., if the need to specify:

`pangolin_largest_ds: { label: 'Pangolin', warningThreshold: 0.2, dangerThreshold: 0.5 },`

and

`pangolin_largest_ds: { label: 'Pangolin', warningThreshold: -0.3, dangerThreshold: 0.6 },`

where the thresholds are different numbers for positive and negative, we might need something different here.

But, disregard if those will always match 👍

sjahl (approved):
Looks good to me. Just the one comment on absolute value inline, but merge away if that's not a problem

browser/src/VariantPage/VariantInSilicoPredictors.tsx
```diff
@@ -19,13 +20,43 @@ const FLAG_DESCRIPTIONS = {
...
+    dangerThreshold: 0,
+    absoluteValue: false,
+  },
+  polyphen_max: {
+    label: 'PolyPhen (max)',
+    warningThreshold: 0.978,
+    dangerThreshold: 0.999,
+    absoluteValue: false,
+  },
+  cadd: { label: 'CADD', warningThreshold: 25.3, dangerThreshold: 28.1, absoluteValue: false },
```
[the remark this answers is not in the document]
rileyhgrant: Yeah this irked me a bit too lol. Oh well.

browser/src/VariantPage/VariantInSilicoPredictors.tsx
rileyhgrant: Yeah good eye, this is the first one that even needed negative value thresholds. For this instance negative denotes a different class of prediction (splice gained vs splice loss) that shares the same thresholds. Without knowing I'd hope its a fair guess that other predictors may follow a similar pattern if they allow negative scores?

We can revisit later down the line if needed.

KoalaQin:
Thanks! I don't see the change yet, it's after the deploy?

PR #1350, opened 2023-12-06
Update paths to variant cooccurrence tsv files

### LINKED ISSUE #1328: Access to `variant_cooccurrence_counts.py`  input files
This is a data question - I did send an email as per the issue template guidance, but I see you are doing active development on `genes` pipeline, so I thought I'd cheekily ask here too. Apologies for cross-posting :grimacing: 

I am working through the `genes` pipeline. Thanks for updating the GTEx and pext paths @rileyhgrant . A subsequent step in the pipeline uses `variant_cooccurrence_counts.py` which references to input data files assigned to `TWO_HET_DATA_PATH` and `HOMOZYGOUS_DATA_PATH`. These files are stored in the `gs://gnomad-browser-data-pipeline` bucket which doesn't have public access. Do you know if these files are able to be shared, or if they are able to be found in one of the other public buckets? I've had a good hunt through the manifests etc, but have not had any luck.

rileyhgrant:
Resolves #1328 

- Minorly updates v2 downloads to use the correct download links
- Updates the variant coocurrence pipeline task to refer to the publicly released `.tsv` files
  - Changes the name of a field in the pipeline task to accomodate the field name change in the public file

I tested running the task locally and it succeeds using the public data. I did not run it in the context of the whole pipeline, fwiw, but the return schema from the tasks are unchanged.

### DIFF
--- a/browser/src/DownloadsPage/GnomadV2Downloads.tsx
+++ b/browser/src/DownloadsPage/GnomadV2Downloads.tsx
@@ -641,14 +641,14 @@ const GnomadV2Downloads = () => {
           </ListItem>
           {/* @ts-expect-error */}
           <ListItem>
-            <GetUrlButtons
+            <GenericDownloadLinks
               label="Variant co-occurrence by gene (homozygous rare variants)"
               path="/release/2.1.1/secondary_analyses/variant_cooccurrence/gnomAD_v2_homozygous_rare_variants_table_for_download.tsv"
             />
           </ListItem>
           {/* @ts-expect-error */}
           <ListItem>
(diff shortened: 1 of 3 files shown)


rileyhgrant:
Requesting your review on this @phildarnowsky-broad as you've been much closer to this feature and pipeline task.

Very minor update of the task itself to reference the publicly released files, and to transmute the changed field name into the one used throughout the rest of the stack (`af_threshold` in the input file, rather than `af_cutoff`)

phildarnowsky-broad (approved):

Issue #1247, opened 2023-11-01
Fix deployctl blog deployment


rileyhgrant:
Deploying a new blog deployment to the v4-cluster when on time crunch revealed that deployctl is broken for the blogs.

The workaround was to use the image from the old blog deployment.

One of the pods was in a crash backoff loop because it did not have `jinja`, need to figure out where the python env gets added and correctly add `jinja`.

- Go through this process when not on the time crunch
- Add more documentation

PR #1367, opened 2024-01-03
Blog deployment fixes

sjahl:
This attempts to resolve some of the blog auth container build issues we saw during the v4 deployment #1247 

I've moved all requirements to the pip requirements file, and locked it with pip-tools. I've also upgraded the container to the latest stable release of python

All the dependencies have been updated, and we've moved from Flask 1 to 3, which introduced some syntax sugar for the `@app` routes, hence the changes there. I'm still working out how best to test this without actually deploying it in prod...


sjahl:
There was not a great way to test this without deploying production. I started up a separate deployment to validate that the startup error was gone, and then pointed the blog service at it to test, and was able to log in fine. Currently running in prod, since it's low risk -- we can redeploy with a clean docker tag after merge.

sjahl:
Closes #1247

mattsolo1 (approved):

PR #1370, opened 2024-01-05
bug: actually use the provided redis port in single host configs

sjahl:
realized when looking at something else that the single-host branch of the redis config logic wasn't passing the REDIS_PORT environment variable through and using it.

### DIFF
--- a/graphql-api/src/cache.ts
+++ b/graphql-api/src/cache.ts
@@ -7,6 +7,12 @@ import logger from './logger'
 import { JsonCache } from './queries/helpers/json-cache'
 import largeGenes from './queries/helpers/large-genes'
 
+type WithCacheOptions = {
+  expiration?: number
+  jsonCacheEnableAll?: boolean
+  jsonCacheLargeGenes?: boolean
+}
+
 let fetchCacheValue = () => Promise.resolve(null)
 let setCacheValue = () => Promise.resolve()
 let setCacheExpiration = () => Promise.resolve()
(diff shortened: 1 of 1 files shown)


sjahl:
it looks like prettier added some additional formatting in here when i saved... the non REDIS_PORT lines aren't really mine, but we should probably commit them if the formatter we all use is suggesting the changes.

graphql-api/src/cache.ts
```diff
@@ -70,7 +72,11 @@ if (config.REDIS_HOST) {
 export const withCache = (fn: any, keyFn: any, options = {}) => {
   // @ts-expect-error TS(2339) FIXME: Property 'expiration' does not exist on type '{}'.
   //
-  const { expiration = 3600, jsonCacheEnableAll = config.JSON_CACHE_ENABLE_ALL, jsonCacheLargeGenes = config.JSON_CACHE_LARGE_GENES } = options
+  const {
```
phildarnowsky-broad, line 75: @sjahl Probably the best way to resolve the type conflicts would be to introduce a type for the options of `withCache`, something like:

```
type WithCacheOptions = {
  expiration?: number
  jsonCacheEnableAll?: string     // or whatever JSON_CACHE_ENABLE_ALL is
  jsonCacheLargeGenes?: string
}
```

And then the start of `withCache` becomes `export const withCache = (fn: any, keyFn: any, options: WithCacheOptions = {}) => { ...`

If you're not pressed for time, this would also be a good juncture to give `fn` and `keyFn` proper types as well, and maybe a name for `fn` that conveys some useful information beyond the fact that it's a function.

sjahl:
Thanks @phildarnowsky-broad -- I think this is ready for another review pass. I've introduced three types, one for the withCache opts, and types for `fn` and keyFn`. The latter two are still quite broad, because it's not clear to me that we really know exactly what is getting passed in here under every circumstance, but it's at the very least more specific than the original `any` types were.

phildarnowsky-broad (approved):

Issue #1154, opened 2023-08-10
Find and remove copy/pasted code


phildarnowsky-broad:
There is a significant amount of code in this codebase written by copy/paste. For many commonly known reasons, this is a very bad practice. Look for opportunities to improve the code by reducing duplication.

phildarnowsky-broad:
Some examples: `variantTableColumns.tsx`, `structuralVariantTableColumns.tsx`, and `structuralVariantTableColumns.tsx` all have multiple functions that are the same from one to the next.

phildarnowsky-broad:
Found a tool to detect copypasta that seems decent--at least, it's fast and I spot-checked some results and verified that they're correct. Here's what the output looks like:

`
~/src/gnomad-browser/browser $ npx jscpd --pattern "src/**/*.ts(x)?"
Clone found (tsx):
 - src/VariantPage/VariantPage.spec.tsx [66:17 - 81:9] (15 lines, 127 tokens)
   src/VariantPage/VariantPage.spec.tsx [49:17 - 64:20]

Clone found (tsx):
 - src/VariantPage/VariantPage.spec.tsx [81:2 - 92:2] (11 lines, 117 tokens)
   src/VariantPage/VariantPage.spec.tsx [64:10 - 58:2]

Clone found (tsx):
 - src/VariantPage/VariantLiftover.tsx [111:22 - 118:22] (7 lines, 77 tokens)
   src/VariantPage/VariantLiftover.tsx [59:22 - 66:22]

Clone found (tsx):
 - src/VariantPage/VariantInSilicoPredictors.tsx [92:3 - 102:2] (10 lines, 115 tokens)
   src/VariantPage/VariantInSilicoPredictors.tsx [78:6 - 89:7]

Clone found (javascript):
 - src/VariantPage/VariantGenotypeQualityMetrics.tsx [178:7 - 189:20] (11 lines, 83 tokens)
   src/VariantPage/VariantGenotypeQualityMetrics.tsx [99:7 - 110:31]

Clone found (javascript):
 - src/VariantPage/VariantGenotypeQualityMetrics.tsx [189:20 - 296:15] (107 lines, 786 tokens)
   src/VariantPage/VariantGenotypeQualityMetrics.tsx [110:31 - 185:2]

Clone found (tsx):
 - src/VariantPage/VariantGenotypeQualityMetrics.tsx [199:13 - 208:17] (9 lines, 107 tokens)
[...]

PR #1384, opened 2024-01-17
feat(styles): make more dry

ignatiusm:
Another PR addressing an item from the list of copy/paste sections on issue #1154. Moves some styled-components that were defined in two places, in to a separate file. Styled components are then imported where needed.

This one did require updating snapshots due to the `className` changes.

Passes tests locally.

FYI @rileyhgrant

### DIFF
--- a/browser/src/ChartStyles.ts
+++ b/browser/src/ChartStyles.ts
@@ -1,0 +1,56 @@
+import styled from 'styled-components'
+
+export const ControlPanel = styled.div`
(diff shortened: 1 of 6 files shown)


rileyhgrant (approved):
Looks good to me, thanks for the contribution as always :)

Testing are passing as you mentioned, and for the sake of it I also checked out this branch and did a once over to visually confirm everything is as expected -- looks great.

PR #1389, opened 2024-01-17
Update frontend and graphql-api CI to run on dependency updates

rileyhgrant:
Updates the frontend and graphql-api github CI yml files to also trigger a run when either the 'package.json' or 'yarn.lock' files are edited.

The primary purpose is to have CI run when dependabot opens PRs that update javascript dependencies, however I think it's also probably good in general to run the test suite whenever any PR wants to update top level javascript dependencies.

### DIFF
--- a/.github/workflows/browser-ci.yml
+++ b/.github/workflows/browser-ci.yml
@@ -5,9 +5,13 @@ on:
       - main
     paths:
       - 'browser/**'
+      - 'package.json'
+      - 'yarn.lock'
   pull_request:
     paths:
       - 'browser/**'
+      - 'package.json'
+      - 'yarn.lock'
 jobs:
   checks:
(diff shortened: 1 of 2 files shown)


sjahl (approved):
Makes sense to me! What are those top level lock/package files actually used for? Are these dependencies that are shared between the browser and API?

rileyhgrant:
> Makes sense to me! What are those top level lock/package files actually used for? Are these dependencies that are shared between the browser and API?

Yeah mostly, also there's a lot of dev dependencies that are tied up in the top level `package.json`, `prettier` `jest`, `eslint` etc. So dev dependencies that are shared between the frontend and the api seems to be a big use case of it. That seems to make sense to me, but not sure how other places do it.

Issue #1393, opened 2024-01-22
v2 gene page CSV export displaying incorrect column info


rileyhgrant:
Per this [Slack Thread](https://atgu.slack.com/files/USLACKBOT/F06EG6VH18X/csv_download_issue_for_adcy10_and_agxt), the v2 downloads appear to be showing the 'afr' set of columns twice, and skipping some other columns.

I am the one who most recently touched this logic, I can likely take a look at this some time this week since it's meeting free week.

Related #1055

gm-bee:
Hi, just wondering if there's an estimate of when this issue would be fixed? We rely on these downloads for an internal process and can't move forward w/o it.

rileyhgrant:
Hiya @gm-bee 

Sorry to have blocked you for this amount of time. I'm planning on merging and deploying the PR this morning. 

For v2 this fixes the `afr` group showing up twice incorrectly, replacing the duplicate `afr` set of columns with the `amr` (admixed american) columns as intended. It also adds a few new columns to the v2 download: gnomad ID then 4 columns about groupmax AF and genetic ancestry group for exomes and genomes in v2.

I'll be merging and deploying those changes this morning, definitely reach out if things are still unsatisfactory, and thanks for your patience.

- Riley

rileyhgrant:
This has been merged and deployed, closing this issue as resolved. Please reach out if you see any further issues.

michellepaczosa:
Thank you for your work resolving this!

Issue #1055, opened 2022-12-07
Additional data to optionally show in gene page variant table (and download)


anneodonnell:

mattsolo1:
What is the additional data?

anneodonnell:
My initial ticket text did not go through so trying this again. 

There are some additional fields available on the variant page that would be helpful to include as optional fields (in the Configure table menu). 

For gnomAD v2, popmax AF, popmax exome FAF, popmax genome FAF. 

For gnomAD v3 (and v4 when it is added), popmax AF, popmax genome FAF (which will be a combined FAF in gnomAD v4). Also, the in silico predictors (REVEL, CADD, SpliceAI, and PrimateAI). Genomic constraint score might also be nice to have here. 

Users should be able to select which variants to include in the table and when they select “export variants to CSV” it should include these fields, if selected/shown. People ask me fairly frequently to be able to download these data across a gene from the gene variant table on gnomAD.

PR #1397, opened 2024-01-23
Add groupmax information to gene page variant CSV export

rileyhgrant:
Adds a few default fields to the gene page CSV export
- gnomad ID
- groupmax AF, groupmax group


nadeaujoshua (approved):

rileyhgrant:
Stalling on merging of this per another issue (#1393). 

As discussed in browser meeting, #1393 combined with #1055 may warrant creating a mapping of gnomAD version to default CSV export tables instead of trying to cleverly create a one size fits all solution.

PR #1398, opened 2024-01-25
Add pnpm version to volta config

mattsolo1:
Pins pnpm version to the [volta](https://volta.sh/) tool manager config.


phildarnowsky-broad:
@mattsolo1 FYI per discussion in another PR I'm checking out pnpm 8.14.3, so far so good, so if no regressions there may as well jump to 8.14.3 in this PR as well.

phildarnowsky-broad (approved):
pnpm 8.14.3 seems unproblematic so let's jump to that here unless you see a good reason not to, otherwise good to go

PR #1404, opened 2024-01-31
Correct one command in deploy docs

phildarnowsky-broad:


sjahl (approved):
Good catch, thanks @phildarnowsky-broad !

PR #1411, opened 2024-02-07
Fix regressions introduced by #1397

rileyhgrant:
Fixes a few bugs introduced by #1397:
- v4 transcript pages not loading variants
- v4 region pages not loading variants
- v2 gene, region, transcript pages not exporting CSV table when there are samples present in only exomes or only genomes


sjahl (approved):
Changes here look fine to me, and I verified that the broken cases I reported this afternoon are fixed in the demo with this branch! 👍

PR #1419, opened 2024-02-14
Update generate caids pipeline: add gnomAD v4 and update Hail usage

nadeaujoshua:
This PR includes 2 updates to the data pipeline that generates a Hail table for `caids`:

1. Add gnomAD v4 to the pipeline
2. Update Hail imports and their usage to be compatible with newer versions of Hail
    * The `get_caids` module uses a number of Hail utils that have either been changed, removed or replaced since its last update

### DIFF
--- a/data-pipeline/caids/get_caids.py
+++ b/data-pipeline/caids/get_caids.py
@@ -11,10 +11,12 @@ from concurrent.futures import ThreadPoolExecutor
 from typing import Awaitable, Callable, TypeVar
 
 import aiohttp
-from hailtop.aiotools import LocalAsyncFS, RouterAsyncFS
-from hailtop.aiogoogle import GoogleStorageAsyncFS
-from hailtop.utils import bounded_gather, sleep_and_backoff, tqdm
+from hailtop.aiotools.router_fs import RouterAsyncFS
+from hailtop.utils import bounded_gather, sleep_before_try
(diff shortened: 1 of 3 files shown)


data-pipeline/caids/get_caids.py
```diff
@@ -158,7 +160,9 @@ async def get_caids(sharded_vcf_url: str, output_url: str, *, parallelism: int =
                 if part_name not in completed_parts:
                     remaining_part_urls.append(part_url)
 
-            with tqdm(total=len(remaining_part_urls)) as progress:
+            logger.warning(f'\n\nParts Counts\nTotal: {len(all_part_urls)}\nCompleted: {len(completed_parts)}\nRemaining: {len(remaining_part_urls)}\n')
+
+            with SimpleCopyToolProgressBar(total=len(remaining_part_urls)) as progress:
```
mattsolo1, line 165: Neat I didn't know this progress bar was a thing.

mattsolo1 (approved):
Looks great to me! (without actually running it)

Next step is to add this to the output to the gnomAD v4 variant pipeline.

https://github.com/broadinstitute/gnomad-browser/blob/5c26e5d697eb874bdc4137fb74326e6699b2fb26/data-pipeline/src/data_pipeline/pipelines/gnomad_v4_variants.py#L65

This also means reloading all variant data in Elasticsearch which is a whole thing unto itself...

Issue #1412, opened 2024-02-08
Improve typing on CSV export


rileyhgrant:
The CSV export can utilize static typing better to avoid unintended regressions. This may also be applicable for the v4-variant-queries.ts file

PR #1420, opened 2024-02-14
Improve typing on variant table CSV export

rileyhgrant:
Resolves: #1412 

Improves typing on the CSV export functionality on variant tables. Per discussion with Phil, the component now utilizes several different types to characterize the differing possible fields of variants from different datasets.

For now, there is no interface that acts as an abstract base type, as the minimal
`VariantTableVariant` type serves as a concrete type for a generic variant table variant (exac and v3), in addition to two more specialized types introduced for v2 and v4 variant table variants. I'm happy to change this if we think having an interface act as the base type is a better pattern.

### DIFF
--- a/browser/src/VariantList/ExportVariantsButton.tsx
+++ b/browser/src/VariantList/ExportVariantsButton.tsx
@@ -348,12 +348,12 @@ export type VariantTableVariant = {
   transcript_id: string
   transcript_version: string
   variant_id: string
-  exome?: {
+  exome: {
     filters: string[]
-  }
-  genome?: {
+  } | null
+  genome: {
     filters: string[]
-  }
+  } | null
 }
 
 type FilteredAlleleFrequency = {
@@ -362,10 +362,10 @@ type FilteredAlleleFrequency = {
 }
 
 type V2VariantTableVariant = VariantTableVariant & {
-  exome?: {
+  exome: {
     faf95: FilteredAlleleFrequency
   }
-  genome?: {
+  genome: {
     faf95: FilteredAlleleFrequency
   }
 }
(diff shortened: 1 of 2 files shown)


browser/src/VariantList/ExportVariantsButton.tsx
```diff
@@ -335,8 +356,31 @@ export type VariantTableVariant = {
...
+type FilteredAlleleFrequency = {
+  popmax: number | null
+  popmax_population: string | null
+}
+
+type V2VariantTableVariant = VariantTableVariant & {
+  exome?: {
+    faf95: FilteredAlleleFrequency
+  }
+  genome?: {
```
phildarnowsky-broad, line 368: Is this coming direct from GraphQL? If so should be `| null` rather than `?`.

phildarnowsky-broad (approved):
One thing to double-check, good to go once it's fixed, assuming it does need it.

Speaking to the question of concrete base class vs. interface as quasi-abstract-class, I think that all else equal there's nothing wrong with having a concrete base class, it's appropriate in a case like this where the schema and logic for all the different types of variants is almost all identical. A case where an interface/abstract base class might be better would be something like a cache where the backing store might be in memory, might be files in a filesystem, rows in a database, or something more exotic. In that case you would use your interface to define the API that each of those classes had to hew to, but the implementation of each would be so different that there's not any chance for code reuse to speak of.

browser/src/VariantList/ExportVariantsButton.tsx
rileyhgrant: Aha, yep right you are, it is indeed coming directly from GraphQL. Added a fixup commit that changes that.

PR #1434, opened 2024-03-08
Clean up a bunch of ts-expect-error and type problems in short tandem…

phildarnowsky-broad:
… reads page

I had a couple of uncommitted hours and chipped away a little bit of tech debt.


rileyhgrant (approved):
Nice once, lgtm!

Issue #1409, opened 2024-02-07
Add mitochondrial gene and regional constraint to browser


ch-kr:
As discussed in the [project meeting](https://docs.google.com/document/d/109NuIQlZ3a8uw__iBMK7FpoO2bFFGECzEDnBpB1xW8E/edit?usp=sharing) yesterday, Nicole Lake has requested that we add her newly calculated mitochondrial constraint results to the browser.

Ideally, we would add her mito gene constraint to the mito gene pages as a table here:
<img width="403" alt="image" src="https://github.com/broadinstitute/gnomad-browser/assets/19142091/37fa7861-3f7b-4178-8b67-0281a3932e89">
using a similar format as the nuclear gene constraint, and mito regional constraint also to the gene pages above the ClinVar and gnomAD variant tracks:
<img width="1443" alt="image" src="https://github.com/broadinstitute/gnomad-browser/assets/19142091/9526f966-5b29-46d7-b3d8-01a3826c990c">
using a similar display as nuclear regional missense constraint.

rileyhgrant:
Related: #1339

phildarnowsky-broad:
Pinged Nicole today to remind her about the two blurbs of help text we're waiting on.

ch-kr:
linking this ticket: https://github.com/broadinstitute/gnomad-browser/issues/1666. #1666 isn't urgent, but it would be great to add this fix when releasing mtDNA constraint metrics

ch-kr:
I forgot to include this in the initial ticket, but would it be possible to create a demo browser with her results to crowd-source feedback?

PR #1451, opened 2024-03-20
Constraint for mitochondrial genes

phildarnowsky-broad:
This adds a table and a region plot of constraint for mitochondrial genes. One unusual thing I've done here is to keep this data in memory on the API server, rather than create a data pipeline. There's so little data--just over 100 rows--that a pipeline felt like massive overkill.

Here's a typical track. Note that, per discussion with Nicole, there aren't really different degrees of constrained regions, it's more a binary choice of "is constrained" or "isn't constrained". Hence just the one kind of block in the track.

Screenshot 2024-05-01 at 3 43 23 PM


phildarnowsky-broad:
resolves #1409

phildarnowsky-broad:
Building a demo did fail for me so there will be at least one more fixup to come

phildarnowsky-broad:
Note to self: need to get copy for both info buttons from Nicole

browser/src/RegionalMissenseConstraintTrack.tsx
```diff
@@ -216,7 +214,7 @@ const RegionTooltip = ({ region, isTranscriptWide }: RegionTooltipProps) => {
     <RegionAttributeList>
       <div>
         <dt>Coordinates:</dt>
-        <dd>{`${region.chrom}:${region.region_start}-${region.region_stop}`}</dd>
+        <dd>{`${region.chrom}:${region.start}-${region.stop}`}</dd>
```
rileyhgrant, line 140: `region_start` and `region_stop` while in hindsight, probably poorly named, are not redundant. Also in hindsight this may have been a good place for a comment in the code discussing this, since it was a specific request and a bit subtle.

The track renders and colors each exon or portion of an exon based on the `start` and `stop` coordinates output from finding the overlap of the two passed in region arrays, but the RMC team wanted the hover over text to reflect the coordinates of the constrained region (`region_start` and `region_stop`), not the coordinates of the exon or portion of the exon (`start` and `stop`). 

The RMC team had reached out to us to change the behavior to be this way, and at that time passing the `region_start` and `region_stop` coordinates so that each output overlapping region would have those values was my solution.

---
Screenshots

In production, since these two exons are part of the same constrained region, the tooltips show the start and stop coordinates for the region:
- Screenshot 2024-05-02 at 09 18 04
-Screenshot 2024-05-02 at 09 18 08

With this change, this behavior has reverted to displaying the coordinates for the block thats rendered:
- Screenshot 2024-05-02 at 09 17 22
- Screenshot 2024-05-02 at 09 17 25

browser/src/ConstraintTrack.tsx
```diff
@@ -52,6 +53,36 @@ type Props<R extends GenericRegion> = {
   valueFn: (region: R) => string
 }
 
+export const regionsInExons = <R extends GenericRegion>(regions: R[], exons: Exon[]): R[] => {
+  const sortedRegions = regions.sort((a, b) => a.start - b.start)
+  const sortedExons = exons.sort((a, b) => a.start - b.start)
+
```
rileyhgrant, line 86: Woohoo, very nicely done.

I honestly avoided touching that function after extending the tests for it since it was a bit unreadable to my eyes at that time, this goes a long way towards making it approachable -- kudos.

browser/src/GenePage/MitochondrialRegionConstraintTrack.tsx
```diff
@@ -0,0 +1,74 @@
...
+type TooltipProps = {
+  region: MitochondrialConstraintRegion
+}
+
+const Tooltip = ({ region }: TooltipProps) => {
+  return (
+    <RegionAttributeList>
+      <div>
+        <dt>Coordinates:</dt>
+        <dd>{`m:${region.start}-${region.stop}`}</dd>
```
rileyhgrant, line 38: This is very minor -- it appears as though regions and variants use a capital M to denote a region in mitochondrial DNA, rather than a lowercase.

https://gnomad.broadinstitute.org/gene/ENSG00000198899?dataset=gnomad_r4

browser/src/ConstraintTable/MitochondrialConstraintTable.tsx
```diff
@@ -0,0 +1,136 @@
...
+  oe: number
+  oeLower: number
+  oeUpper: number
+}) => (
+  <tr>
+    <th scope="row">{category}</th>
+    <td>{expected.toFixed(1)}</td>
+    <td>{observed.toFixed(1)}</td>
+    <td>
+      {oe.toFixed(2)} ({oeLower.toFixed(2)}-{oeUpper.toFixed(2)})
```
rileyhgrant, line 34: I wonder if we could label this with `o/e = ...` to be more in line with the non-mito gene constraint table, or since there's only one 'contraint metric' in this cell, just label the column 'Observed/Expected'. It took me a few seconds to understand what that cell was showing without the context of some sort of label.

There are also a few other bells and whistles (that visual indication of o/e) that could be added to be more in line, but this may be getting into more of a question for Matt or KC type territory.

Mito gene constraint table
- Screenshot 2024-05-02 at 09 51 57


Non-mito gene constraint table
- Screenshot 2024-05-02 at 09 51 52

rileyhgrant (changes requested):
Nicely done on the whole, one main comment with requested changes about a regression in what the RMC tooltip now displays, and a few minor thoughts. Apart from that lots of great things in here.

The refactoring of the merge regions function in particular is great to see an example of that, it's way more comprehensible now.

Also the work towards unifying the constraint tracks is excellent.

browser/src/ConstraintTrack.tsx
phildarnowsky-broad: refactoring this bit was very satisfying

phildarnowsky-broad:
@rileyhgrant this latest push should address all your feedback. I came up with a solution to the `region_start`/`stop` (now called `unclamped_start`/`stop`) conundrum that I like pretty well, and attempted to make it clearer what that was about.

rileyhgrant (approved):
Nice, lgtm!

I like the solution on how to neatly keep `unclamped` start and stop around for rendering tooltips as needed.

phildarnowsky-broad:
@rileyhgrant this should address all your feedback, just needed a bit of light post-rebase cleanup. Ready for another look.

phildarnowsky-broad:
@rileyhgrant please disregard earlier comment, meant to post that on a different PR

phildarnowsky-broad:
@rileyhgrant FYI two small followups to come here in the next hour or so

PR #1452, opened 2024-03-20
Fix broken snapshot

phildarnowsky-broad:
PR #1435 somehow passed CI despite the fact that its snapshots needed updating, which is disquieting.


phildarnowsky-broad:
Overlooked that this was approved, but #1454 makes the same fix while also addressing the underlying problem, so closing this in favor of that.

PR #1436, opened 2024-03-08
Update metadata to render genomic constraint for v4

rileyhgrant:
Resolves #1421 

Updates the metadata so that non-coding constraint renders on v4 region and variant pages.

---

Screenshot 2024-03-08 at 16 11 39

PR #1454, opened 2024-03-25
Update region page snapshot with v4 genomic constraint

rileyhgrant:
Related #1436 

Updates the snapshot for the region page, as genomic constraint now displays on v4 region pages.

Also updates the CI so that typescript related CI (frontend and API) run when the metadata is edited, this snapshot was mistakenly missed initially as the CI didn't run since no frontend files changed in #1436.


phildarnowsky-broad (approved):
Nice catch

PR #1473, opened 2024-04-03
Update genes and short variants for v4.1

### LINKED ISSUE #1423: Add new flag to variants page when variant is not called in one data type
Very long context:
Variants that are only present in one data type (exomes or genomes) currently only display frequencies in a single data type. For example:
image
https://gnomad.broadinstitute.org/variant/2-151663543-C-T?dataset=gnomad_r4
only shows frequency information from the v4 exomes because this variant was not called in the v4 genomes. Given the genomes coverage in this region:
image
the most likely reason why this variant doesn't have any allele number (AN) information from the genomes is due to the fact that we do not joint call our two data types into a single callset (we have separate exomes and genomes callsets). This means that the v4 genomes likely all had homozygous reference genotype calls at this locus, so this variant was not emitted in that callset (despite it being present in 7 exome samples). 

The team is working on a new type of computation that generates AN across all possible sites for both data types. Once we have created this new releasable file, we will be able to incorporate new AN information to variants that previously only had frequency information from one data type but have good coverage in the other data type (such as 2-151663543-C-T above).

Request:
After we update the ANs / joint frequencies, I would like to update this "No variant" to make it very clear when a variant was truly not seen / not possible to have been detected / zero AN in either the exomes or genomes. My current idea is to add a hover over to the "No variant" box explaining this, something like:

> This variant was not called in the gnomAD <data type>; no samples had any genotype call (no samples had reference or alternate calls).

Would it be possible to add this new hover over text and share a screenshot/demo with the wider project to get feedback?

### LINKED ISSUE #1442: Add new flag for variants with discrepant frequencies
Julia has been hard at work generating combined frequencies across the exomes and genomes/running statistical tests on these freqs. We would like to add a flag to variants with very discrepant frequencies between the two data types (as flagged by the Cochran–Mantel–Haenszel test) as part of the v4.1 release. The flag likely should have a hover-over that includes the actual statistic and p-value.

Tagging @jkgoodrich in case I've mixed up the messaging and/or to provide more details

### LINKED ISSUE #1475: v4.1 demo requests
Ticket tracking requests/feedback associated with v4.1 demo browser

- [x] Add the test name (Cochran–Mantel–Haenszel) to the hover over text in the discrepant frequencies flag
- [x] Shift the p-value in the hover over text (same flag as above) to scientific notation and drop some of the decimal points. For example, for [this variant](http://34.36.169.116/variant/1-55039736-C-T?dataset=gnomad_r4), we should display something like 9.2x10^-6 or 9.20x10^-6
- [x] Add a question mark next to "Filters" that links to the [flags](http://34.36.169.116/help/what-do-the-flags-on-the-browser-mean) help page (like the question mark next to "Grpmax filtering AF") and methods team to update text on flags help page to add new discrepant freqs flag and link to new joint freqs stats help page

### LINKED ISSUE #1491: Update hover over text for AC0
Current text: 
image

Proposed text:
"Allele count is zero due to removal of low quality genotypes (GQ < 20; DP < 10; and AB < 0.2 for het calls) at this site."

rileyhgrant:
Resolves #1423, 
Resolves #1442, 
Resolves #1475,
Resolves #1491,

Updates:
- **pipelines**
  - genes pipeline now uses v4.1 constraint data (minor update to one task to not drop a field)
  - v4 variants pipeline annotates variants with `joint` frequency data from the new release hailtable
- **graphql-api**
  - add typing for new `joint` frequency data being returned
  - minor refactors to DRY error prone code
- **frontend** 
  - query and use new `joint` frequency data
  - minor refactors to DRY error prone code

### DIFF
--- a/browser/src/VariantPage/VariantOccurrenceTable.tsx
+++ b/browser/src/VariantPage/VariantOccurrenceTable.tsx
@@ -43,7 +43,26 @@ const NoWrap = styled.span`
 type VariantContext = 'exome' | 'genome' | 'joint'
 const renderGnomadVariantFlag = (variant: Variant, context: VariantContext) => {
   if (!variant[context]) {
-    return context === 'joint' ? <div /> : <Badge level="error">No variant</Badge>
+    let badgeName = 'No variant'
+    let badgeDescription
+
+    if (variant.joint) {
+      if (
+        (context === 'exome' && variant.joint.filters.includes('not_called_in_exomes')) ||
+        (context === 'genome' && variant.joint.filters.includes('not_called_in_genomes'))
+      ) {
+        badgeName = 'No data'
+        badgeDescription = `This variant was not called in the gnomAD ${context} callset; no ${context} samples had any genotype call (no reference or alternate calls)`
+      }
+    }
+
+    return context === 'joint' ? (
+      <div />
+    ) : (
+      <Badge level="error" tooltip={badgeDescription}>
+        {badgeName}
+      </Badge>
+    )
   }
   const { filters } = variant[context]!
 
@@ -169,6 +188,11 @@ export const GnomadVariantOccurrenceTable = ({
   const isPresentInGenome = Boolean(variant.genome)
   const hasJointFrequencyData = Boolean(variant.joint)
 
+  const notCalledInExomes =
+    hasJointFrequencyData && variant.joint!.filters.includes('not_called_in_exomes')
+  const notCalledInGenomes =
+    hasJointFrequencyData && variant.joint!.filters.includes('not_called_in_genomes')
+
   const exomeAlleleCount = isPresentInExome ? variant.exome!.ac : 0
   const exomeAlleleNumber = isPresentInExome ? variant.exome!.an : 0
   const genomeAlleleCount = isPresentInGenome ? variant.genome!.ac : 0
@@ -291,8 +315,20 @@ export const GnomadVariantOccurrenceTable = ({
                 <TooltipHint>Allele Count</TooltipHint>
               </TooltipAnchor>
             </th>
-            {showExomes && <td>{isPresentInExome && exomeAlleleCount}</td>}
-            {showGenomes && <td>{isPresentInGenome && genomeAlleleCount}</td>}
+            {showExomes && (
+              <td>
+                {isPresentInExome && exomeAlleleCount}
+                {notCalledInExomes && '-'}
+                {hasJointFrequencyData && !isPresentInExome && !notCalledInExomes && 0}
+              </td>
+            )}
+            {showGenomes && (
+              <td>
+                {isPresentInGenome && genomeAlleleCount}
+                {notCalledInGenomes && '-'}
+                {hasJointFrequencyData && !isPresentInGenome && !notCalledInGenomes && 0}
+              </td>
+            )}
             {showTotal && <td>{totalAlleleCount}</td>}
           </tr>
           <tr>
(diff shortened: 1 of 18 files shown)


graphql-api/src/queries/variant-datasets/gnomad-v4-variant-queries.ts
```diff
@@ -252,9 +252,11 @@ const shapeVariantSummary = (subset: any, context: any) => {
 
     const exomeFilters = variant.exome.filters || []
     const genomeFilters = variant.genome.filters || []
+    const jointFilters = variant.joint.filter || []
 
     const hasExomeVariant = variant.exome.freq[subset].ac_raw
     const hasGenomeVariant = variant.genome.freq[subset].ac_raw
+    const hasJointVariant = variant.joint.freq[subset].ac_raw
```
phildarnowsky-broad, line 263: One thing I got caught on a couple times while working on the non-UKB dataset is code similar to this pattern. Non-UKB doesn't have genomes so `variant.genome.freq[subset]` would be `undefined` for non-UKB, and then you'd get a crash when trying to dereference `ac_raw` on it.

Is that going to be a problem with the joints?

graphql-api/src/queries/variant-datasets/gnomad-v4-variant-queries.ts
```diff
@@ -264,7 +266,11 @@ const shapeVariantSummary = (subset: any, context: any) => {
       genomeFilters.push('AC0')
     }
 
-    const hasJointFafData = variant.faf95_joint && variant.faf99_joint
+    if (variant.exome.freq[subset].ac === 0 && !jointFilters.includes('AC0')) {
+      jointFilters.push('AC0')
+    }
+
+    // const hasJointFafData = variant.faf95_joint && variant.faf99_joint
```
phildarnowsky-broad, line 277: can get deleted

graphql-api/src/queries/variant-datasets/gnomad-v4-variant-queries.ts
```diff
@@ -343,6 +343,31 @@ const shapeVariantSummary = (subset: any, context: any) => {
   }
 }
 
+const getMultiVariantSourceFields = (
```
phildarnowsky-broad, line 351: nice refactor here

browser/src/VariantPage/VariantOccurrenceTable.tsx
```diff
@@ -43,15 +43,23 @@ const NoWrap = styled.span`
...
   const { filters } = variant[context]!
 
   if (filters.length === 0) {
-    return <Badge level="success">Pass</Badge>
+    return context === 'joint' ? <div /> : <Badge level="success">Pass</Badge>
   }
 
-  return filters.map((filter: any) => <QCFilter key={filter} filter={filter} />)
+  return filters.map((filter) => {
+    let data = {}
```
phildarnowsky-broad, line 55: as a rule we should try and avoid `let` if it can be avoided without making the code too convoluted. looks like you could rewrite this neatly like so:

```
return filters.map((filter) => {
  const data = filter === 'discrepant_frequencies' ? {cochran_p_value: blah blah} : {}
  return <QCFilter key={filter} filter={filter} data={data} />
})
```
```

phildarnowsky-broad (changes requested):
There's one main thing I'd like you to check and a couple other bits and pieces. Overall looking very solid though.

graphql-api/src/queries/variant-datasets/gnomad-v4-variant-queries.ts
rileyhgrant: Yep it is, thanks for catching this in review before our two branches met in production and joint's lack of a non-ukb subset caused problems

browser/src/VariantPage/VariantOccurrenceTable.tsx
rileyhgrant: Aha, assigning a variable with a ternary here is pretty neat!

rileyhgrant:
Heya @phildarnowsky-broad ,

Thanks for the review. I've addressed your comments, and as such I am formally re-requesting a review. 

Thanks!

phildarnowsky-broad (approved):

PR #1478, opened 2024-04-09
Remove graphql-api checks from browser CI

rileyhgrant:
Currently the browser CI yaml file runs:
- eslint
- tsc
- jest

For all of the relevant ts/js files, so for both the graphql-api and the frontend. This PR splits this, so that each CI job is responsible only for checking its respective directory.


.github/workflows/graphql-api-ci.yml
```diff
@@ -35,6 +36,6 @@ jobs:
       - name: Run ESLint
         run: pnpm eslint graphql-api
       - name: Type check
-        run: pnpm ts-node ./browser/build/buildHelp.ts && pnpm typecheck
+        run: pnpm ts-node ./browser/build/buildHelp.ts && cd graphql-api && pnpm tsc --noEmit --project ../tsconfig.json
```
phildarnowsky-broad, line 39: The typecheck actually should run on both codebases if either are changed, since IIRC there are some types defined in one and used in the other, so we want to make sure those remain consistent.

phildarnowsky-broad (changes requested):
Needs one tweak, otherwise looking good

.github/workflows/graphql-api-ci.yml
rileyhgrant: Aha, ok, thanks for the catch!

PR #1479, opened 2024-04-09
Add e2e test with Playwright

rileyhgrant:
Adds files to accomodate a demo of testing with Playwright to work towards getting initial minimal e2e tests to use in helping QC the gnomAD browser

### DIFF
--- a/tests/e2e/GenePage.playwright.spec.ts
+++ b/tests/e2e/GenePage.playwright.spec.ts
@@ -1,0 +1,48 @@
+import { test, expect, type Page } from '@playwright/test'
+
+let page: Page
+
+test.afterAll(async () => {
+  await page.close()
+})
+
+test.describe('Gene page', () => {
+  test.describe('v4', () => {
+    // Make one API call on the gene page and share that result with other tests
+    test.beforeAll(async ({ browser }) => {
+      page = await browser.newPage()
+      await page.goto('/')
+      await page.getByRole('link', { name: 'PCSK9' }).click()
+    })
+
+    test('gene query', async () => {
+      await expect(page.getByText('Loading gene')).toHaveCount(0)
+
+      await expect(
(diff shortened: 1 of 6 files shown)


phildarnowsky-broad:
@rileyhgrant I've been trying these out on my own machine and found them to be very flaky due to timeouts. Has that also been your experience?

rileyhgrant:
Hmm, when running against the right API, I've not had them flaky due to timeouts unless I run many times concurrently.

I was having difficulty getting these to give a meaningful message if the query fails, when it does it appears to fail due to a timeout -- because the components that should load if the query succeeds don't load within the given timeout.

This branch is currently rebased onto a branch that has API changes, how are you running the tests and which tests are timing out?

phildarnowsky-broad:
> If running based off this branch, without the env var, it would be running with the local frontend changes that hit the production API, in which case we would expect several of the tests to be failing (prod API doesn't yet have joint data available to query)

Yes I saw that, got caught up by that at first but even using the updated API I'm often getting legitimate timeouts. I'm running the tests different ways; sometimes I've been running them with `GNOMAD_API_URL=http://localhost:8010/api pnpm test:playwright`, and sometimes using the UI mode as described here: https://playwright.dev/docs/running-tests#run-tests-in-ui-mode, also using a local API.

phildarnowsky-broad:
Actually it looks like these are more stable right now @rileyhgrant so I'm guessing I was getting rate-limited earlier and/or there was a lot of production traffic. I think I'm OK with the current state of it so just going to give this one more look over and get it back to you.

phildarnowsky-broad (approved):
🎉

rileyhgrant:
Ah ok I misunderstood, yeah, rate limiting and/or heavy production traffic is certainly a complicating factor.

We'll continue to improve these to make them better suit our needs (being able to pass in which dataset to check could be nice).

PR #1483, opened 2024-04-11
Remove returned error when variant has no ClinVar entry

rileyhgrant:
Our API prevents nginx caching when an error is returned, since variants can, and often do, legitimately do not have a ClinVar entry for the given variant id, this removes the error from the response to allow caching of many more variants.


phildarnowsky-broad (approved):
LGTM

PR #1486, opened 2024-04-12
Increase redis memory request/limit to prevent OOMKills when we use the entirety of redis 'maxmemory'

sjahl:
I think we're hitting an edge case where redis is getting OOMKilled when it allocates exactly 36GiB of ram to cache... or exceeds whatever the precision value for '36Gi' is in kubernete's mind.

This bumps the pod request/limit to 39GiB -- my understanding is that. We have around ~6GiB allocatable on the redis GKE node right now, so 39 should keep us under that limit without other infra changes.


rileyhgrant (approved):
lgtm!

PR #1489, opened 2024-04-16
fix(frontend): edit combined freq help text

### LINKED ISSUE #1482: Small fixes to joint freq stats help page
I noticed the joint frequency statistics help page has been released (https://gnomad.broadinstitute.org/help/combined-freq-stats) and went to check it out. On the page, I noticed two small things:

Would it be possible to reformat the -4 in 10^-4 to be superscript? (and remove the ^)
image

Also as a quick note: the p value distribution image is missing from the help page. I believe @jkgoodrich is going to update this image (to be fixed in an upcoming PR)
image

rileyhgrant:
Resolves #1482 

Minor edit:
- "10^-4" -> "10<sup>-4</sup>"

### DIFF
--- a/browser/help/topics/combined-freq-stats.md
+++ b/browser/help/topics/combined-freq-stats.md
@@ -5,7 +5,7 @@ title: 'Combined genomes and exomes frequency statistics'
 
 **Contingency Table Test (Chi-squared or Fisher's exact test):**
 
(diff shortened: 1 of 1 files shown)


ch-kr (approved):
LGTM but I think @jkgoodrich is planning to update the figure (new version pending)

rileyhgrant:
Oops, thanks KC, will drop the figure change and make this a teeny tiny PR to just the superscript formatting

Issue #1448, opened 2024-03-20
Expose the non_ukb v4 short variant subset


mattsolo1:
@sambaxter suggested non_ukb subset should be available in the browser for v4 exome variants. 

I believe the frequencies are already in Elasticsearch.

https://github.com/broadinstitute/gnomad-browser/blob/560a828e49d1a489d0ca993441d29f20fa5459ef/data-pipeline/schemas/gnomad_v4_variants/annotate_gnomad_v4_variants/output/gnomad_v4_variants_annotated_1.ht.schema#L30

Therefore it should be easy to add this in the same way we typically expose subsets, and we don't need to load additional data.

ch-kr:
As discussed in this morning's standup: the non-UKB subset only impacts the exomes within gnomAD v4. Another subset in v2 that impacts only one data type is the non-cancer subset; this subset also only impacts the exomes in v2, because all v2 genomes are non-cancer.

Here's an example variant: https://gnomad.broadinstitute.org/variant/1-55505679-G-A?dataset=gnomad_r2_1
![image](https://github.com/broadinstitute/gnomad-browser/assets/19142091/5187eeff-88a5-4715-835a-618187e604ac)
and [non-cancer](https://gnomad.broadinstitute.org/variant/1-55505679-G-A?dataset=gnomad_r2_1_non_cancer):
![image](https://github.com/broadinstitute/gnomad-browser/assets/19142091/9d20e75e-2968-453b-b8cb-7e57c0aef8a0)

Commenting here in case looking at the behavior of the v2 non-cancer subset is helpful when implementing the non-UKB v4 subset

rileyhgrant:
Closed by #1490

PR #1490, opened 2024-04-16
Expose non-UK Biobank subset in browser

phildarnowsky-broad:
Here we add the non-UK Biobank subset of v4.1 to the dataset selector in the gene, variant, region, and transcript pages.

This also includes a refactoring of the dataset metadata to write out the metadata for each dataset explicitly, rather than having a function per metadata field that takes the dataset name as input--that was proving to be unsustainable. This approach is bulkier but much more straightforward and easier to maintain.


phildarnowsky-broad:
Note to self: get real sample counts before merging

phildarnowsky-broad:
resolves #1448

dataset-metadata/metadata.ts
```diff
@@ -19,6 +19,7 @@ export const datasetLabels = {
   gnomad_sv_r4: 'gnomAD SVs v4.0',
   gnomad_cnv_r4: 'gnomAD CNVs v4.0',
   gnomad_r4: 'gnomAD v4.0.0',
+  gnomad_r4_non_ukb: 'gnomAD 4.1 (non-UKB)',
```
rileyhgrant, line 22: Very minor -- based on the others in the list, I think we should add a "v" and maybe make it "4.1.0" , as in "`gnomad v4.1.0 (...`" just for consistency with the other entries in the list here

Screenshot 2024-04-18 at 08 59 12

graphql-api/src/queries/variant-datasets/gnomad-v4-variant-queries.ts
```diff
@@ -14,12 +14,14 @@ import largeGenes from '../helpers/large-genes'
 
 const GNOMAD_V4_VARIANT_INDEX = 'gnomad_v4_variants'
 
+type Subset = 'all' | 'non_ukb'
```
rileyhgrant, line 17: Nice on continually improving the types

rileyhgrant (approved):
LGTM!

Looked through commit by commit and only found one tiny thing -- an inconsistency in style of dataset selector name.

Nice on continuing to improve the type system, and on this subset. All the relevant pages (gene, transcript, region, variant) looked good in my local environment to my eyes.

PR #1495, opened 2024-04-17
more data nodes for 4.1 dataload

sjahl:


sjahl:
@rileyhgrant this is currently applied to prod, but i think we should merge this, since 4 nodes is looking like the new normal for the time being... need a +1 to merge though

rileyhgrant (approved):

rileyhgrant:
> @rileyhgrant this is currently applied to prod, but i think we should merge this, since 4 nodes is looking like the new normal for the time being... need a +1 to merge though

Oops, yep good call. +1'd

PR #1501, opened 2024-04-19
Genomes from "all" subset appear in variant occurrence table in v4 non-UKB

phildarnowsky-broad:


phildarnowsky-broad:
@rileyhgrant note that this isn't branched off of main, but off of the `v4.1-indices` branch which I've resurrected for this purpose

rileyhgrant:
Should this also return genome information for the bulk variant queries? Gonna tag @ch-kr in for review for a sanity check 

Screenshot 2024-04-19 at 11 17 12

In non-ukb, should this also include genome information in this table? AN summed across the two data types, and displaying that some come from genomes, instead of all the data only coming from exomes?

rileyhgrant (changes requested):
Codewise this looks good!

I think the bulk variant queries (by gene, region, transcript) might also need this treatment. Tagging in @ch-kr to double check

ch-kr:
sorry saw this ping in slack first, so copying my response here -- to maintain the same behavior as the previous exomes-only subset, I agree that the gene, region, and transcript views should also still display genome frequencies even when the non-UKB subset is selected

rileyhgrant (approved):
Nice, lgtm!

PR #1509, opened 2024-04-22
Move to artifact registry

sjahl:
with https://github.com/broadinstitute/tgg-terraform-modules/pull/39, update our deployctl config so that images are built and tagged with an artifact registry URL.

The intention is that each of our GCP projects will end up with a "gnomad" namespace / registry, with multiple images inside of it. So we'll have a `$LOCATION-docker.pkg.dev/$PROJECT/gnomad/gnomad-api`, `$LOCATION-docker.pkg.dev/$PROJECT/gnomad/gnomad-browser`, etc.

Note that one side effect of this change, is that you may need to add additional credential helper lines to your `~/.docker/config.json` to handle additional non-gcr.io URLs. See https://cloud.google.com/artifact-registry/docs/docker/pushing-and-pulling#cred-helper

My credHelpers config looks like

```
  "credHelpers": {
    "gcr.io": "gcloud",
    "us-central1-docker.pkg.dev": "gcloud",
    "us-east1-docker.pkg.dev": "gcloud",
    "us-docker.pkg.dev": "gcloud",
    "us-west1-docker.pkg.dev": "gcloud",
    "us.gcr.io": "gcloud"
  },
```

### DIFF
--- a/deploy/deployctl/config.py
+++ b/deploy/deployctl/config.py
@@ -82,27 +82,27 @@ class Configuration:
 
     @property
     def api_image_repository(self):
-        return f"gcr.io/{self.project}/gnomad-api"
+        return f"us-docker.pkg.dev/{self.project}/gnomad/gnomad-api"
 
     @property
     def browser_image_repository(self):
-        return f"gcr.io/{self.project}/gnomad-browser"
+        return f"us-docker.pkg.dev/{self.project}/gnomad/gnomad-browser"
 
     @property
     def reads_server_image_repository(self):
-        return f"gcr.io/{self.project}/gnomad-reads-server"
+        return f"us-docker.pkg.dev/{self.project}/gnomad/gnomad-reads-server"
 
     @property
     def reads_api_image_repository(self):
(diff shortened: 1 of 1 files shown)


phildarnowsky-broad:
@sjahl 

> there's a possibility you'll need to tell your local docker config to use the gcloud credential helper for more than just gcr.io domains

Is there an error or some such we'd expect to see if we do need to do this and haven't?

sjahl:
@phildarnowsky-broad 
> Is there an error or some such we'd expect to see if we do need to do this and haven't?

Yes, when attempting to push an image that you've build, you'd see something along the lines of:

`denied: Unauthenticated request. Unauthenticated requests do not have permission "artifactregistry.repositories.uploadArtifacts" on resource "projects/exac-gnomad/locations/us/repositories/gnomad" (or it may not exist)`

rileyhgrant (approved):
LGTM, I added the lines to my docker config and was able to just use deployctl as usual to build and push an image to the artifact registry

PR #1513, opened 2024-04-24
fixup(browser): typo in joint downloads link

sjahl:


rileyhgrant (approved):
Oops, my fault, good catch!

lgtm

PR #1518, opened 2024-04-24
Fixes to dataset labels

phildarnowsky-broad:
Made the `SearchBox` consistent with the `DatasetSelector`. Also in related discussion with @ch-kr we decided to use plain "ExAC" as the label for `exac`, rather than "ExAC v1.0".


rileyhgrant (approved):
lgtm!

PR #1524, opened 2024-04-29
Remove several genes from large genes list

rileyhgrant:
These genes are actually small enough that they will load without the cache, albeit somewhat slowly, and having them not on this list allows users to browse transcripts of them


rileyhgrant:
I jumped the gun on this one, and misunderstood a comment on an issue.

I thought I needed to revert this, but adding these genes to the large genes list actually produced the expected behavior, hence the closing of this PR

PR #1526, opened 2024-04-30
increase Ingress backend timeout from 30 -> 60

sjahl:
From debugging this morning -- changing this is an effort to try an ensure that we can load the TTN JSON cache and get it into the HTTP cache... hoping a few more seconds is enough of a bandaid there.


sjahl:
This just needs a `deployctl production apply-ingress` to deploy

rileyhgrant (approved):
lgtm!

Issue #1525, opened 2024-04-30
Cannot read properties of undefined (reading 'exons')


Jangseokhui:
**Description**: on the MYBPC3 gene page(gnomAD v4.1.0), click the 'Expand to all variants' button, and crash the page

**Error message**: Cannot read properties of undefined (reading 'exons')

**Stack trace**:
```
TypeError: Cannot read properties of undefined (reading 'exons')
    at T (https://gnomad.broadinstitute.org/js/972-0f01076c3cca278a3dd5.js:157:3641)
    at aa (https://gnomad.broadinstitute.org/js/bundle-41cfd6264b3744bcecbf.js:39:178930)
    at $u (https://gnomad.broadinstitute.org/js/bundle-41cfd6264b3744bcecbf.js:39:231023)
    at Il (https://gnomad.broadinstitute.org/js/bundle-41cfd6264b3744bcecbf.js:39:218302)
    at Cl (https://gnomad.broadinstitute.org/js/bundle-41cfd6264b3744bcecbf.js:39:218230)
    at Ol (https://gnomad.broadinstitute.org/js/bundle-41cfd6264b3744bcecbf.js:39:218093)
    at vl (https://gnomad.broadinstitute.org/js/bundle-41cfd6264b3744bcecbf.js:39:215080)
    at https://gnomad.broadinstitute.org/js/bundle-41cfd6264b3744bcecbf.js:39:164852
    at t.unstable_runWithPriority (https://gnomad.broadinstitute.org/js/bundle-41cfd6264b3744bcecbf.js:39:319258)
    at $o (https://gnomad.broadinstitute.org/js/bundle-41cfd6264b3744bcecbf.js:39:164629)
```

**Route**: /gene/ENSG00000134571?dataset=gnomad_r4

**Browser**: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/124.0.0.0 Safari/537.36

rileyhgrant:
@Jangseokhui 

Thanks for reporting this issue! Apologies for the slow reply -- this is now fixed in prod.

https://gnomad.broadinstitute.org/gene/ENSG00000134571?dataset=gnomad_r4

Thanks for your patience

PR #1532, opened 2024-05-01
Fix ClinVar track crash

rileyhgrant:
Resolves: #1525 

Frameshift variants are displayed with a line that is either solid or dotted, depending on if the region being shifted through is an exon that is part of the transcript or not.

In rare cases, there can be a frameshift variant from ClinVar on a transcript that the browser doesn't display, here we manually set the exon array to an empty array to prevent a crash when trying to render the line.

Alternatively, we could edit the logic to always return one type of line or another for the entire length of the frameshift, but without knowledge of which exons are part of the transcript, we can very possibly make a mistake and be inconsistent with how we display things. However, by not displaying any line, we are already being inconsistent with how we display frameshift variants. We should discuss on which is the lesser of the two evils.

One thing to try and better understand is why we don't display some transcripts that ClinVar has entries for.

### DIFF
--- a/browser/src/ClinvarVariantsTrack/ClinvarAllVariantsPlot.tsx
+++ b/browser/src/ClinvarVariantsTrack/ClinvarAllVariantsPlot.tsx
@@ -210,7 +210,9 @@ const VariantLine = ({
   }
 
   if (category === 'frameshift') {
-    const transcript = transcripts.find((t) => t.transcript_id === variant.transcript_id)
+    const transcript: Transcript | undefined = transcripts.find(
+      (t) => t.transcript_id === variant.transcript_id
+    )
     const [endpoint1, endpoint2] = getGlobalFrameshiftCoordinates(variant, transcript)
     const frameshiftMinPos = Math.min(endpoint1, endpoint2)
     const frameshiftMaxPos = Math.max(endpoint1, endpoint2)
--- a/browser/src/ClinvarVariantsTrack/ClinvarAllVariantsPlot.tsx
+++ b/browser/src/ClinvarVariantsTrack/ClinvarAllVariantsPlot.tsx
@@ -216,15 +216,19 @@ const VariantLine = ({
     const frameshiftMaxPos = Math.max(endpoint1, endpoint2)
     const terminationPos =
       transcript && transcript.strand === '+' ? frameshiftMaxPos : frameshiftMinPos
-    // @ts-expect-error TS(2532) FIXME: Object is possibly 'undefined'.
-    const frameshiftExonRegions = transcript.exons
-      .sort((e1, e2) => e1.start - e2.start)
-      .filter((e) => e.start <= frameshiftMaxPos && e.stop >= frameshiftMinPos)
-      .map((e) => ({
-        start: Math.max(e.start, frameshiftMinPos),
-        stop: Math.min(e.stop, frameshiftMaxPos),
-        feature_type: e.feature_type,
-      }))
+
+    // if a frameshift variant-consequence pair from ClinVar exists for a transcript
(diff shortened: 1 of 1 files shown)


rileyhgrant:
Tagging @mattsolo1 for review because of a big picture question:

Framshift variants in the expanded clinvar plot are rendered with a line that differs in texture based on if its going through an exon or not, this is transcript dependent: 
- Screenshot 2024-05-02 at 10 02 24

In this (seemingly rare) case, there is a frameshift variant from ClinVar with an associated transcript that the browser doesn't display. As far as I can tell, the two main options are:

- A. Do not display a line, this avoids the problem of rendering the wrong texture of line since we have no knowledge of what exons the transcript includes. This is problems in that its inconsistent with how other frameshift variants are rendered if it has no line

- B. Display a line for the length of the frameshift variant, and either render the line all in one texture, or use the canonical or mane select transcript to determine how to texture the line. This has problems in that the line may show the wrong information, since the browser has no knowledge of what the actual transcript includes to inform the texture of the line

I realize this is probably pretty niche, but I wanted to check in on what you thought is the better option (or if there's some other better thing to do here), fwiw I think I'm in favor of option B, and using the canonical or mane select transcript to inform the texture of the line

rileyhgrant:
From discussion with Moriel,

Since it seems as though this is super uncommon, really picking either of the options seems reasonable. 

A third option brought up is to not even display ClinVar variants that are not on the transcripts we display. In the spirit of trying to keep the display consistent (but possibly incorrect), I'll likely go with option B -- display a line based on the canonical transcript.

browser/src/ClinvarVariantsTrack/ClinvarAllVariantsPlot.tsx
```diff
@@ -216,15 +216,19 @@ const VariantLine = ({
...
-    const frameshiftExonRegions = transcript.exons
-      .sort((e1, e2) => e1.start - e2.start)
-      .filter((e) => e.start <= frameshiftMaxPos && e.stop >= frameshiftMinPos)
-      .map((e) => ({
-        start: Math.max(e.start, frameshiftMinPos),
-        stop: Math.min(e.stop, frameshiftMaxPos),
-        feature_type: e.feature_type,
-      }))
+
+    // if a frameshift variant-consequence pair from ClinVar exists for a transcript
```
phildarnowsky-broad, line 220: So this fix is fine far as it goes, but I'd suggest you also do two other things:

1. add a type annotation a couple lines above here reflecting that `transcript` might be undefined, i.e. change line 213 to something like: `const transcript: Transcript | undefined = ...`. That way we won't get bitten here by a similar case in the future.
2. Update `getGlobalFrameshiftCoordinates` also to reflect that `transcript` might be `undefined`, the logic might be fine as-is but a type annotation on there would be good in any case.

phildarnowsky-broad (changes requested):
Looks like a good fix but I had a couple of suggestions about expanding on the fix.

browser/src/ClinvarVariantsTrack/ClinvarAllVariantsPlot.tsx
rileyhgrant: Aha very good call, I didn't think to add a type annotation on the variable itself. 

I added a fixup commit to address point 1.

Point 2. is already addressed before this PR (getGlobalFrameshift... expects the possibility of `undefined` and handles it if so), so that's why theres not fixup commit to address that.

rileyhgrant:
Applied changes, with note in response to comment about why one change was not included (`getGlobal...` already handles `undefined` transcripts)

Re-requesting review at your convenience!

phildarnowsky-broad (approved):

PR #1541, opened 2024-05-07
Use joint frequency data consistently

### LINKED ISSUE #1540: Error in display of combine genomes and exomes AN/AF
When a variant is genome or exome only the allele frequencies and allele numbers are displayed incorrectly in the variant table and gene page variant table.  There are slight variations on this issue (details bewlo with screen shots)

- Screenshots
  For the variant [6-145635482-G-GAA](https://gnomad.broadinstitute.org/variant/6-145635482-G-GAA?dataset=gnomad_r4) the combined AN should be 1508760 and AF 6.628e-7, and that is displayed correctly in the table at the top of the page, however in the "Genetic Ancestry Group Frequencies" table and the [variant table on the gene page](https://gnomad.broadinstitute.org/gene/ENSG00000112425?dataset=gnomad_r4) only the genome frequency is displayed.
image
image

For variants with an [AC0 flag in either exome or genome](https://gnomad.broadinstitute.org/variant/6-145735197-C-A?dataset=gnomad_r4) the data is displayed correctly on the variant page but not on the gene page 
image
image

rileyhgrant:
Resolves #1540 

Modifies the `combineExomeAndGenomeData` function to use joint data if theres only exome or genome data, and theres joint data

---

Screenshots:
- Screenshot 2024-05-08 at 15 02 15

- Screenshot 2024-05-08 at 15 03 12
- Screenshot 2024-05-08 at 15 03 45

### DIFF
--- a/browser/src/VariantList/mergeExomeAndGenomeData.ts
+++ b/browser/src/VariantList/mergeExomeAndGenomeData.ts
@@ -1,4 +1,5 @@
-import { Population, SequencingType } from '../VariantPage/VariantPage'
+import { Filter } from '../QCFilter'
+import { Population, SequencingType, Variant } from '../VariantPage/VariantPage'
 
 // safe math on possibly null values
 const add = (n1: any, n2: any) => (n1 || 0) + (n2 || 0)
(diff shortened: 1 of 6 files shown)


browser/src/VariantList/mergeExomeAndGenomeData.ts
```diff
@@ -1,4 +1,5 @@
-import { Population, SequencingType } from '../VariantPage/VariantPage'
+import { Filter } from '../QCFilter'
```
phildarnowsky-broad, line 1: nice refactor here, much clearer what is happening now

browser/src/VariantPage/VariantPage.tsx
```diff
@@ -146,8 +146,10 @@ export type Population = {
   id: PopulationId
   ac: number
   an: number
-  ac_hemi: number | null
-  ac_hom: number
+  ac_hemi?: number | null
```
phildarnowsky-broad, line 149: Do we in fact want the `?` here? I.E. can these really be `number | null | undefined`?

browser/src/VariantPage/GnomadPopulationsTable.tsx
```diff
@@ -102,14 +103,24 @@ export class GnomadPopulationsTable extends Component<
...
 
     const mergedPopulations = mergeExomeGenomeAndJointPopulationData({
       exomePopulations: includeExomes ? exomePopulations : [],
       genomePopulations: includeGenomes ? genomePopulations : [],
-    })
+      jointPopulations,
+      // @ts-expect-error: silence TS's warning -- there is no legitimate "" id in
+      //   PopulationId, but the source array from joint population has a summary value
+      //   with an id of "", filter it out here to display the correct values
+    }).filter((mergedAncestry) => mergedAncestry.id !== '')
```
phildarnowsky-broad, line 123: You might be able to avoid adding the `ts-expect-error` by adding an explicit cast: `filter((mergedAncestry) => (mergedAncestry.id as string) !== '')`. Definitely the lesser evil here if it works.

phildarnowsky-broad (commented):
Looks generally good but there are two small things I think are worth checking into before we call this good.

browser/src/VariantPage/VariantPage.tsx
rileyhgrant: Ah good point, I rather short-sightedly added the `?` since it was getting assigned from another type that can be `undefined` in a particular circumstance, though on second thought in that circumstance (joint data exists) we know the value won't be undefined -- good catch.

Will address in a fixup commit.

rileyhgrant:
Thanks for the prompt review -- I addressed the comments and re-tagged you!

phildarnowsky-broad (approved):

Issue #1530, opened 2024-05-01
Export variants to CSV for gnomAD v4.1.0 (non-UKB) not working


eva-mariastauffer:
I looked up KANSL1 in gnomAD v4.1.0 (non-UKB) https://gnomad.broadinstitute.org/gene/ENSG00000120071?dataset=gnomad_r4_non_ukb. I selected pLoF only variants and wanted to export variants to CSV.
Clicking the "Export variants to CSV" is not working and no CSV is produced. It does work for v4.1.0 but not for the non-UKB  version.

phildarnowsky-broad:
Hi @eva-mariastauffer , thanks for reporting this. I wanted to let you know that we have a fix under review for this so hopefully this functionality should work properly on the public site within the next few days.

phildarnowsky-broad:
Hi @eva-mariastauffer , the fix is deployed and I have verified that I can download the non-UKB variant CSV data from the gene in question. Thanks again for bringing this to our attention, please let us know if you have further trouble.

PR #1544, opened 2024-05-10
Fix CSV export of non-UKB v4 variants

phildarnowsky-broad:
The root of the problem was that the browser was trying to export some joint-frequency columns, which the non-UKB subset doesn't have. This fix uses the corresponding columns from the exome instead.


phildarnowsky-broad:
fixes #1530

rileyhgrant (approved):
lgtm!

Nicely done.

PR #1547, opened 2024-05-17
Fix single source gen-anc frequencies on the Variant page

### LINKED ISSUE #1545: De-selecting a source on variant page does not update frequency information
https://atgu.slack.com/archives/CNL7NA6H2/p1715601882889859

On the variant page, one can de-select either the exomes or genomes checkbox, if there is frequency information for both.

https://gnomad.broadinstitute.org/variant/1-55051215-G-GA?dataset=gnomad_r4

Currently, the browser uses the joint information preferentially per #1540, this accidentally introduced this regression.

rileyhgrant:
Resolves #1545


rileyhgrant:
Tagging you in for review, @ch-kr, in addition to Phil, because I wanted to check on one thing display decision wise.

The v4 exomes and genomes have identical genetic ancestry groups with the exception of `ami`, which is included presumably because the data was already there from v3 genomes. When both boxes are selected, or just the genomes box is selected, the table will include `ami` with the respective values there.

When the genomes checkbox is unselected, and only the v4 exome data is shown, should the `ami` row in the table disappear? The alternative would be to include a row for `ami` and populate it with `AC: 0, AN: 0, AF: -`

I'm not sure there's any value to showing it if it's always AC and AN of 0, but also including it keeps it more consistent.

Either is equally doable, I wanted to check on your opinion of what is more clear/preferable.

ch-kr:
Thanks for the ping!

> When the genomes checkbox is unselected, and only the v4 exome data is shown, should the ami row in the table disappear? The alternative would be to include a row for ami and populate it with AC: 0, AN: 0, AF: -

Quick question before I answer: based on this [variant](https://gnomad.broadinstitute.org/variant/1-55043996-G-A?dataset=gnomad_r4), it looks like the `ami` row currently exists for variants only present in the exomes. Does this question apply to these variants in addition to when a user explicitly selects or unselects the data type checkboxes?

rileyhgrant:
> Thanks for the ping!
> 
> > When the genomes checkbox is unselected, and only the v4 exome data is shown, should the ami row in the table disappear? The alternative would be to include a row for ami and populate it with AC: 0, AN: 0, AF: -
> 
> Quick question before I answer: based on this [variant](https://gnomad.broadinstitute.org/variant/1-55043996-G-A?dataset=gnomad_r4), it looks like the `ami` row currently exists for variants only present in the exomes. Does this question apply to these variants in addition to when a user explicitly selects or unselects the data type checkboxes?

Gah ok yeah I hadn't thought of that, good catch.

The question specifically was for behavior when the variant is present in both exomes and genomes.

Now I suppose there's another question of what to do for the variants that are not present in one data type, and thus don't allow de-selecting the type its not in. My initial thought there is leave it untouched, e.g. the example variant still includes the genome AN data? But also then if there's joint data maybe it makes sense to allow users to remove the ANs from the other data type that contribute to the overall AN per ancestry group.

ch-kr:
thanks for clarifying! I think the most consistent behavior is to continue displaying the `ami` row both for variants only present in the exomes and for variants present in both data types (and where users can choose to unselect the genomes data). This reinforces that the v4 dataset is comprised of both exomes and genomes and also matches the behavior seen when toggling between the full dataset and a subset. For example, here's an example v3 [variant](https://gnomad.broadinstitute.org/variant/1-10149-CCT-C?dataset=gnomad_r3) -- the `ami` AC is 1 when all of v3 is selected and becomes 0 when the v3 [non-TOPMed](https://gnomad.broadinstitute.org/variant/1-10149-CCT-C?dataset=gnomad_r3_non_topmed) subset is selected.

On a slightly related note, this question sort of ties into this slack [thread](https://atgu.slack.com/archives/CNNTF8Z46/p1715792706949689) -- we should likely display the genetic ancestry group sample count/counts of each sex karyotype per group for v3/v4 on our stats page (across all groups, without any merged groups, which is what the table on the stats page currently contains)
image

rileyhgrant:
> thanks for clarifying! I think the most consistent behavior is to continue displaying the `ami` row both for variants only present in the exomes and for variants present in both data types (and where users can choose to unselect the genomes data). This reinforces that the v4 dataset is comprised of both exomes and genomes and also matches the behavior seen when toggling between the full dataset and a subset. For example, here's an example v3 [variant](https://gnomad.broadinstitute.org/variant/1-10149-CCT-C?dataset=gnomad_r3) -- the `ami` AC is 1 when all of v3 is selected and becomes 0 when the v3 [non-TOPMed](https://gnomad.broadinstitute.org/variant/1-10149-CCT-C?dataset=gnomad_r3_non_topmed) subset is selected.
> 
> On a slightly related note, this question sort of ties into this slack [thread](https://atgu.slack.com/archives/CNNTF8Z46/p1715792706949689) -- we should likely display the genetic ancestry group sample count/counts of each sex karyotype per group for v3/v4 on our stats page (across all groups, without any merged groups, which is what the table on the stats page currently contains) image

Perfect, we'll go with that behavior, thanks so much for the reply and opinion/decision.

browser/src/VariantList/mergeExomeAndGenomeData.ts
```diff
@@ -1,29 +1,61 @@
+import { DatasetId } from '@gnomad/dataset-metadata/metadata'
 import { Filter } from '../QCFilter'
 import { Population, Variant } from '../VariantPage/VariantPage'
+import { PopulationId, getPopulationsInDataset } from '@gnomad/dataset-metadata/gnomadPopulations'
 
 // safe math on possibly null values
 const add = (n1: number | null | undefined, n2: number | null | undefined) => (n1 || 0) + (n2 || 0)
 
+// include placeholders for any ancestries missing from the dataset
```
phildarnowsky-broad, line 9: This looks like it should work but as a rule it's good to avoid approaches using mutable data unless using immutable data makes the algorithm much more complex or less efficient. (Don't want to bloat up this comment more by reviewing things you may already have thought about, but if the rationale for why this is so isn't clear I'm happy to say more about it).

Now assuming that the ordering of the ancestries in the output doesn't matter, one thing you could do is something like:

```
const fullAncestries = versionAncestries.flatMap((versionAncestry) =>
 findAncestries(fullAncestries, versionAncestry) || emptyAncestries
)
```

...where you have `emptyAncestry` and `findAncestry` defined appropriately.

browser/src/VariantList/mergeExomeAndGenomeData.ts
```diff
@@ -1,29 +1,61 @@
...
+        ac_hemi: jointPopulation.hemizygote_count!,
+        ac_hom: jointPopulation.homozygote_count!,
+      }))
+
+    const reshapedJointPopulaitonsWithAllAncestries = addMissingAncestries(
+      reshapedJointPopulations,
+      datasetPopulations
     )
+
+    return reshapedJointPopulaitonsWithAllAncestries
```
phildarnowsky-broad, line 58: typo, `Populaitons`

dataset-metadata/gnomadPopulations.ts
```diff
@@ -32,6 +32,8 @@ export const GNOMAD_POPULATION_NAMES = {
 
 export type PopulationId = keyof typeof GNOMAD_POPULATION_NAMES
 
+export type PopulationIdAndChromosome = PopulationId | `${PopulationId}_XX` | `${PopulationId}_XY`
```
phildarnowsky-broad, line 35: nice typesystem trick here

dataset-metadata/metadata.ts
```diff
@@ -1139,3 +1139,16 @@ export const baseDatasetForReferenceGenome = (genome: ReferenceGenome): DatasetI
 
 export const hasJointFrequencyData = (datasetId: DatasetId): boolean =>
   getMetadata(datasetId, 'hasJointFrequencyData')
+
+export const getTopLevelDataset = (datasetId: DatasetId) =>
+  /* eslint-disable no-nested-ternary */
```
phildarnowsky-broad, line 1144: this is hardly the most egregious nested ternary I've ever seen but still, let's go with this kind of style instead:

```
if(isV4(datasetId)) {
  return 'v4'
}

if(isV3(datasetId)) {
  return 'v3'
}
...
// no if needed here since we're at the default case
return 'default'
```

Multiple returns in a function are a little controversial but personally I think that's fine especially in a situation like this where all the different cases are going to be very similar in form. Some languages allow "pattern matching" of values, which is kind of a formalization of this style built into the language.

phildarnowsky-broad (changes requested):
small changes requested but at root this looks good

rileyhgrant:
Alrighty, added some fixup commits to address comments from code review. Thanks, as always, for your review! 

Re-requested you @phildarnowsky-broad

phildarnowsky-broad (approved):

PR #1554, opened 2024-05-24
feat(data-pipelines): allow pinning of hail in requirements

rileyhgrant:
Allows pinning of hail in `data-pipeline/requirements.txt`, and pins it.

`deployctl`'s `start_cluster` command now checks if the locally installed version of hail is the same as the version pinned in `requirements.txt`, and raises and error if not. 

Since `deployctl` calls `hailctl` in its `start_cluster` command, this approach allows pinning of hail by manually checking the versions match.

### DIFF
--- a/deploy/deployctl/subcommands/dataproc_cluster.py
+++ b/deploy/deployctl/subcommands/dataproc_cluster.py
@@ -3,6 +3,7 @@ import os
 import subprocess
 import sys
 import typing
+import importlib.metadata
 
 from deployctl.config import config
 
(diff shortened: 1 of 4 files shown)


rileyhgrant:
Bumping pylint because of https://github.com/pylint-dev/pylint/issues/3882

deploy/deployctl/subcommands/dataproc_cluster.py
```diff
@@ -24,6 +25,19 @@ def start_cluster(name: str, cluster_args: typing.List[str]) -> None:
         raise RuntimeError("project configuration is required")
 
     with open(os.path.join(DATA_PIPELINE_DIRECTORY, "requirements.txt")) as requirements_file:
+        requirements_hail_version = next(
+            (line.strip().split("==")[1] for line in requirements_file if line.strip().startswith("hail==")), None
+        )
+        local_hail_version = importlib.metadata.version("hail")
```
sjahl, line 31: It looks like `.version()` raises an error (PackageNotFoundError) if hail isn't installed (rather than returning None):

```
>>> local_hail_version = importlib.metadata.version("hail")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/sjahl/.asdf/installs/python/3.9.16/lib/python3.9/importlib/metadata.py", line 569, in version
    return distribution(distribution_name).version
  File "/Users/sjahl/.asdf/installs/python/3.9.16/lib/python3.9/importlib/metadata.py", line 542, in distribution
    return Distribution.from_name(distribution_name)
  File "/Users/sjahl/.asdf/installs/python/3.9.16/lib/python3.9/importlib/metadata.py", line 196, in from_name
    raise PackageNotFoundError(name)
importlib.metadata.PackageNotFoundError: hail
```

You may want to catch that and handle it, since I think we'll just crash there instead of making it to the `if not` check on `local_hail_version`

sjahl (changes requested):
Just the one thing inline re: catching an error. Normally, I'd pause before jumping from python 3.7 -> 3.9, but 3.7 is ancient, and it looks like i've been running 3.9 in my local venv for who knows how long and deployctl has worked fine!

PR #1556, opened 2024-05-24
Hide readviz by default

### LINKED ISSUE #633: Hide readviz by default?
BAI files for v3.1 are very large (~6 MB) and more than one may be loaded for a variant. Consider hiding the reads data section by default to avoid downloading so much data when it may not be needed.

Ideally, there should be an option to always show readviz to support users that frequently look at reads.

rileyhgrant:
Resolves #633 

Hides readviz by default on Variant pages to try and save on egress fees.

Gives two options,
- a button to load the data just this time
- a checkbox to always load readviz data, this uses `localStorage` to save the value of this on load and when toggling this

Also adds tracking to these two buttons to try and see how often people use these

### DIFF
--- a/browser/src/ReadData/ReadData.tsx
+++ b/browser/src/ReadData/ReadData.tsx
@@ -154,7 +154,26 @@ type OwnReadDataProps = {
   showHemizygotes?: boolean
 }
 
-type ReadDataState = any
+type SequencingType = 'exome' | 'genome'
+
+type TrackCategory = 'het' | 'hom' | 'hemi'
+
+type ReadDataTracks = {
+  het: number
+  hom: number
+  hemi: number
+}
+
+type ReadDataState = {
+  tracksAvailable: {
+    exome: ReadDataTracks
+    genome: ReadDataTracks
+  }
+  tracksLoaded: {
+    exome: ReadDataTracks
+    genome: ReadDataTracks
+  }
+}
 
 type ReadDataProps = OwnReadDataProps & typeof ReadData.defaultProps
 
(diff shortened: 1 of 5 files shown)


browser/src/ReadData/ReadData.tsx
```diff
@@ -307,8 +325,8 @@ class ReadData extends Component<ReadDataProps, ReadDataState> {
   }
 
   loadInitialTracks() {
-    ;['exome', 'genome'].forEach((exomeOrGenome) => {
-      ;['het', 'hom', 'hemi'].forEach((category) => {
+    ;(['exome', 'genome'] as SequencingType[]).forEach((exomeOrGenome) => {
```
phildarnowsky-broad, line 343: if this works, it'd be more idiomatic to write `['exome', 'genome'].forEach((exomeOrGenome: SequencingType) => ...`. I _think_ TS should be able to infer that properly. Same deal in the next line.

browser/src/ReadData/ReadData.tsx
```diff
@@ -517,102 +541,126 @@ const ReadDataContainer = ({ datasetId, variantIds }: ReadDataContainerProps) =>
...
+        <Button id="load-once" disabled={loadOnce} onClick={() => setLoadOnce(!loadOnce)}>
+          Load once
+        </Button>
+        <div>
+          <br />
+          <span>
+            <CheckboxInput
+              id="always-load"
+              checked={alwaysLoad}
+              onChange={(e: any) => {
```
phildarnowsky-broad: there should be a type you can use for `e` here, `MouseEvent` or something similar IIRC

phildarnowsky-broad (changes requested):

browser/src/ReadData/ReadData.tsx
rileyhgrant: I would have liked that better too, but unfortunately TS complains about doing it that way, as the `forEach` is being called on an array of strings.

rileyhgrant:
Screenshots of the UI: 

- Screenshot 2024-06-17 at 15 48 28
- Screenshot 2024-06-17 at 15 48 17
- Screenshot 2024-06-17 at 15 48 49

phildarnowsky-broad (approved):
Cut, print, that's a wrap

sjahl (approved):
This looks good to me from a UI / Text copy perspective

PR #1561, opened 2024-05-31
Add python dependabot configs for our various roots, set freq to weekly for all

sjahl:
This sets dependabot frequencies to weekly, since daily seemed excessive. It also adds pip/pip-compile/github-actions package configs for our various requirements files, so that they get managed independently (or, at least, the hope is that dependabot will do that with this config)


rileyhgrant (approved):
lgtm!

Issue #1462, opened 2024-04-01
Adding hemizygous counts in variant table on gene page for gnomAD SV's


msingerb:
For all gene pages we have the option to configure the variant table and see how many homozygous/hemizygous individuals there are for a single variant. It is incredibly helpful to show the number of hemizygous individuals for X-linked genes in particular. The hemizygous individuals are the ones that are male and carry a single copy of the variant on the X-chromosome. This information is currently provided for the normal browser, but not for the SV browser. I am asking for it to be added as a viewable option in the configure options for the variant table on the gene page for the SV browser. The photos below show it as an option in the normal browser and missing from the SV browser. This information is stored in the individual variant pages for SV's (see screenshot), but not viewable on the gene pages. This makes it impossible to easily filter to variants that have been seen in hemizygosity across the gene. Instead we have to open every variant page in the gene to determine the zygosity of the variant.

![image](https://github.com/broadinstitute/gnomad-browser/assets/47390031/4454ec73-d127-4fe1-a223-87c4bac4d66d)
![image](https://github.com/broadinstitute/gnomad-browser/assets/47390031/9ea2e12f-661a-4bc6-b0fc-69386f16401a)
![image](https://github.com/broadinstitute/gnomad-browser/assets/47390031/07073e52-b8d6-4627-8705-82f34eeda640)

phildarnowsky-broad:
Hi @msingerb , this should be working in production now. Thanks for bringing it to our attention, and have a nice weekend!

PR #1566, opened 2024-06-11
Add hemizygous count to structural variant tables

phildarnowsky-broad:
Per https://github.com/broadinstitute/gnomad-browser/issues/1462, the hemizygous AC was available in short variant but not structural variant tables. This adds that field to SV tables.

Note that I'm breaking my own first rule here and not adding test coverage: these components turn out to be coupled together with some other logic in a way that makes writing a coherent test for them a real pain. I don't love doing this but this is a very basic change that fixes something the end-user already needs, and it's not really clear to me yet how to best refactor this to make testing it less of a circus.

### DIFF
--- a/browser/src/GenePage/StructuralVariantsInGene.tsx
+++ b/browser/src/GenePage/StructuralVariantsInGene.tsx
@@ -29,6 +29,7 @@ const StructuralVariantsInGene = ({ datasetId, gene, zoomRegion, ...rest }: Prop
       gene(gene_id: $geneId, reference_genome: $referenceGenome) {
         structural_variants(dataset: $datasetId) {
           ac
+          ac_hemi
           ac_hom
           an
           af
--- a/browser/src/RegionPage/StructuralVariantsInRegion.tsx
+++ b/browser/src/RegionPage/StructuralVariantsInRegion.tsx
@@ -30,6 +30,7 @@ const StructuralVariantsInRegion = ({ datasetId, region, zoomRegion, ...rest }: 
       region(chrom: $chrom, start: $start, stop: $stop, reference_genome: $referenceGenome) {
         structural_variants(dataset: $datasetId) {
           ac
+          ac_hemi
           ac_hom
           an
           af
--- a/browser/src/StructuralVariantList/structuralVariantTableColumns.tsx
+++ b/browser/src/StructuralVariantList/structuralVariantTableColumns.tsx
@@ -103,6 +103,16 @@ const structuralVariantTableColumns = [
   },
 
   {
+    key: 'hemizygote_count',
+    heading: 'Number of Hemizygotes',
+    contextNotes: 'Only shown when viewing X or Y chromosomes',
+    minWidth: 100,
+    compareFunction: makeNumericCompareFunction('ac_hemi'),
+    render: (variant: any) => renderAlleleCountCell(variant, 'ac_hemi'),
+    shouldShowInContext: (context: any) => context.chrom === 'X' || context.chrom === 'Y',
+  },
+
+  {
     key: 'length',
     heading: 'Size',
     minWidth: 100,
--- a/browser/src/StructuralVariantList/StructuralVariants.tsx
+++ b/browser/src/StructuralVariantList/StructuralVariants.tsx
@@ -56,6 +56,7 @@ const DEFAULT_COLUMNS = [
   'an',
   'af',
   'homozygote_count',
+  'hemizygote_count',
 ]
 
 const sortVariants = (variants: any, { sortKey, sortOrder }: any) => {


phildarnowsky-broad:
fixes #1462

browser/src/GenePage/StructuralVariantsInGene.tsx
```diff
@@ -29,6 +29,7 @@ const StructuralVariantsInGene = ({ datasetId, gene, zoomRegion, ...rest }: Prop
       gene(gene_id: $geneId, reference_genome: $referenceGenome) {
         structural_variants(dataset: $datasetId) {
           ac
+	  ac_hemi
```
rileyhgrant, line 32: Nitpick: I think the formatter doesn't run due to our noted strange practicing of having a big ol' string that represents the query -- I'd prefer consistent indentation here but fully realize this is nitpicky and this is non-blocking

browser/src/RegionPage/StructuralVariantsInRegion.tsx
```diff
@@ -30,6 +30,7 @@ const StructuralVariantsInRegion = ({ datasetId, region, zoomRegion, ...rest }:
       region(chrom: $chrom, start: $start, stop: $stop, reference_genome: $referenceGenome) {
         structural_variants(dataset: $datasetId) {
           ac
+	  ac_hemi
```
rileyhgrant, line 33: Same as other comment, consistent indentation here would be nice

browser/src/StructuralVariantList/structuralVariantTableColumns.tsx
```diff
@@ -102,6 +102,16 @@ const structuralVariantTableColumns = [
...
   },
 
+  {
+    key: 'hemizygote_count',
+    heading: 'Number of Hemizygotes',
+    contextNotes: 'Not shown when viewing Y chromosome',
+    minWidth: 100,
+    compareFunction: makeNumericCompareFunction('ac_hemi'),
+    render: (variant: any) => renderAlleleCountCell(variant, 'ac_hemi'),
+    shouldShowInContext: (context: Context) => context.chrom !== 'Y',
```
rileyhgrant, line 112: These attributes for the context and whether or not to show in the context appear to be duplicated from the above column (homozygote count) when I think they should be different.

For the short variants, this column has a note that it only appears for the sex chromosomes, and logic to achieve this. 

From `variantTableColumns.tsx`:

```
{
    key: 'hemizygote_count',
    heading: 'Number of Hemizygotes',
    description: 'Number of individuals hemizygous for alternate allele',
    contextNotes: 'Only shown when viewing X or Y chromosomes',
    grow: 0,
    minWidth: 100,
    compareFunction: makeNumericCompareFunction('ac_hemi'),
    render: (variant: Variant) => renderAlleleCountCell(variant, 'ac_hemi'),
    shouldShowInContext: (context: any) => context.chrom === 'X' || context.chrom === 'Y',
  },
```

I think we should at least mimic this behavior to stay consistent, this may also be an opportunity to re-use this column definition, rather than duplicating it.

rileyhgrant: I also think we should add this column to the default columns rendered for SVs, and let the individual column definition logic handle whether or not its rendered.

The default columns are defined in `StructuralVariants.tsx`, on line 49:

```
const DEFAULT_COLUMNS = [
  'source',
  'consequence',
  'class',
  'pos',
  'length',
  'ac',
  'an',
  'af',
  'homozygote_count'
]
```

We could add `hemizygote_count` to the above list to achieve this.

rileyhgrant (changes requested):
Two small nitpicks, and one small change requested to maintain consistency of when this new column renders on the structural variant pages as compared to the short variant pages.

browser/src/GenePage/StructuralVariantsInGene.tsx
phildarnowsky-broad: @rileyhgrant this looks like it's aligned correctly now--I did see, earlier, some odd discrepancies between the diffs and my editor. not sure what to make of all this but I will clean it up if there is anything to clean up.

browser/src/StructuralVariantList/structuralVariantTableColumns.tsx
phildarnowsky-broad: @rileyhgrant it's tempting to me to go through these lists of column specifiers and look for more copypasta to extract, but in the absence of tests and presence of other things I have to do, I'm leaving that for another day.

phildarnowsky-broad: @rileyhgrant ...actually, the two definitions of `hemizygote_count` are similar but not compatible, I'm just going to leave them separate and fix up the new SV version and call it a day until such time as I have the wherewithal to really overhaul this code.

phildarnowsky-broad:
@rileyhgrant ready for another gander

rileyhgrant (approved):
LGTM!

PR #1577, opened 2024-06-18
feat(data-pipelines): Allow for starting Dataproc clusters using custom OS images

nadeaujoshua:
This PR adds some helper functions within the `./deployctl dataproc-cluster start ...` command to facilitate the creation of Dataproc clusters using [a custom OS image](https://cloud.google.com/dataproc/docs/guides/dataproc-images).

The primary advantage of using pre-built images is that they'll provide a standard, consistent environment to start data pipelines with, using the correct/validated configs, dependencies, and versioning so that the user does not have to manage these factors themselves. The Dataproc start cluster time has also been cut down a few minutes.

These helpers are needed because `deployctl`'s start command uses `hailctl`, which would run redundant commands that have already been executed in the custom OS image.

To provide a transition period for our workflows, the `dataproc-cluster start ...` command has been updated to make this opt-in; the user must specify the `--image` flag with the custom image name they'd like to use. The inclusion of this flag is what triggers running the helper function to start the cluster. If it is not used, as was the case before this update, the original `deployctl` code will be run for now.

This PR also removes `poetry` from the `data-pipeline` directory, as we will revert to managing dependencies using the `requirements.txt` file located there.

Resolves https://github.com/broadinstitute/gnomad-browser-team/issues/64.


deploy/deployctl/subcommands/dataproc_cluster.py
```diff
@@ -20,10 +85,110 @@ def list_clusters() -> None:
...
+
+
 def start_cluster(name: str, cluster_args: typing.List[str]) -> None:
     if not config.project:
         raise RuntimeError("project configuration is required")
 
+    cluster_args_dict = _parse_flags(cluster_args)
+    if "image" in cluster_args_dict:
+        return start_custom_image_cluster(name, cluster_args_dict)
+    # TODO: Add "else" block to confirm with user to use legacy hailctl initialization
```
rileyhgrant, line 190: Just noting a `TODO:` left here, wasn't sure if you meant to get back to this before merging this PR.

data-pipeline/README.md
```diff
@@ -58,6 +58,11 @@ The ClinVar pipelines must be run on clusters with an appropriate version of VEP
   ./deployctl dataproc-cluster start <cluster-name>
   ```
 
+  NOTE: Starting with `gnomad version 4.1`, it is highly recommended to specify a
+  custom OS image that was built for that gnomAD version. To see these images, run
+  `gcloud compute images list --filter="family=dataproc-custom-image"`. You can then
```
rileyhgrant, line 63: Running this tells me 0 images were found:

```
❯ gcloud compute images list --filter="family=dataproc-custom-image"
Listed 0 items.
```

My gcloud config's project is set to `exac-gnomad`, maybe that's the reason?

rileyhgrant (approved):
lgtm!

Apologies for letting this hang for such an atrociously long time, I was just looking at my requested reviews in GitHub and noticed this 😬 

Looks good to me, nicely done! I just had two minor mostly for-my-curiousity comments on this.

Issue #1570, opened 2024-06-14
V4 SV page variant count discrepancy


rileyhgrant:
Originally posted on the Forum: https://discuss.gnomad.broadinstitute.org/t/gnomad-svs-v4-discrepancy-between-hemizygote-male-xy-counts/340

There is a discrepancy on individual SV pages for variants on chrom X in the summing of total number of individuals with the variant vs individuals who are hemi or homozygous with the variant. 

For example
https://gnomad.broadinstitute.org/variant/X-31121763-C-G?dataset=gnomad_r4

The total AC is 19, with 0 homozygous individuals and 11 hemizygous individuals. This does not add up.

![Screenshot 2024-06-14 at 16 13 00](https://github.com/broadinstitute/gnomad-browser/assets/59549713/f2f2fd1a-f57c-4505-9e45-40738ff12cd7)

phildarnowsky-broad:
@rileyhgrant NB that we actually have _two_ separate bugs, that (due to, AFAICT, coincidence) affect hemi counts in short variants and structural variants respectively, and we've conflated the two in this issue. I'll come back in a little bit and separate and clear up the issues in a little bit.

PR #1578, opened 2024-06-20
Fix hemizygote count logic in v4 structural dataset pipeline

phildarnowsky-broad:
Turns out the following are not the same:

```
In [48]: ds.filter(ds.info.PAR is False).count()
Out[48]: 0

In [49]: ds.filter(ds.info.PAR == False).count()
Out[49]: 2143849
```


phildarnowsky-broad:
fixes #1570

rileyhgrant (approved):

PR #1587, opened 2024-06-26
Update loading docs to reflect current practice, and numbers the steps

sjahl:
Note that I think I've encoded a procedural change here. In the old days, when we needed to increase the total available space on the cluster, we used to initiate a blue-green style deployment of the elasticsearch cluster. This involved:

1. spinning up a ingest nodeset and loading data
2. spinning up a new persistent data nodeset (either blue or green)
3. copying all data (new and existing data) to new blue/green
4. shutting down ingest pool
5. shutting down old blue/green

However, since that procedure was put in place, a few things are now the case:

- We can online increase disk sizes to make small adjustments, where we could not before
- The SSD disks we now use are fast enough that data shuffling when adding new nodes to the green/blue nodeSet doesn't cause downtime, so there's less of a requirement to limit the amount of times data moves.

Thus, I don't think the elasticsearch blue/green process is really necessary any more. Lately, the process we've followed has been more like this, which is what is encapsulated in these doc updates:

1. spinning up a ingest nodeset and loading data
2. (if necessary) increase the size of the persistent data nodeset
3. copy data to it
4. shut down old ingest pool

Open to thoughts on this !

### DIFF
--- a/deploy/docs/LoadingLargeDatasets.md
+++ b/deploy/docs/LoadingLargeDatasets.md
@@ -3,60 +3,109 @@
 To speed up loading large datasets into Elasticsearch, spin up many temporary pods and spread indexing across them.
 Then move ES shards from temporary pods onto permanent pods.
 
-- Set [shard allocation filters](https://www.elastic.co/guide/en/elasticsearch/reference/current/shard-allocation-filtering.html)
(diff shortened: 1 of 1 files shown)


phildarnowsky-broad (approved):

rileyhgrant (approved):
lgtm!

PR #1596, opened 2024-07-12
Adds a command line flag for specifying the number of API replicas

sjahl:
As a stop gap until we have a more complete deployctl-less pipeline, this adds a flag to deployctl that will allow developers to specify fewer replicas for demos (and for us to configure more for prod).

Another approach that I may take on this: add a `--demo` flag instead that sets a reasonable amount of resources and replicas, and maybe pins the demo deployments to an autoscaling pool of GKE VMs 🤔


rileyhgrant:
Close in favor of #1634 

This was an alternate strategy, Steve ended up adding the `--demo` flag mentioned here, and we use that.

PR #1602, opened 2024-07-30
Add GTEx and Pext to gnomAD v4.1

rileyhgrant:
Resolves: https://github.com/broadinstitute/gnomad-browser-team/issues/69
Resolves: #1655 

The pext/gtex integration already exists for gnomAD v2 and exac (the two top level datasets on GRCh37). This PR adds this data for the datasets on GRCh38 (gnomAD v4, and v3).

Due to their being different number of tissues, and some that are unique to each dataset, and due to the dislike of the usage of a wide data format for storing and requesting the tissue data, pext/gtex tissues are now stored as an array of structs, rather than a single large struct with a key for each tissue name.

Adds
- pipeline steps to include GTEx and pext based off the GTEx v10 release
  - reworks the schema to use arrays of tissues for a long, rather than wide, table
- graphql changes to account for the new schema
- frontend changes to query and use typing for the new schema


browser/src/GenePage/GeneTranscriptsTrack.tsx
```diff
@@ -129,15 +140,36 @@ const GeneTranscriptsTrack = ({
...
+                  transcript.gtex_tissue_expression.map(
+                    (tissueExpression) => tissueExpression.value
+                  )
+                )!
+                const maxExpression = max(
+                  transcript.gtex_tissue_expression.map(
+                    (tissueExpression) => tissueExpression.value
+                  )
+                )!
+                const tissueMostExpressedIn = transcript.gtex_tissue_expression.find(
```
elissa-alarmani, line 164: Since `.find(...)` might return undefined if no matching item is found it could helpful to consider checking if `transcript` exists before accessing properties

elissa-alarmani (approved):
Overall, this looks great, there are no jarring errors, and the structure of the cold and implementation appears solid and well-organized from the data-pipeline to the api to the frontend. 

That being said, as this is one of my first larger reviews, I'm open to any feedback or suggestions where I might have missed!

PR #1603, opened 2024-07-30
Fix compose browser startup

sjahl:
Looks like some things got lost in translation in the switch to pnpm, which left our compose config in a broken state. Since this is a very easy way to start up a local development browser, I think it's worth maintaining even if it's not getting much usage from local devs.

This adds `bash` to (only) the build stage of the browser image so that we can use the same app start.sh script that we have defined for usage elsewhere.

This set of changes gets me a local browser listening on :8008, and pointed at the public gnomad API, after running:

```bash
./development/env.sh browser build
./development/env.sh browser up
```


rileyhgrant (approved):
lgtm!

Ran the mentioned commands with and without specifying an API to point to, and observed the expected behavior.

PR #1612, opened 2024-08-21
Move elasticsearch to deployments repo

sjahl:
This removes the elasticsearch manifest templating and apply bits from deployctl, and moves them to https://github.com/broadinstitute/gnomad-deployments/tree/1b9bbf5455965b0c8777d160ace3ec88810641f8/elasticsearch

Note, the elasticsearch subcommand of deployctl isn't going away completely, since it still provides the `load-dataset` command that is separate from the kubernetes config side of elasticsearch. There's also a utility function in there for retrieving the ES password, and I believe that's also used somewhere in the data pipeline code.

The template and apply functionality that deployctl was giving us is largely replaced by:

```
kustomize init # to create new cluster configs
kustomize build # to view rendered templates
kustomize build prod | kubectl diff/appy -f - # to diff and apply configs
```

See the documentation in the deployments repo/elasticsearch readme for more info.


rileyhgrant (approved):
lgtm

With the one teeny and inconsequential comment that if you weren't going to squash these commits down, then the second one has a typo in it ('maniests' instead of 'manifests')

Thanks for continually keeping the docs up to date every time you change something!

PR #1623, opened 2024-09-06
Split variant flags

### LINKED ISSUE #1614: Fix variant flag access based on now-deleted `variant.flags` field
A recent bug report regarding VRS revealed that the `flags` field on variants from the API is not being fully populated in all cases, due to the code in question failing to account for how, in the ES variant index, we split `variant.flags` into `variant.exome.flags` and `variant.genome.flags` respectively. There are three known places where this manifests:

- VA: right now the API queries for VA default to using the exome if present, the genome otherwise. instead of doing that, split them up explicitly so that the query has to say whether it wants exome data, genome data, or both. the burden of deciding how to reconcile different data (including flags) between the two data types then also moves to the requester, which is logical anyway since every requester's use case is going to be different and what makes sense for one may not make sense to the other
- Variant table in gene and region page: for `monoallelic`, which can vary exome to genome, make it clear which data type the flag pertains to. could be as simple as displaying them in the table as e.g. "Monoallelic in exome" or "Monoallelic in genome". Also either move LOFTEE flags into their own field, or append to both `exome.flags` and `genome.flags` on the fly, TBD.
- Variant page: We call out variants in a par or lcr here, which should always be the same between exome and genome, so just a simple fix to the logic here so that it's not looking for the nonexistent variant.flags anymore, but rather displays the warning if the appropriate flag is present in either exome or genome or both

phildarnowsky-broad:
This updates the API and frontend to support flags both on a variant generally, or on the exome and/or genome specifically, and fixes a number of related bugs related to display of flags.

fixes #1614

### DIFF
--- a/graphql-api/src/queries/variant-datasets/shared/flags.ts
+++ b/graphql-api/src/queries/variant-datasets/shared/flags.ts
@@ -10,7 +10,7 @@ const isLofOnNonCodingTranscript = (transcriptConsequence: any) =>
   LOF_CONSEQUENCE_TERMS.has(transcriptConsequence.major_consequence) && !transcriptConsequence.lof
 
 const getLofteeFlagsForGeneContext = (variant: any, geneId: any) => {
-  const flags = variant.flags || []
+  const flags = [...(variant.flags || [])]
 
   const allConsequences = variant.transcript_consequences || []
   const consequencesInGene = allConsequences.filter((csq: any) => csq.gene_id === geneId)
@@ -49,7 +49,7 @@ const getLofteeFlagsForGeneContext = (variant: any, geneId: any) => {
 }
 
 const getLofteeFlagsForRegionContext = (variant: any) => {
-  const flags = variant.flags || []
+  const flags = [...(variant.flags || [])]
 
   const allConsequences = variant.transcript_consequences || []
   const lofConsequences = allConsequences.filter((csq: any) => csq.lof)
@@ -86,7 +86,7 @@ const getLofteeFlagsForRegionContext = (variant: any) => {
 }
 
 const getLofteeFlagsForTranscriptContext = (variant: any, transcriptId: any) => {
-  const flags = variant.flags || []
+  const flags = [...(variant.flags || [])]
 
   const allConsequences = variant.transcript_consequences || []
   const consequenceInTranscript = allConsequences.find(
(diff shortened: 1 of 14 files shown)


phildarnowsky-broad:
@rileyhgrant I've decided to put the VA/VRS-related fixes in the other VA topic branch I've been working on, which I'll rebase on top of this, hence this PR doesn't include VA fixes.

graphql-api/src/queries/variant-datasets/shared/flags.ts
```diff
@@ -126,3 +126,22 @@ export const getLofteeFlagsForContext = (context: any) => {
       throw Error(`Invalid context for getFlags: ${context.type}`)
   }
 }
+
+export const getFlagsForContext = (context: any, variant: any) => {
+  const baseVariantFlags = getLofteeFlagsForContext(context)(variant)
+  const baseExomeFlags = variant.exome?.flags || []
+  const baseGenomeFlags = variant.genome?.flags || []
+
+  const regionalFlags = ['par', 'segdup', 'lcr'].filter(
```
rileyhgrant, line 135: Based on the hail tables help text, it appears as though the par flag might be named "non_par"? 

I did a quick grep and didn't see any place where we rename this flag in our pipeline, so just wanted to check.

rileyhgrant (approved):
LGTM, very nicely done, I especially like the care taken to the testing, and to a neatly organized series of commits.

Just one minor comment about the naming of a common region flag (based on the docs, it might be named "non_par", rather than "par", but I wasn't quite sure and wanted to double check).

Nice one!

graphql-api/src/queries/variant-datasets/shared/flags.ts
phildarnowsky-broad: The v4 data pipeline does address this, https://github.com/broadinstitute/gnomad-browser/blob/b804dbeb4b476ba40ce9354c1cd51be88dc1577c/data-pipeline/src/data_pipeline/datasets/gnomad_v4/gnomad_v4_variants.py#L286, but it would be smart for me to double-check that it's correct.

rileyhgrant: Ah whoops, my bad for not finding that, and nice catch!

PR #1624, opened 2024-09-06
Clarify log scaled site quality metrics

### LINKED ISSUE #1615: Note that x-axis is log-scaled for certain site quality metrics
When the methods team creates the variant annotation histogram json files, we [log-transform](https://github.com/broadinstitute/gnomad_qc/blob/ec87a9941211aea3b2f5f38bb48df1c047e240dc/gnomad_qc/v4/create_release/make_var_annot_hists.py#L26C22-L26C63) the following annotations:
```
LOG10_ANNOTATIONS = ["AS_VarDP", "QUALapprox", "AS_QUALapprox"]
"""
List of annotations to log scale when creating histograms.
"""
```

Would it be possible to add `log` to the x-axis label for these metrics to make this clearer on the variant pages? For example, for this [variant](https://gnomad.broadinstitute.org/variant/8-101719383-TAA-T?dataset=gnomad_r4), would it be possible to change the x-axis label to `log(AS_VarDP)`?
image

elissa-alarmani:
Resolves #1615

Regarding the site quality metrics histogram, for the metrics `["AS_VarDP", "QUALapprox", "AS_QUALapprox"]`, here we add the 'log' label to the x-axis and update the displayed value as the calculated log value.


browser/src/VariantPage/VariantSiteQualityMetrics.tsx
```diff
@@ -571,7 +571,10 @@ const yTickFormat = (n: any) => {
   return `${n}`
 }
 
-const formatMetricValue = (value: any, metric: any) => {
+const formatMetricValue = (value: any, metric: any, isLog?: boolean) => {
+  if (isLog) {
+    return `log(${value})`
```
ch-kr, line 576: naive question, is the default log here log10? that is the log we use in production (see slack [thread](https://atgu.slack.com/archives/CRA2TKTV0/p1725041402573139?thread_ts=1725037232.215369&cid=CRA2TKTV0))

browser/src/VariantPage/VariantSiteQualityMetrics.tsx
```diff
@@ -625,11 +628,29 @@ const SiteQualityMetricsHistogram = ({
 }: SiteQualityMetricsHistogramProps) => {
   const isLogScale = metric === 'SiteQuality' || metric === 'AS_QUALapprox' || metric === 'DP'
```
ch-kr, line 629: SiteQuality, AS_QUALapprox, and DP aren't currently log scaled on the variant pages as far as I can tell (https://gnomad.broadinstitute.org/variant/1-55051215-G-GA?dataset=gnomad_r4) -- should they be?
image

browser/src/VariantPage/VariantSiteQualityMetrics.tsx
```diff
@@ -625,11 +628,29 @@ const SiteQualityMetricsHistogram = ({
...
   const secondaryValues = exomeBinValues ? genomeBinValues : null
 
-  const primaryMetricValue = exomeMetricValue !== null ? exomeMetricValue : genomeMetricValue
-  const secondaryMetricValue = exomeMetricValue !== null ? genomeMetricValue : null
+  const originalPrimaryMetricValue =
+    exomeMetricValue !== null ? exomeMetricValue : genomeMetricValue
+
+  const primaryMetricValue =
+    isLogMetrics && originalPrimaryMetricValue > 0
+      ? Math.log(originalPrimaryMetricValue)
```
ch-kr, line 645: same question about the log here

ch-kr (commented):
thank you for adding! I added a couple basic questions.

also quick note: I'm not the best placed to review this code -- perhaps @rileyhgrant or @phildarnowsky could also take a look?

phildarnowsky:
@ch-kr please note I use @phildarnowsky-broad for work code

browser/src/VariantPage/VariantSiteQualityMetrics.tsx
elissa-alarmani: good catch! I updated to use log base 10, by default it was log base e.

browser/src/VariantPage/VariantSiteQualityMetrics.tsx
elissa-alarmani: Just updated that for the labeling, does the graph look correctly scaled?

browser/src/VariantPage/__snapshots__/VariantPage.spec.tsx.snap
```diff
@@ -8061,7 +8061,7 @@ exports[`VariantPage with the dataset "gnomad_r3" has no unexpected changes 1`]
                           dy="0em"
                           x={-35}
                         >
-                          SiteQuality
+                          log(SiteQuality)
```
phildarnowsky-broad, line 8064: looks like `log(whatever)` got interpreted as a string literal rather than getting evaluated and interpolated

phildarnowsky-broad (changes requested):

browser/src/VariantPage/__snapshots__/VariantPage.spec.tsx.snap
elissa-alarmani: If I'm not mistaken I think for the chart labels we might not want to evaluate but display as a log of the metric like "log(SiteQuality)" directly as a label, @ch-kr is that the display we wanted?

browser/src/VariantPage/VariantSiteQualityMetrics.tsx
```diff
@@ -571,7 +571,10 @@ const yTickFormat = (n: any) => {
   return `${n}`
 }
 
-const formatMetricValue = (value: any, metric: any) => {
+const formatMetricValue = (value: any, metric: any, isLog?: boolean) => {
```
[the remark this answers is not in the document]
elissa-alarmani: good catch, `isLog` isn't undefined anywhere so removing the `?`

browser/src/VariantPage/__snapshots__/VariantPage.spec.tsx.snap
phildarnowsky-broad: oh right sorry, wrong context

browser/src/VariantPage/VariantSiteQualityMetrics.tsx
elissa-alarmani: that the graphs are already log scaled and just needed the labeling additions?

ch-kr: ah thanks for clarifying. it seems like the graphs aren't actually log scaled. for that variant ([1-55051215-G-GA](https://gnomad.broadinstitute.org/variant/1-55051215-G-GA?dataset=gnomad_r4)), the browser is showing this for the exomes: 
image
the value for AS_QUALapprox in the v4.1 exomes release HT is 262031, and that value is not log transformed

elissa-alarmani: Screenshot 2024-09-13 at 10 58 16 AM

Using the same variant and view provided this what we have right now after adding the "log()" around the values, would it be better to evaluate or leave it like that with the log labeling?

ch-kr: I'm a little confused -- the plot doesn't seem to be log scaled in that second screenshot. The x-axis in the first screenshot (untransformed plot) has the same range as the second log-transformed screenshot, which shouldn't be the case. 

Also, `log10(262031)` is `5.418`, and it looks like that value is now being labeled at the correct location in the second screenshot with the vertical line and arrow. However, in the untransformed plot, 262031 seems to be near the middle of the distribution, and it should stay at the same spot once the rest of the data are transformed (but it is currently far off to the left).

let me know if this doesn't make sense or if I've completely missed something

ch-kr (approved):
I can't comment on the actual code, but the latest plots shared via slack LGTM!

phildarnowsky-broad (approved):

PR #1629, opened 2024-09-13
Split exome and genome frequencies in VA data

phildarnowsky-broad:
VA data was previously always using exome frequencies if available, genome if not. However, given that this is a general-purpose API, it's hard to say if users will want exomes, genomes, or both for their use cases. Hence, rather than try to make that decision for them, we allow (and require) them now to say which they want in their queries.

### DIFF
--- a/graphql-api/src/graphql/types/va.graphql
+++ b/graphql-api/src/graphql/types/va.graphql
@@ -100,8 +100,7 @@ type VAQualityMeasures {
   heterozygousSkewedAlleleCount: Int
 }
 
-"A measure of the frequency of an Allele in a cohort."
-type VACohortAlleleFrequency {
+type VACohortAlleleFrequencyData {
   id: String!
   type: String!
   label: String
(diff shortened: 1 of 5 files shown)


phildarnowsky-broad:
@rileyhgrant I think the commit message explains my motivation here OK but LMK if you'd like to dig into more detail.

graphql-api/src/graphql/types/va.graphql
```diff
@@ -124,10 +123,11 @@ type VACohortAlleleFrequency {
   This creates a recursive relationship and subcohorts can be further subdivided into more subcohorts.
   This enables, for example, the description of different ancestry groups and sexes among those ancestry groups.
   """
-  subcohortFrequency: [VACohortAlleleFrequency]
+  subcohortFrequency: [VACohortAlleleFrequencyData]
```
phildarnowsky-broad, line 126: note to self: `!` here

rileyhgrant (approved):
lgtm!

PR #1630, opened 2024-09-16
Fix filtered variants in variant table bug

### LINKED ISSUE #1466: v2 showing filtered variants in variant table
### What you did:

I was searching for variants within a region in a gene using the variant table of a gene page and searching with the "c." nomenclature. I had unchecked the "display neighboring variants" box.

I expected to see a list of variants that passed QC and had the exact sequence present in my search.

I got a list of variants with the exact sequence in my search that included "filtered variants" even though I had not checked the box to show "filtered variants". The same thing happens no matter what I use to search in the search box and only happens when using v2, not v4.

image
image

elissa-alarmani:
Resolves #1466

Fixes bug where the "Display neighboring variants" filtering was interferring with the "Filtered variants" filter by being able to receive filtered variants in the search results even when the filter was not checked on.

### DIFF
--- a/browser/src/VariantList/filterVariants.ts
+++ b/browser/src/VariantList/filterVariants.ts
@@ -91,10 +91,8 @@ const filterVariants = (variants: Variant[], filter: VariantFilterState, selecte
 
   filteredVariants = filteredVariants.filter((v: Variant) => v.exome || v.genome)
 
-  if (filter.searchText) {
-    if (!filter.includeContext && filter.includeFilteredVariants) {
-      filteredVariants = getFilteredVariants(filter, variants, selectedColumns)
-    }
+  if (filter.searchText && !filter.includeContext) {
+    filteredVariants = getFilteredVariants(filter, filteredVariants, selectedColumns)
   }
 
   // Indel and Snp filters.
--- a/browser/src/VariantList/filterVariants.ts
+++ b/browser/src/VariantList/filterVariants.ts
@@ -91,8 +91,10 @@ const filterVariants = (variants: Variant[], filter: VariantFilterState, selecte
 
   filteredVariants = filteredVariants.filter((v: Variant) => v.exome || v.genome)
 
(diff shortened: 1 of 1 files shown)


browser/src/VariantList/filterVariants.ts
```diff
@@ -91,8 +91,10 @@ const filterVariants = (variants: Variant[], filter: VariantFilterState, selecte
 
   filteredVariants = filteredVariants.filter((v: Variant) => v.exome || v.genome)
 
-  if (filter.searchText && !filter.includeContext) {
-    filteredVariants = getFilteredVariants(filter, variants, selectedColumns)
+  if (filter.searchText) {
```
rileyhgrant, line 94: Here I think we may need one more branch, to handle the unchecking of "display neighboring variants".

With this change, this re-enables the behavior of allowing the user to decide whether or not to include the filtered variants (via the "Filtered variants" checkbox), but disables the behavior of allowing the user to filter the list to have just the rows that are a partial match (via the "Display neighboring variants" checkbox).

E.g., in the screenshot below, the "display neighboring variants" checkbox is unchecked, so we'd expect the variant list to only include rows that match the search terms

Screenshot 2024-09-17 at 10 24 57.

rileyhgrant (changes requested):
Thanks for tackling this Elissa! I know that chasing down random bugs isn't always the most exciting task, and appreciate your working on them.

I just have one comment on what I believe to be an unintended regression.

No particular priority on this PR, the original ticket is a bit of a niche bug, as its only for v2, so if you have other things on your plate those can definitely take precedence!

rileyhgrant (approved):
Nice, works as intended and LGTM!

As an aside, something like this is a good opportunity to add a test to catch any future regressions (set up a test object and test the corresponding filter functions separately and in conjunction).

Honestly though, in this case the original bug was niche enough, and this feature is so infrequently touched compared to other features, that I think the time would be better spent on other things, rather than testing this.

PR #1633, opened 2024-09-24
Add some API documentation for endusers

phildarnowsky-broad:
Here we start adding docs covering the basics of using our GraphQL API, including how to fetch all variant frequencies for a gene.

Note the original brief was to replicate the data exported to CSV off the gene page variant table, but I wasn't able to do that exactly, since it turns out some of that data is calculated in the frontend code rather than the API.


browser/help/faq/technical-details/how-do-i-query-a-batch-of-variants-do-you-have-an-api.md
```diff
@@ -2,6 +2,6 @@
 question: 'How do I query a batch of variants? Do you have an API?'
 ---
 
-We currently do not have a way to submit batch queries on the browser, but we are actively working on developing a public API for gnomAD. We currently maintain a [GraphQL](https://graphql.org) API for internal use and groups can contact us if they want permission to use this API until a more robust public solution is launched.
+The [GraphQL](https://graphql.org) API used by the browser is open to the public. Please see our [data page](/data) for details.
```
ch-kr, line 5:
```suggestion
The [GraphQL](https://graphql.org) API used by the browser is open to the public. Please see our [data page](/data) for more details on how to use our API.
```

browser/src/DataPage/exampleRawGraphQLQuery.ts
```diff
@@ -0,0 +1,126 @@
...
+#
+query VariantsInGene {
+  # We could alternatively look up the gene by Ensembl ID by using the
+  # "gene_id" argument rather than "gene_symbol".
+  #
+  # The gene symbol is given with quotes, but not the reference genome or
+  # dataset. This is because the gene symbol is defined in the API schema as a
+  # string, but the reference genome is an enumerated type that can have only
+  # the value GRCh37 or GRCh38, and the dataset argument is constrained
+  # similarly. See
```
ch-kr, line 14: minor rewording:
In this query, the gene symbol must be provided with quotes ("BRCA1"). This does not apply to the reference genome build or gnomAD dataset version because the gene symbol is defined as a string, but the reference genome and dataset version are defined as enumerated types that only have specific values (e.g., reference genome can only be one of GRCh37 or GRCh38).

open to additional rewordings on top of this as well!

browser/src/DataPage/exampleRawGraphQLQuery.ts
```diff
@@ -0,0 +1,126 @@
...
+      hgvsp
+
+      # VEP annotation
+      consequence
+
+      # Flags on the variant generally
+      flags
+
+      # The API separates the frequency data into exome data, genome data, and
+      # joint-frequency data. Which of these are available depends on the
```
ch-kr, line 42:
```suggestion
      # joint frequency data. Which of these are available depends on the
```
I don't think we need a hyphen here but optional

browser/src/DataPage/exampleRawGraphQLQuery.ts
```diff
@@ -0,0 +1,126 @@
...
+      # dataset used, and not all variants in a dataset necessarily have all
+      # of the possible kinds.
+      #
+      # Note that the exome and genome data have the same schema, but joint-
```
ch-kr, line 46:
```suggestion
      # Note that the exome and genome data have the same schema, but joint
```

ch-kr (commented):
thank you for adding this, Phil! I have some minor rewording suggestions, but the documentation looks really nice.

p.s. I didn't review the code 😆 but I assume this was expected

phildarnowsky-broad:
@rileyhgrant this version incorporates all the feedback I got on the demo

browser/src/Routes.tsx
```diff
@@ -169,7 +169,9 @@ const Routes = () => {
 
       <Route exact path="/team" component={TeamPage} />
 
-      <Route exact path="/downloads" component={DownloadsPage} />
+      <Redirect from="/downloads" to="/data" />
```
rileyhgrant, line 172: I like the redirect to catch old links, however, it appears as though this doesn't allow for old links with a hash to properly link to the new page. e.g. `.../downloads#v4` will just redirect to `.../data`

Not sure if there's an easy way around this. There are a fair number of blog posts that have linked to a specific a tag on the downloads page for ease of user navigation, so just wanted to bring this up.

rileyhgrant (commented):
Docs look good!

I just have one comment about the redirects, as the current way does not include the fragments, and many blog posts links to the current downloads page include fragments, e.g. https://gnomad.broadinstitute.org/downloads#v4-structural-variants

browser/src/Routes.tsx
phildarnowsky-broad: @rileyhgrant nice catch--I am simply going to route both `/downloads` and `/data` to this page.

rileyhgrant (approved):

PR #1647, opened 2024-10-29
Fix single source exome/genome values in variant table

### LINKED ISSUE #1639: ACs and ANs don't update in gene page variant table when switching exomes/genomes off and on
For variants with both exome and genome data, when selecting/deselecting exomes or genomes on the gene page, the AC and AN figures in the variant table don't update, but rather continue to show the combined totals for exome and genome.

rileyhgrant:
Resolves #1639 

When the `joint` frequency data was added as a 3rd source of truth for frequency information, the function that merges exome and genome data was modified to preferentially use the `joint` values if they exist. This caused the table to not update AC and AN values when a user de-selected either the exomes or genomes.

This PR passes an additional prop to the merging function to have it only preferentially use the `joint` data when both the exome and genome data are selected by the user. This way, when a source is de-selected, the function falls back to using only the data source selected.

### DIFF
--- a/browser/src/VariantList/mergeExomeAndGenomeData.ts
+++ b/browser/src/VariantList/mergeExomeAndGenomeData.ts
@@ -125,14 +125,16 @@ type MergedVariant = Variant & {
 export const mergeExomeAndGenomeData = ({
   datasetId,
   variants,
+  preferJointData = false,
 }: {
   datasetId?: DatasetId
   variants: Variant[]
+  preferJointData?: boolean
 }): MergedVariant[] => {
   const mergedVariants = variants.map((variant: Variant) => {
     const { exome, genome, joint } = variant
 
-    if (joint) {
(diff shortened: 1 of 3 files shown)


browser/src/VariantList/mergeExomeAndGenomeData.ts
```diff
@@ -125,14 +125,16 @@ type MergedVariant = Variant & {
 export const mergeExomeAndGenomeData = ({
   datasetId,
   variants,
+  preferJointData = true,
 }: {
   datasetId?: DatasetId
   variants: Variant[]
+  preferJointData?: boolean
```
phildarnowsky-broad, line 132: since you have the default value there you should be able to leave the `?` off

phildarnowsky-broad (approved):
one minor change noted, good to go once that is done

PR #1652, opened 2024-11-05
Update ClinVar pipeline for new XML format

### LINKED ISSUE #1569: Update ClinVar pipeline

rileyhgrant:
Resolves #1569 

- Updates the ClinVar pipeline's XML parsing step to parse the new ClinVar XML format
- Minor edits to accomodate ClinVar updates
	- On the frontend, renames ClinVar's 'Clinical significance" to "germline classification", following ClinVar's change in terminology
	- Adds entries to the Gold Star mapping dictionary to accomodate new terminology
- Bumps Hail to 0.2.128 to allow for VEPing of GRCh37 varints using Hail's built in convenience function
- Still includes the commented out Aliases for ClinVar indices along with a "TODO:" comment. Following the practice of several previous PRs, temporarily specifying an index, rather than an alias, in this PR will allow for an easy rollback if needed. Once this proves stable in production, another PR will go back to using the aliases
 
---

Screenshots of a local browser and API querying an indice in the production ES cluster that uses data from the output of this PR:

- Screenshot 2024-11-08 at 16 00 52

- Screenshot 2024-11-08 at 16 01 06

- Screenshot 2024-11-08 at 16 01 09

- Screenshot 2024-11-08 at 16 01 19

- Screenshot 2024-11-08 at 16 02 31

- Screenshot 2024-11-08 at 16 02 37

### DIFF
--- a/data-pipeline/src/data_pipeline/datasets/clinvar.py
+++ b/data-pipeline/src/data_pipeline/datasets/clinvar.py
@@ -162,25 +162,22 @@ def _parse_variant(variant_element):
     return variant
 
 
-def import_clinvar_xml(clinvar_xml_path):
-    release_date = None
+def parse_clinvar_xml_to_tsv(
+    input_xml_path,
+    output_tsv_path,
+    parse_variant_function,
+):
+    release_date = ""
 
-    clinvar_xml_local_path = os.path.join("/tmp", os.path.basename(clinvar_xml_path))
-    print("Copying ClinVar XML")
-    if not os.path.exists(clinvar_xml_local_path):
-        subprocess.check_call(["gsutil", "cp", clinvar_xml_path, clinvar_xml_local_path])
-
-    print("Parsing XML file")
-    with open("/tmp/clinvar_variants.tsv", "w", newline="") as output_file:
+    with open(output_tsv_path, "w", newline="") as output_file:
         writer = csv.writer(output_file, delimiter="\t", quotechar="", quoting=csv.QUOTE_NONE)
         writer.writerow(["locus_GRCh37", "alleles_GRCh37", "locus_GRCh38", "alleles_GRCh38", "variant"])
 
-        open_file = gzip.open if clinvar_xml_local_path.endswith(".gz") else open
-        with open_file(clinvar_xml_local_path, "r") as xml_file:
+        open_function = gzip.open if str(input_xml_path).endswith(".gz") else open
+        with open_function(input_xml_path, "r") as xml_file:
             # The exact number of variants in the XML file is unknown.
             # Approximate it to show a progress bar.
             progress = tqdm(total=1_100_000, mininterval=5)
-
             xml = ElementTree.iterparse(xml_file, events=["end"])
             for _, element in xml:
                 if element.tag == "ClinVarVariationRelease":
@@ -188,8 +185,7 @@ def import_clinvar_xml(clinvar_xml_path):
 
                 if element.tag == "VariationArchive":
                     try:
-                        variant = _parse_variant(element)
-
+                        variant = parse_variant_function(element)
                         locations = variant.pop("locations")
                         writer.writerow(
                             [
@@ -204,7 +200,6 @@ def import_clinvar_xml(clinvar_xml_path):
                                 json.dumps(variant),
                             ]
                         )
-
                         progress.update(1)
 
                     except SkipVariant:
@@ -219,7 +214,24 @@ def import_clinvar_xml(clinvar_xml_path):
                     # https://stackoverflow.com/questions/7697710/python-running-out-of-memory-parsing-xml-using-celementtree-iterparse
                     element.clear()
 
-            progress.close()
+                progress.close()
+
+            return release_date
+
+
+def import_clinvar_xml(clinvar_xml_path):
+    release_date = None
+
+    clinvar_xml_local_path = os.path.join("/tmp", os.path.basename(clinvar_xml_path))
+    print("Copying ClinVar XML")
+    if not os.path.exists(clinvar_xml_local_path):
+        subprocess.check_call(["gsutil", "cp", clinvar_xml_path, clinvar_xml_local_path])
+
+    print("Parsing XML file")
+    output_file = "/tmp/clinvar_variants.tsv"
+    release_date = parse_clinvar_xml_to_tsv(
+        input_xml_path=clinvar_xml_local_path, output_tsv_path=output_file, parse_variant_function=_parse_variant
+    )
 
     subprocess.check_call(["hdfs", "dfs", "-cp", "-f", "file:///tmp/clinvar_variants.tsv", "/tmp/clinvar_variants.tsv"])
 
--- a/data-pipeline/src/data_pipeline/datasets/clinvar.py
+++ b/data-pipeline/src/data_pipeline/datasets/clinvar.py
@@ -1,11 +1,11 @@
 import csv
-import datetime
 import gzip
 import json
 import os
 import subprocess
 import sys
 from xml.etree import ElementTree
+from xml.sax.saxutils import quoteattr
 
 import hail as hl
 from tqdm import tqdm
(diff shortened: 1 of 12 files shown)


data-pipeline/src/data_pipeline/datasets/clinvar.py
```diff
@@ -162,34 +162,30 @@ def _parse_variant(variant_element):
...
 
 
-def import_clinvar_xml(clinvar_xml_path):
-    release_date = None
-
-    clinvar_xml_local_path = os.path.join("/tmp", os.path.basename(clinvar_xml_path))
-    print("Copying ClinVar XML")
-    if not os.path.exists(clinvar_xml_local_path):
-        subprocess.check_call(["gsutil", "cp", clinvar_xml_path, clinvar_xml_local_path])
+def parse_clinvar_xml_to_tsv(
```
phildarnowsky-broad, line 171: nice little refactor here, also a good demonstration of one reason significant indentation is a bad idea: it makes it much more likely to introduce an error in this kind of refactoring by moving a block of code but neglecting to adjust indentation

data-pipeline/tests/pipeline/test_new_clinvar_xml_format.py
```diff
@@ -0,0 +1,293 @@
+import gzip
```
phildarnowsky-broad: I notice that we introduce this file here and then delete it a few commits down the line. Generally speaking it's best not to introduce something into a PR and then delete it later, but rather to rebase the branch before opening the PR so that the stuff in question just never appears in the PR in the first place.

Now in this case, I think we may actually want to save this script, since presumably we're going to want to test this process again at some point in the future, but I'll leave it to your judgment about how likely that is. If we do save it, it would be good to get rid of the `_NEW_foo` function names and other rough edges.

data-pipeline/src/data_pipeline/datasets/clinvar.py
```diff
@@ -41,21 +40,26 @@ class SkipVariant(Exception):
     pass
 
 
+def escape_xpath_value(value):
```
phildarnowsky-broad, line 43: kind of surprised there's not a way in the library to do this, seems like it would come up

data-pipeline/src/data_pipeline/datasets/clinvar.py
```diff
@@ -65,61 +69,71 @@ def _parse_submission(submission_element, trait_mapping_list_element):
 
         if trait_mapping_list_element is not None:
             xref_elements = trait_element.findall("XRef")
+
```
phildarnowsky-broad, line 72: I like the desire to introduce some explanation in here. One thing, though, is that when you end up writing a series of comments that walk you through a process step by step, it's a sign to consider refactoring by extracting helper functions to make the code more self-documenting.

Here I would specifically recommend extracting in two stages: in one, you break this whole big chonk of code out into a function called something like `find_mapping_elements` which takes `trait_element` as input and spits out the corresponding `preferred_name_element` and `mapping_element`. 

Then I'd extract each of the individual steps out of `find_mapping_elements` to something like:

```
def find_mapping_elements(trait_element, xref_elements):
  (preferred_name_element, mapping_element) = find_mapping_elements_by_explicit_xref(trait_element, xref_elements) ||
      find_mapping_elements_by_preferred_name(trait_element) ||
      find_mapping_elements_by_classification_name(trait_element)
```

And looking at it, everything from line 103 on is concerned with adding Medgen IDs rather than finding these mappings, so let's extract that stuff into its own set of functions.

data-pipeline/src/data_pipeline/datasets/clinvar.py
```diff
@@ -65,61 +69,71 @@ def _parse_submission(submission_element, trait_mapping_list_element):
...
     return submission
 
 
-def _parse_variant(variant_element):
+def _parse_variant(variant_element, tqdm_pbar=None):
     variant = {}
 
-    if variant_element.find("./InterpretedRecord") is None:
+    if variant_element.find("./ClassifiedRecord") is None:
         raise SkipVariant
```
phildarnowsky-broad, line 121: So, recognizing that you didn't introduce this one, you generally don't want to use exceptions for control flow that can be done, as it were, through normal channels. Here rather than raise an exception in the handful of places it happens, I would `return null`, thus bailing out of the function early, and adjust the logic that calls `_parse_variant` to handle `null` results.

phildarnowsky-broad (changes requested):
Good stuff overall, I did spot some things, mainly legacy stuff that we should take this opportunity to clean up.

data-pipeline/src/data_pipeline/datasets/clinvar.py
```diff
@@ -47,6 +47,77 @@ def escape_xpath_value(value):
...
+
+    if mapping_element is None:
+        mapping_element, preferred_name_element = find_mapping_elements_by_name(
+            trait_element, submission_element, trait_mapping_list_element
+        )
+
+    return (mapping_element, preferred_name_element)
+
+
+def _associate_condition_with_medgen_id(submission_element, trait_mapping_list_element, trait_element):
```
phildarnowsky-broad, line 97: Nice one, you see how both the overall process and the individual parts are both easier to understand now.

PR #1653, opened 2024-11-05
Update variant page grp frequency table styling

rileyhgrant:
The Variant page's genetic ancestry group frequency table recently had its styling updated to make the division between genetic ancestry groups more clear, as the subtler grey dividing line between the total freq information and the sex specific freq information in a given ancestry group was thicker than the lines between the groups.

This change unintentionally made the total sex specific (XX/XY) frequencies dividers become less clear. This PR slightly modifies the styling to make the header and footer more distinct than individual rows, and make clear the XX/XY section of the table.

It keeps the changes to thicknesses of lines that were requested and modified in a previous PR.

- Production (current) stylings on left, PR stylings on right
	Screenshot 2024-11-05 at 16 13 10

The main change is to make the line dividing XX/XY less distinct than the lines above and below it

i.e. modify stylings here

Production:
	Screenshot 2024-11-05 at 16 16 23

PR:
	Screenshot 2024-11-05 at 16 15 55

### DIFF
--- a/browser/src/VariantPage/PopulationsTable.tsx
+++ b/browser/src/VariantPage/PopulationsTable.tsx
@@ -7,10 +7,24 @@ import { logButtonClick } from '../analytics'
 const Table = styled(BaseTable)`
   min-width: 100%;
 
+  tr.strong-border {
+    td,
+    th {
+      border-top: 2px solid #333;
+    }
+  }
+
   tr.border {
     td,
     th {
-      border-top: 2px solid #aaa;
+      border-top: 2px solid #888;
+    }
+  }
+
+  tr.subtle-border {
+    td,
+    th {
+      border-top: 2px solid #bbb;
     }
   }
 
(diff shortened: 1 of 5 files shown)


browser/src/VariantPage/PopulationsTable.tsx
```diff
@@ -66,26 +76,29 @@ const renderPopAF = (af: number) => {
...
-      name: string
-      ac: number
-      an: number
-      ac_hemi?: number
-      ac_hom?: number
-    }[]
   }[]
+}
+
+type OwnPopulationsTableProps = {
```
phildarnowsky-broad, line 95: while we're in here let's get rid of the `Own` here and call this `PopulationTableProps` or even just `Props` if it's not ambiguous. IIRC the `Own` that crops up in lots of names of types of component properties is an artifact of how the automatic ts-migrate script converted Proptypes, neither of which is relevant anymore.

phildarnowsky-broad (approved):
looking good just one very small touchup noted, good to merge once you've done that

PR #1658, opened 2024-11-21
Sort ClinVar consequence in table by pathogenicity

### LINKED ISSUE #1656: Sorting by clinical significance is alphabetical, could it be by severity/level?
At the moment, when you sort the variants table by clinical significance, the sort is alphabetically by label (such that Descending order results in Uncertain Significance being sorted at the top).

It would be nice if sorting was done by the level of significance. Do we map these levels to ints in the database in a way that the sorting could be done that way, rather than alphabetically by its label? This might also help with sorting consistency moving forward if the labels change.

Screenshot 2024-11-21 at 1 03 55 PM

rileyhgrant:
Resolves #1656 

When sorting the variant table by ClinVa, sort it by Pathogenicity rather than lexicographically.

Right now it sorts to more pathogenic first on first click, as I suspect that's what users are more likely to want to see, that's very easy to change if needed.

---

Also currently includes a commit to rename 'clinical significance' to 'germline classification' in the relevant locations. ClinVar moved to [several different](https://www.ncbi.nlm.nih.gov/clinvar/docs/clinsig/#:~:text=Prior%20to%202024%2C%20ClinVar%20had%20a%20single,somatic%20classification%20of%20clinical%20impact%2C%20and%20oncogenicity.) classifications, and the germline classification one is the one that we have always displayed as the "clinical significance".

Sorted descending:
Screenshot 2024-11-21 at 14 01 20

Sorted ascending
Screenshot 2024-11-21 at 14 01 23

Unsorted:
Screenshot 2024-11-21 at 13 58 47

### DIFF
--- a/browser/help/topics/clinvar-variant-categories.md
+++ b/browser/help/topics/clinvar-variant-categories.md
@@ -3,27 +3,27 @@ id: clinvar-variant-categories
 title: ClinVar variant categories
 ---
 
-ClinVar variants shown in this track are divided into four categories based on clinical significance.
+ClinVar variants shown in this track are divided into four categories based on germline classification, formerly called clinical significance.
 
-* Pathogenic / likely pathogenic
-  * Pathogenic
-  * Likely pathogenic
-  * Pathogenic/Likely pathogenic
-  * association
-  * risk factor
-* Uncertain significance / conflicting
-  * Uncertain significance
-  * Conflicting interpretations of pathogenicity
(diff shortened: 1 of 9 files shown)


sjahl (approved):
Thanks Riley, I'll let Phil chime in on any Typescript or other design things, but the approach of building a mapping of strings to numbers and sorting that way seems reasonable!

PR #1667, opened 2025-01-06
Filter gene variant queries to exons consistently

rileyhgrant:
Currently, our frontend determines what to display in the region viewer based partially on the presence of CDS exons: https://github.com/broadinstitute/gnomad-browser/blob/c7276f25d160059429f039ad0b7e0f445fcdc491/browser/src/GenePage/GenePage.tsx#L315-L353

In the case of non-coding genes, this results in the Coverage track and Frequency-Bubble track filtering variants on non-coding genes in the case where there are coding transcripts, but the variant table including variants from those non-coding transcripts. Since one of these filters occurs on the frontend only for the region viewer, we are displaying inconsistent numbers of variants between the track and the table, and between which component on the frontend a user looks at vs the API response.

---

This PR updates our API to filter the variant response for non coding genes to only include variants in the padded exons regions, to match the frontend region viewer logic. This should increase consistency in how we display different views.

### DIFF
--- a/graphql-api/src/queries/clinvar-variant-queries.ts
+++ b/graphql-api/src/queries/clinvar-variant-queries.ts
@@ -7,6 +7,7 @@ import { fetchAllSearchResults, fetchIndexMetadata } from './helpers/elasticsear
 import { mergeOverlappingRegions } from './helpers/region-helpers'
 import { getConsequenceForContext } from './variant-datasets/shared/transcriptConsequence'
 import largeGenes from './helpers/large-genes'
+import { getFilteredRegions } from './variant-datasets/gnomad-v4-variant-queries'
 
 const CLINVAR_VARIANT_INDICES = {
   GRCh37: 'clinvar_grch37_variants',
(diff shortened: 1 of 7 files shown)


rileyhgrant:
Context from this Slack Thread (Circa January 2025): https://the-tgg.slack.com/archives/C03P7FA3W3T/p1736196651991669?thread_ts=1730231253.938389&cid=C03P7FA3W3T

> ... not as familiar with how we query and display non coding things. The 150 variants that are currently being displayed are all the variants that match WASIR1's gene_id. I think I’m missing context on how we’ve wanted to display variants in non-coding genes, the Gene page keeps its exon centric view for the coverage track, gene frequency bubble track etc, but falls back to just querying everything with the gene id, leading to a kind of half and half approach between querying everything and being exon centric. I suspect we’ll likely want to still be exon centric and filter non coding gene’s variant response down to their padded exon regions.

> Ah right, the consensus in standup was that either way is probably fine, so with that we are indeed going to change the non coding display to be exon centric, so the 197 variants for WASIR1 in exons including filtered variants ...

The reason for more variants being returned from the API when it's non-coding is due to the `should` clause in the elasticsearch search. When there are no `CDS` exons (e.g. a non-coding transcript) then the should clause for the regions effectively drops off, as there are no regions, thus it just queries based on `gene_id`.

phildarnowsky-broad (changes requested):
Nice catch here. The code looks good but let's add at least some unit tests for the basic cases on `getFilteredRegions` itself.

phildarnowsky-broad (approved):

PR #1682, opened 2025-02-10
Add useTableSort hook

phildarnowsky-broad:
This takes some rendering and sorting logic that originated in the mitochondrial variant population frequency table and extracts it into a hook we can re-use in other functional components, such as the upcoming revisions to the short tandem repeats table.


rileyhgrant (approved):
LGTM! 

Nice one, hopefully this can get reused in those Class Components you call out if/when they get reworked.

PR #1708, opened 2025-04-03
Integrate ts region viewer

phildarnowsky-broad:


phildarnowsky-broad:
@rileyhgrant 
> Woohoo! Pretty sweet to see so many of those ts expect errors drop away.

IKR, most fun I've had in a while.

PR #1711, opened 2025-04-11
Updates to STR pages

phildarnowsky-broad:
This begins with some refactoring, and goes on to make two user-facing changes:

* The min/max repeats fields for reads are now text inputs that only reload the reads when they lose focus, or when the user hits "Enter".
* When using the "color by" option on the allele size distribution plot, the order of each individual stack follows the order in the legend.


rileyhgrant (approved):
lgtm!

PR #1729, opened 2025-05-13
Use new GBTK packages

phildarnowsky-broad:


browser/src/App.tsx
```diff
@@ -72,7 +72,7 @@ const Banner = styled.div`
 
 const BANNER_CONTENT = (
   <>
-    Help us continue to improve gnomAD by taking 5 minutes to fill out our {/* @ts-expect-error */}
+    Help us continue to improve gnomAD by taking 5 minutes to fill out our
     <ExternalLink href="http://broad.io/2024_survey">user survey</ExternalLink>.
```
rileyhgrant, line 76: Never have I been so excited to see a particular flavor of `@ts-expect-error` to be removed systematically. Nice!!!

rileyhgrant (approved):
LGTM!

In addition to going commit by commit, I started a local frontend and poked around, all seemed stable.

Huzzah to no more `@ts-expect-error` on every `ExternalLink` use!!

browser/src/App.tsx
phildarnowsky-broad: Yeah this was about 1 `@ts-expect-error` in 8 deleted in one commit, just over a thousand more to go.

PR #1743, opened 2025-06-12
Add OpenTelemetry gateway to deployment and instrumentation to API

phildarnowsky-broad:
This adds an OpenTelemetry gateway per browser deployment, set up to pass metrics and traces to GCP's monitoring tools. It also adds the basic setup to instrument the API, which includes automatic traces on certain of the libraries that we use that supports that, such as graphql and redis.


rileyhgrant (approved):
lgtm!

PR #1755, opened 2025-07-02
Add more links to TR pages

bw2:
See https://atgu.slack.com/archives/CNLMF0JLV/p1751463911325989


phildarnowsky-broad (changes requested):
The code looks OK but CI is now failing due to content changes, you'll need to re-record the snapshots with `pnpm jest -u` from the top-level directory.

Also as a rule it's best not to use a link to Slack as the only description of the change, as this is a public repo and some readers won't have access to our Slack.

phildarnowsky-broad (approved):

PR #1805, opened 2025-10-17
Build and use patched gene, variant, and transcript data for RNU4ATAC

phildarnowsky-broad:
This adds indices and API changes to allow us to correct the incorrect data related to gene RNU4ATAC without having to rebuild every index. I've tried to write the API changes in terms of helpers that we can re-use in the future, on the theory that this is probably not the last error in the data we'll ever find.


rileyhgrant (approved):

rileyhgrant:
We did not end up going this route, rather we waited for gnomAD v4.1.1. and used the patched data (with VEP115) from that input data.

PR #1823, opened 2025-11-21
Allow certain constraint fields to be null

phildarnowsky-broad:
Logs showed errors caused by missing GnomadConstraint.exp_mis and ExacConstraint.syn_z. Checking against the original Hail tables, confirmed that there are genes with missing values for these. Adjusted GraphQL schema and frontend code to allow for this.


rileyhgrant (approved):
lgtm!

PR #1885, opened 2026-06-08
Update Reads NGINX config to allow for HTTP/2

nadeaujoshua:
This PR is part of work for https://github.com/broadinstitute/gnomad-browser-team/issues/116.

We have an NGINX container for the Reads server, and it currently uses HTTP/1.1 by default. This config causes a problem for running Reads in GCP Cloud Run, because [Cloud Run service payloads are capped at 32MiB](https://docs.cloud.google.com/run/quotas#request_limits). Reads responses can be much larger than the cap.

Seeing that HTTP/2 request sizes are not limited in the docs, we hand-tested running Reads on Cloud Run with HTTP/2, and confirmed that responses are also not subject to the size cap.

This PR updates the Reads NGINX container config to allow using HTTP/2, so that we can then deploy the Reads service to prod using Cloud Run, making our new K8s-less infrastructure simpler and easier to manage.

Setting the ENABLE_HTTP2 env var at container run time will set the HTTP/2 config in NGINX to `on`. Using this mechanism keeps everything backwards-compatible in case other Browser deployments are not using HTTP/2.


rileyhgrant (approved):
lgtm!

Issue #698, opened 2021-02-01
Correlate information between tracks


nawatts:
When hovering over a tracks, show a line to make it easy to match up information between tracks. Preferably, show a marker next to the line with information about features in that track at the hovered location. 

![Screen Shot 2021-02-01 at 4 17 28 PM](https://user-images.githubusercontent.com/1156625/106523262-db398880-64ae-11eb-8ebb-1c6698a631d2.png)

nawatts:
Have a way to lock this in place to avoid movement while scrolling? Or better yet, be able to add/remove multiple markers.

nawatts:
Allow setting a marker from a button in a variant's table row.

PR #1891, opened 2026-06-16
Add cross track hover line

ch-kr:
Adds basic functionality suggested in https://github.com/broadinstitute/gnomad-browser/issues/698. Does not currently cover these two points:
- Have a way to lock this in place to avoid movement while scrolling? Or better yet, be able to add/remove multiple markers.
- Allow setting a marker from a button in a variant's table row.


browser/src/GenePage/GenePage.tsx
```diff
@@ -530,112 +637,43 @@ const GenePage = ({ datasetId, gene, geneId }: Props) => {
...
               includeNonCodingTranscripts={includeNonCodingTranscripts}
               includeUTRs={includeUTRs}
-              preferredTranscriptId={preferredTranscriptId}
-              preferredTranscriptDescription={preferredTranscriptDescription}
+              zoomRegion={zoomRegion}
+              hasOnlyNonCodingTranscripts={!hasCodingExons && hasNonCodingTranscripts}
+              cursorClick={cursorClick}
+              // This is a render prop that wraps tracks, not a component type, so the
+              // "unstable nested component" concern about remounting does not apply.
+              // eslint-disable-next-line react/no-unstable-nested-components
```
mattsolo1: The trackWrapper arrow function on line 655 creates a new <SynchronizedCursor> each render, which could cause unmount/remount. You could wrap it in useCallback to keep a stable reference and drop the eslint-disable.

mattsolo1 (commented):
Looks great! I tested this out locally and found it to behave as expected (e.g. popping out various tracks like pext, transcripts).

In the future it's useful to include some screenshots in the, PR for user-facing features/changes for people to have an idea of the work, e.g:

Image

A few notes:

- Please squash the commit history before merge. A lot of the intermediate commits ("claude fix for test error", "add missing newline", "removing extra files") are noisy. You can do a rebase/squash into more descriptive, sensible commits & force push to this branch.
- An AI plan file was committed then removed — issue-698-plan.md was added in the first commit and deleted. The squash would also get rid of this from history.
- Don't mention "Claude, etc" so much in the commit messages. We just want descriptions of what changed. In fact, Claude can write good commit messages so you can ask it to do it if you're not sure how to organize the rebase or what the commit messages should be.
- Cross track hover wasn't added to SV/CNV/mito pages (not sure if that was intended)
- Comments could be trimmed down, AI tends to go overboard.

I'll wait for others on the team to review, since this is our first AI coded PR.

klaricch:
Looks good and I like the suggestion for an option to lock it in place or add a marker. 

One suggestion: the label background for the genomic position text above the cross-track cursor line is #fff, so it looks more white than the #fafafa page background — could match the background color so it blends in.

One visual question: would we want the line to go behind the ClinVar category toggles rather than passing through them? see screenshot 
line_edit

ch-kr:
Thanks for the reviews @mattsolo1 and @klaricch!

Based on guidance from @rileyhgrant, I've rebased this branch off the `standardize-agent-contributions` branch (see [PR](https://github.com/broadinstitute/gnomad-browser/pull/1903)) and asked Claude to clean up this branch. This should have addressed Matt's comments about streamlining commits.

For the other points:
> Cross track hover wasn't added to SV/CNV/mito pages (not sure if that was intended)

I checked with Moriel about the value add for these other pages, and she suggested that this would be useful to add to the mito pages but less useful to add to the SV/CNV pages. I've now extended this feature to the mito gene pages (e.g., MT-ATP6):
image


> One suggestion: the label background for the genomic position text above the cross-track cursor line is #fff, so it looks more white than the #fafafa page background — could match the background color so it blends in.

Thank you! I meant to fix this earlier and missed it. The line now looks like this (PCSK9):
image
This also addresses this point:
> One visual question: would we want the line to go behind the ClinVar category toggles rather than passing through them? see screenshot

The line now goes behind the "gnomAD variants" and "ClinVar variants" headers as well as text and other buttons in those tracks. This change also applies to the pext and transcript tracks:
image
though the line does pass through the text entry field for "Filter tissues" in the pext track; this is because the hiding the line behind that field made it look like the line was broken/interrupted for no reason
image

Also related to the screenshots above, I added two additional suggestions from Moriel:
- The line color is now a lighter gray
- There is a new toggle to turn on this functionality

Toggle off (line only exists in gnomAD variants track, consistent with production browser:
Demo:
image
[Prod](https://gnomad.broadinstitute.org/gene/ENSG00000169174):
image

Toggle on:
image

> I like the suggestion for an option to lock it in place or add a marker.

I added a lot of changes to polish up the original request (adding this cross track hover line and having it hide behind certain display elements), so I wasn't sure if this would be too difficult to review if I implemented it now. I can add it now if that is preferred; just let me know.

One additional question: this new cross track hover line is currently only implemented on gene pages. Is this something I should extend to transcript pages, and if yes, should that happen in this PR?

browser/src/GenePage/GenePage.tsx
ch-kr: I asked Claude about this:
> Done — wrapped it in useCallback and dropped the eslint-disable. One note: the remount risk doesn't actually apply here, since trackWrapper is invoked rather than used as a component type ([Variants.tsx:408](vscode-webview://0c924ss7s684cn4nfvvei22cfk4c0qp8d09is4p7o4ct01lk8j7h/browser/src/VariantList/Variants.tsx#L408)), so React reconciles by the stable SynchronizedCursor type. But useCallback is a clean way to remove the lint suppression and keep a stable reference, so happy to make the change.

Issue #1679, opened 2025-02-05
Update ctrl+f functionality in gnomAD browser


ch-kr:
A few users in the 2024 user survey provided feedback that it would be helpful if `ctrl+f` would search all of the variants in a gene page (e.g., [PCSK9](https://gnomad.broadinstitute.org/gene/ENSG00000169174)) rather than just the variants displayed in the table at any given moment. Konrad pointed out that you can update the behavior of `ctrl+f` to search the contents of your webpage and provided the example that `ctrl+f` in Google docs searches all of the text in that doc rather than the just the bar at the top ("File", "Edit", "View", etc.); it would be great if we could update the behavior of `ctrl+f` on the gnomAD gene pages as well.

ch-kr:
reviving this request from the user survey -- would this be a lot of effort to implement?

PR #1892, opened 2026-06-17
Add custom Ctrl/⌘+F find bar

mike-w-wilson:
Replaces the browser's native find (Ctrl/⌘+F) with an in-app find bar that
searches the whole page, including every variant in the virtualized table —
not just the rows currently rendered.

- Searches page text (CSS Custom Highlight API) and the full variant list
  (reusing the table's client-side search).
- Results are one top-down sequence — page text, variant matches, page text —
  shown as "i of N"; Enter / Shift+Enter and the ‹ › buttons step through all,
  scrolling the table and window to the current match.
- The query drives the table filter but isn't shown in the "Search variant
  table" box, which is disabled while the bar is open and restored on close.
- The table keeps its own highlighting; the page overlay covers only matches
  outside it.

Adds `FindBar/{FindBarContext,PageFindBar,useVariantFindBridge}` with unit
specs, wires it into `App` and the shared `Variants` table, and adds gene-page
e2e tests. Frontend-only.

Screenshot 2026-06-17 at 1 27 48 PM
Screenshot 2026-06-17 at 1 28 01 PM

### Testing
Unit: `FindBar` + `VariantList` Jest suites. E2e (`GenePage.playwright.spec.ts`)
must run serially (`--workers=1`, as `test:playwright` does) — the suite shares
one loaded page across tests.

### Follow-up (maybe a separate PR, `@gnomad/ui`)
`VariantTable` scrolls a match into view by bumping an internal nonce and
querying the Grid's `.grid-row-highlight` row, because `Grid` doesn't expose
`scrollToItem(index, align)` or access to a row's DOM node. Adding an `align`
option (or row-node access) to `@gnomad/ui`'s `Grid` would let this become a
typed API call and drop both the nonce and the internal class-name coupling.


bpblanken:
Linked ticket: https://github.com/broadinstitute/gnomad-browser/issues/1679

We looked into this one today and have decided to close in favor of a different implementation.  There were a few issues:
- The new find bar widget overrided existing search in ways that broke the existing behavior of searching for a word on the page.
- There is existing search functionality in the Variants Table that this approach didn't behave with particularly well - extending and duplicating rather than working alongside.
- There is slash-command that seems to work already... we can maybe make sure command-F integrates with that.
- Some of the highlighting of variant rows was inconsistent (sometimes only a top-border highlight, sometimes only a bottom-border one, sometimes both).

PR #1894, opened 2026-06-18
chore: `black` -> `ruff format`

bpblanken:
- Replaces `black` with `ruff format`
- Reasoning: it's faster, reduces number of dependencies, under active development.


mattsolo1 (approved):
I think CI only checks `data-pipeline/src/data_pipeline` and `deploy/deployctl`, so these files were not formatted:

- `data-pipeline/caids/get_caids.py`
- `data-pipeline/caids/import_caids.py`
- `reads/reference-data/gtf2bed.py`


Otherwise looks good!

bpblanken:
ah, those got caught on my local branch and I missed on push, will add on a second pr.

PR #1896, opened 2026-06-19
chore: add uv package management; split out dev dependencies, hard pins in r…

bpblanken:
…equirements

This migrates the package manage to `uv`, which is now overwhelmingly standard in the python community and integrated into other TGG projects.  Enables hard-pins in requirements.txt from a one-time resolution in `uv.lock`.


.github/workflows/data-pipeline-ci.yml
```diff
@@ -19,25 +19,20 @@ jobs:
...
-        run: |
-          pip install wheel
-          pip install -r data-pipeline/requirements.txt
+        working-directory: data-pipeline
+        # Install the exact dependencies from uv.lock without re-resolving or updating
+        # the lockfile; fails if pyproject.toml has drifted from uv.lock.
+        run: uv sync --frozen
       - name: Check formatting
-        run: ruff format --check data-pipeline/src/data_pipeline data-pipeline/tests data-pipeline/caids reads development/scripts
+        run: uv run --project data-pipeline ruff format --check data-pipeline/src/data_pipeline data-pipeline/tests data-pipeline/caids reads development/scripts
```
bpblanken: ` --project data-pipeline` assists with this module not being in the root of the repo.

CONTRIBUTING.md
```diff
@@ -22,10 +22,20 @@ This should be enough to use the Docker Compose development environment. However
...
+  ```
+
+  Run pipeline tools through uv, e.g. `uv run pytest` or `uv run ruff check src/data_pipeline`.
+
+- For the remaining Python tooling (linters and deploy scripts), install dependencies with pip.
 
   ```
-  pip install -r data-pipeline/requirements.txt
   pip install -r requirements-dev.txt
   pip install -r deploy/deployctl/requirements.txt
```
rileyhgrant: If we specified the `dev` group in the `pyproject.toml`, can we install those dependencies alongside with our `uv sync` by passing `--group dev` to use that as our source of truth?

e.g from:
```
cd data-pipeline && uv sync --frozen
```
to:
```
cd data-pipeline && uv sync --group dev --frozen
```

If we do that, can we remove the `./requirements-dev.txt` file, and the `./pipelines/requirements.txt` files to ensure we always use the `pyproject.toml` as the source of truth, or do we want to keep those around for a bit for some particular reason?

rileyhgrant (commented):
Nice! I'm definitely on board with using `uv`, thanks for getting this going.

Are we planning on using this for all the python package management? Having half our dependencies specified in a `requirements___.txt` file, and half in a `pyproject.toml` file (some that exist side by side and overlap, in the case of dev and pipeline dependencies) seems like it could cause some confusion at some point. If there's no particular reason to keep them around, I'd be in favor or removing those two redundant `___.txt` files that now have dependencies specified in the `.toml` and `.lock` files.

As a specific question, we have a bit of a hacky hail version check in our `deploy` dir: https://github.com/broadinstitute/gnomad-browser/blob/aa420a1c9611c8aeba9d4846e4e37f5f2cc67542/deploy/deployctl/subcommands/dataproc_cluster.py#L28-L42

We had that in the past because a few times different team members had differing versions of hail locally on accident, which through `hailctl` caused the dataproc version of hail that ran a pipeline to sometimes differ from the version pinned in the pipeline requirements file. 

Maybe this check isn't needed anymore, and/or maybe there's a better way to do this. Just something that came to mind here, if we're moving to a `pyproject.toml` over a `requirements.txt`, since that snippet above reads the `requirements.txt` file.

CONTRIBUTING.md
bpblanken: Yes, the `--group dev` or `--dev` should work!  I half-tried to scope this change to just the `data-pipeline` folder and didn't notice that the `requirements-dev.txt` change had leaked out.

I think it's probably worth just migrating the whole repo to this pattern - I'll proceed with that.

.github/workflows/data-pipeline-ci.yml
```diff
@@ -19,25 +19,22 @@ jobs:
...
       - name: Check formatting
-        run: ruff format --check data-pipeline/src/data_pipeline data-pipeline/tests data-pipeline/caids reads development/scripts
+        run: uv run ruff format --check data-pipeline
       - name: Run Ruff
-        run: ruff check data-pipeline/src/data_pipeline
+        run: uv run ruff check data-pipeline
       - name: Run Pyright
-        run: pyright --project data-pipeline
+        run: uv run pyright --project data-pipeline
+      - name: Run pytest
```
bpblanken, line 39: Note the added tests!

data-pipeline/check.sh
```diff
@@ -4,22 +4,25 @@ echo "┏━━━ Clean ━━━━━━━━━━━━━━━━━━
...
 find . -name "*.pyc" -exec rm -f {} +
 
+echo "┏━━━ Syncing dependencies ━━━━━━━━━━━━━━━━━━━"
+# Install the exact dependencies from uv.lock into .venv without re-resolving or
+# updating the lockfile. Errors out if pyproject.toml has drifted from uv.lock.
+uv sync --frozen --project ..
+
 echo "┏━━━ Running pyright ━━━━━━━━━━━━━━━━━━━"
-pyright
+uv run --project .. pyright
```
bpblanken, line 13: this is pretty verbose but I wasn't able to come up with a better solution after opting for lifting `pyproject.toml` to the repo root.  I think it's fine, for now.

requirements.txt
```diff
@@ -0,0 +1,285 @@
+# This file was autogenerated by uv via the following command:
+#    uv export --no-hashes
```
bpblanken, line 2: this is autogenerated by `uv export --no-hashes > requirements.txt`, which should be in the `CONTRIBUTING.md`.  Kept in order to maintain the deployctl dataproc deployment.

requirements-dev.txt
```diff
@@ -1,2 +0,0 @@
-ruff==0.15.17  # This should be kept in sync with the version in .pre-commit-config.yaml
```
bpblanken, line 1: removed, uv manages dev dependencies entirely now.

pyproject.toml
```diff
@@ -1,22 +1,42 @@
-[tool.ruff]
-line-length = 120
-target-version = "py39"
-
-[tool.pylint.basic]
-# ds: frequently used name for a variable containing a Hail table
-good-names = [
-  "ds",
+[project]
```
bpblanken, line 1: I opted to just lift this to the root and merge the `deployctl` dependency in here (`jinja`) so that python development can be managed in a single place.

data-pipeline/pytest.ini
```diff
@@ -1,4 +1,5 @@
 [pytest]
+pythonpath = src
```
bpblanken, line 2: helps the tests pass via `./check.sh`

.github/workflows/data-pipeline-ci.yml
rileyhgrant: Woohoo!

data-pipeline/check.sh
rileyhgrant: Yeah I agree that this is totally fine.

deploy/README.md
```diff
@@ -12,9 +12,7 @@
 
   To allow pushing images to GCR, register `gcloud` as a Docker credential helper with `gcloud auth configure-docker`.
 
-- Python dependencies
-
-  `pip install -r deploy/deployctl/requirements.txt`
+- Python requirements are managed by the `data-pipeline`'s `pyproject.yaml` and `uv`.
```
rileyhgrant: Minor pedantic nit: I think we have a `pyproject.toml`, not a `yaml`, and that it lives in the project root now, not the `data-pipelines`?

deploy/deployctl/subcommands/dataproc_cluster.py
```diff
@@ -25,7 +25,7 @@ def start_cluster(name: str, cluster_args: typing.List[str]) -> None:
     if not config.project:
         raise RuntimeError("project configuration is required")
 
-    with open(os.path.join(DATA_PIPELINE_DIRECTORY, "requirements.txt")) as requirements_file:
+    with open(os.path.join(HOME_DIRECTORY, "requirements.txt")) as requirements_file:
         requirements_hail_version = next(
             (line.strip().split("==")[1] for line in requirements_file if line.strip().startswith("hail==")), None
         )
```
rileyhgrant, line 31: I think that looking directly at the `pyproject.toml` file in the project root should find the version with this exact same code snippet, as the dependency pinning in the `.toml` file is pretty similarly formatted:

```
dependencies = [
  "elasticsearch~=7.17",
  "hail==0.2.127",
  "tqdm",
  "loguru",
```

But I also understand that's a change with a fairly minor benefit in that we could remove the frozen `requirements.txt` that's only used for this (kind of janky) check, so I'll leave the decision to you, and just wanted to mention it.

pyproject.toml
rileyhgrant: Works for me! I always thought that single `deployctl` dependency being split out did feel a bit odd, and that it really felt more like a single set of python dependencies for the whole project.

rileyhgrant (commented):
Overall looking good to me, thanks for moving the whole project to using `uv`. 

Two quick questions before I formally approve (one just for my sake to know in the future re: squash vs merge and how we communicate that)

1. Do you plan on squashing this PR into a single commit to merge into main, or rebasing it? I did this review assuming the squash route, so didn't look at individual commit messages/contents, but rather at the whole diff.

    Also, how have you seen this done in the past? Do people just write a line in the PR about "this will be rebased into main, please review each individual commit" or something like that?

2. Could we pin the version of `uv` used in the `.tool-versions` file in the root of the repo?

Also, thanks for getting tests running in CI! I'm a bit embarrassed I didn't realize that they haven't been doing that, and I'm glad that they are now :)

bpblanken:
@rileyhgrant 

I tend to really prefer squash-merging.  The intermediate commits on a PR tend to get very hairy (no matter how hard I try!) and consist of lots of back-and-forth.  A single PR should consist of some atomic unit of work that can be referred back to.  

and yep I can add to .tool-versions!

deploy/README.md
bpblanken: Yep this is just wrong!

deploy/deployctl/subcommands/dataproc_cluster.py
bpblanken: If I'm reading that code right, the entire set of requirements is being passed into the `hailctl` right?  At least in theory it's better to have the fully resolved dependency tree.

rileyhgrant (approved):
LGTM!

mattsolo1 (changes requested):
> **Note:** This review was generated by an AI code-review agent. Findings were verified locally where possible (see note at the end); treat as input for human review.

**`dataproc` cluster start will fail on the `uv export`-generated `requirements.txt`**

`deploy/deployctl/subcommands/dataproc_cluster.py` builds the cluster `--packages` list from the full contents of `requirements.txt`:

```python
requirements = [line.strip() for line in requirements_file.readlines()]   # L45
...
f"--packages={','.join(requirements)}",                                    # L62
```

Every line is forwarded to `pip` via `--packages`. That held for a hand-maintained file, but this PR switches to a file generated by `uv export --no-hashes`, which emits comment and blank lines. On this branch that is **185 of 285 lines**, e.g.:

```
# This file was autogenerated by uv via the following command:
#    uv export --no-hashes
    # via hail
    # via aiohttp
```

`gcloud dataproc clusters create ... --packages=...` passes these tokens directly to `pip`, which cannot parse entries like `# via hail`. `HOME_DIRECTORY` resolves to the repo root, so the file being read is the root `uv export` `requirements.txt`. Result: a real `deployctl dataproc start` fails. CI does not cover this path (no cluster is started), so it passes green.

Suggested fix — filter comments and blanks when constructing the package list:

```python
requirements = [
    line.strip()
    for line in requirements_file
    if line.strip() and not line.strip().startswith("#")
]
```

**Related (same root cause):** `uv export` without `--no-dev` includes the dev group, so `pyright==1.1.410`, `pytest==8.4.2`, and `ruff==0.15.18` are currently part of `requirements.txt` and would be installed on every dataproc cluster via `--packages`. Adding `--no-dev` to the export command (and updating the `CONTRIBUTING.md` snippet) removes these; the filter above would otherwise still forward them.

**Verification:** reproduced locally by simulating `','.join(...)` against this branch's `requirements.txt` (185/285 lines are comment/blank). The runtime `gcloud`/pip failure itself was not executed (requires a live dataproc start).

bpblanken:
😢 I ran into something very similar [in seqr](https://github.com/broadinstitute/seqr-loading-pipelines/blob/main/loading_pipeline/lib/tasks/dataproc/create_dataproc_cluster.py#L23-L29)

mattsolo1 (approved):

PR #1897, opened 2026-06-19
refactor: low stakes refactoring of caids

bpblanken:
Simple things:
- ReferenceGenome enum.
- lifting 5000 to a constant.
- a shared set of hail lifted to `select_locus_alleles`.
- moving string paths into a config.
- scaffolding of a directory which can be shared with `src` (I think we will eventually move it all there, but alas).  
- removal of `hl.init()` which is called by default when hail initializes.  `hl.init()` has value when non-default args are passed.


data-pipeline/core/enums.py
```diff
@@ -0,0 +1,13 @@
+from enum import Enum
+
+
+# Subclass `(str, Enum)` rather than `enum.StrEnum`: StrEnum is added in Python 3.11,
+# and the data pipeline targets Python 3.9.
+class ReferenceGenome(str, Enum):
+    GRCh37 = "GRCh37"
+    GRCh38 = "GRCh38"
```
bpblanken, line 8: Lots more utility in `src` for this class:

```
bblanken@wmbbe-67c data-pipeline % grep 'GRCh38' src -r | wc  -l
25```

**Review by nadeaujoshua — APPROVED** [review]

File: data-pipeline/core/config.py [hunk 2]
```diff
@@ -0,0 +1,23 @@
...
+# Most datasets are a single Hail Table; gnomAD v2.1.1 is the join of separate
+# exomes and genomes tables, and ExAC is a VCF rather than a Hail Table.
+DATASETS = {
+    "ExAC": "gs://gcp-public-data--gnomad/legacy/exac_browser/ExAC.r1.sites.vep.vcf.gz",
+    "gnomAD v2.1.1": {
+        DataType.EXOMES: "gs://gcp-public-data--gnomad/release/2.1.1/ht/exomes/gnomad.exomes.r2.1.1.sites.ht",
+        DataType.GENOMES: "gs://gcp-public-data--gnomad/release/2.1.1/ht/genomes/gnomad.genomes.r2.1.1.sites.ht",
+    },
+    "gnomAD v3.1.1": "gs://gcp-public-data--gnomad/release/3.1.1/ht/genomes/gnomad.genomes.v3.1.1.sites.ht",
+    "gnomAD v4.0": "gs://gcp-public-data--gnomad/release/4.0/ht/genomes/gnomad.genomes.v4.0.sites.ht",
```
rileyhgrant, line 22: See main comment for more context, and the filepaths for convenience, but if we're creating this config file, we should use the most recent v4 variant dataset hail tables here, (`v4.1.1` rather than `v4.0`), and include the `genomes`, `exomes` and `joint` hail tables.

I realize this was created based off the `export_vcfs` file as the source of truth. Just commenting because I noticed that file was unfortunately out of date!

rileyhgrant (commented):
I like the idea of pulling this out into a config, it'll make everything much simpler to keep in sync. All the low stakes refactoring portions looks good to me!

It looks like the values for the `config.py` file were populated from the `export_vcfs.py`, which makes sense, though that file is, unfortunately, out of date.

In that file, it looks like it only takes in the genomes sites table for gnomAD v4 from which to export VCFs, which are then used to generate CAIDS. There are actually both an exomes, and genomes, sites table for gnomAD v4, a la v2, so we might be missing some CAIDS in production for v4.

The v4.0.0 data should work, as it should just be based off variant IDs, and so the updates that have dealt with annotations and and frequencies should produce the same set of CAIDs, but the most recent dataset for v4 is not v4.1.1, not v4.0.0.

The v4.1.1 paths for exomes and genomes are:

```
gs://gcp-public-data--gnomad/release/4.1.1/ht/exomes/gnomad.exomes.v4.1.1.sites.ht/

gs://gcp-public-data--gnomad/release/4.1.1/ht/genomes/gnomad.genomes.v4.1.1.sites.ht/
```

Also, our data pipeline still uses the joint frequency data from the v4.1.0 release, as the v4.1.1 release was primarily updates to constraint, the addition of some flags, and update of loftee flags. So the frequency information from the joint table is still good.

```
gs://gcp-public-data--gnomad/release/4.1/ht/joint/gnomad.joint.v4.1.sites.ht/
```

That shouldn't be needed for CAIDs, but is probably good to throw into the config file, if we're starting it.

I can open an issue with the CAIDs to at least investigate if the CAIDS for variants present in exomes, but not genomes, are indeed missing in production.

I realize that these two things here:

- update to the `config.py` while it's being made
- the possible issue with missing some CAIDs due to the existing pipeline missing the exome sites table

Aren't directly pertinent to this refactor, but having eyes on this portion of the code brought them up.

As far as the refactors themselves, they look good to me.

Requesting the change that the config file has the v4.1.1 variants tables, rather than the v4.0.0 ones.

rileyhgrant (changes requested):
Woops, did 'comment' not, 'request changes' last time. Fixing that.

See big comment above.

---

Once that's all done, I won't be able to re-review, but you should be able to merge with another approval (e.g. Josh's approval that already exists). That's my only comment!

data-pipeline/core/config.py
bpblanken: Cool, happy to make this change.  

Has this process actually run against gnomad 4.1 though?  I spot-checked the date on the v4 caids that I could find I see:

```
       0 B  2024-03-06T19:24:06Z  gs://gnomad-browser-data-pipeline/caids/gnomad_v4_caids.ht/#1709753046082055  metageneration=1
     148 B  2024-02-13T20:52:27Z  gs://gnomad-browser-data-pipeline/caids/gnomad_v4_caids.ht/README.txt#1707857547552041  metageneration=1
       0 B  2024-02-13T20:52:27Z  gs://gnomad-browser-data-pipeline/caids/gnomad_v4_caids.ht/_SUCCESS#1707857547552232  metageneration=1
 15.84 KiB  2024-02-13T20:52:45Z  gs://gnomad-browser-data-pipeline/caids/gnomad_v4_caids.ht/metadata.json.gz#1707857565615487  metageneration=1
                                 gs://gnomad-browser-data-pipeline/caids/gnomad_v4_caids.ht/globals/
                                 gs://gnomad-browser-data-pipeline/caids/gnomad_v4_caids.ht/index/
                                 gs://gnomad-browser-data-pipeline/caids/gnomad_v4_caids.ht/rows/
```

which pre-dates the 4.1.1 release.

bpblanken: ah I see the ticket.

PR #1900, opened 2026-06-19
perf: remove lodash `omit` calls

bpblanken:
Some thoughts:
- JS object allocation is _expensive_!  Avoiding `lodash` is a strategy I've seen in the past:
- `omit` does not `clone`, but does several things:
```
1. flatRest arg packaging + baseFlatten
2. castPath/isKey per key (twice — once in step 1, once in baseUnset)
3. getAllKeysIn → prototype-chain walk
4. copyObject copying all N top-level keys (not just the kept ones)
5. delete on the few omitted keys

```
- Wrote a brief performance harness w/ 100k variants, which attempts to mimic the pattern.  `Isolated` tests the `omit` option in isolation, `Realistic` includes mimicked  `getFlagsForContext`, `createInSilicoPredictorsList`, `stripPops` etc.

```
┌───────────────┬───────────────┬──────────────┬──────────────────────┐
│  Measurement  │     omit      │ destructure  │        delta         │
├───────────────┼───────────────┼──────────────┼──────────────────────┤
│ (1) Isolated  │ 10.675 µs/var │ 1.517 µs/var │ 7.04× · 916 ms saved │
├───────────────┼───────────────┼──────────────┼──────────────────────┤
│ (2) Realistic │ 10.988 µs/var │ 1.613 µs/var │ 6.81× · 938 ms saved │
└───────────────┴───────────────┴──────────────┴──────────────────────┘
```
- A [relevant blog post](https://williaminfante.medium.com/replacing-lodash-omit-using-object-restructuring-and-the-spread-syntax-d7af1607a390) I found.

The performance impact in the harness is obviously a severe over estimate of the actual performance gain, as most of the latency exists in elastic & elsewhere, but I think this is an easy win.

### DIFF
--- a/graphql-api/src/queries/clinvar-variant-queries.ts
+++ b/graphql-api/src/queries/clinvar-variant-queries.ts
@@ -1,4 +1,4 @@
-import { omit, throttle } from 'lodash'
+import { throttle } from 'lodash'
 
 import { withCache } from '../cache'
 import logger from '../logger'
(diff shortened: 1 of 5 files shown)


rileyhgrant:
I know this is still in the draft state, but just was snooping and wanted to say this is exciting!

PR #1902, opened 2026-06-19
Add `Makefile` for automatable checks

rileyhgrant:
Related: #1903 

Currently, to mimic all the steps that we check in our GitHub CI locally before submitting a PR, the clearest way to do this imo is to look at what gets run in GitHub via the `yaml` file, and run each of those commands locally. Candidly, this leads to me personally relying on the GitHub CI to run those checks more often than not, which is not ideal. Additionally, specifying what checks an LLM should perform becomes either a matter of keeping a `.md` file in sync with the CI `.yaml` files, or pointing LLM towards the CI files directly.

Here, we create a `Makefile` that holds calls to automated tools, e.g. formatters, linters, and test runners. That way it's easier to keep deterministic tool runs in sync for local checks, automated Github CI, and (soon) LLM agents.

There are `validate-...` targets for ease of running all the checks locally. For now, the CI doesn't call the make targets to run many steps in succession, for better output in the CI UI in terms of knowing what particular check failed (e.g. `validate` failed, versus `lint` faile). Happy to change this, of course, if we want to do this differently.

This way, local developer with or without LLMs can use e.g. `validate-browser` to run all the checks that the browser CI runs, `validate-graphql-api` for the API, etc. There's also a `validate-all` that runs every check defined.

---

We could instead use a `Justfile` to be more of a square peg fitting into a square hole, but the typical workarounds of a `.PHONY` file as target for the `Makefile` seems reasonable for now, given many OSs ship with `make`.

Also totally happy to do something that is more straightforward, I just want a way to keep this in sync more easily then keeping commands manually written in docs (for both developers, and soon LLM agents) in sync with the check that actually get run in our CI in GitHub.


bpblanken:
this is a good idea love it!

.github/workflows/browser-ci.yml
```diff
@@ -31,15 +33,17 @@ jobs:
...
           cache: 'pnpm'
-      - name: Install dependencies
-        run: pnpm install --frozen-lockfile
-      - name: Run ESLint
-        run: pnpm eslint browser
-      - name: Run stylelint
-        run: pnpm stylelint 'browser/**/*.(js|jsx|ts|tsx)'
-      - name: Type check
-        run: pnpm ts-node ./browser/build/buildHelp.ts && pnpm typecheck
+      - name: Install JS dependencies
```
bpblanken, line 36: The dependencies get cached by the line above right?

.github/workflows/browser-ci.yml
```diff
@@ -31,15 +33,17 @@ jobs:
...
+      - name: Lint CSS
+        run: make lint-css-browser
+      - name: Check types
+        run: make typecheck-browser
       - name: Verify clean build
-        run: cd browser && pnpm build
-      - name: Run Jest tests
-        run: pnpm jest --selectProjects browser
+        run: make build-browser
+      - name: Run tests
```
bpblanken, line 48: Is it worth actually running the full `make` in addition to the independent pieces just to validate that we haven't missed a step?

.github/workflows/browser-ci.yml
rileyhgrant: Yep, they should still be cached by the above step even though we're invoking `pnpm` through `make`. 

This step should still indeed be grabbing the cached versions to add them into `node_modules` when it gets a cache hit, rather than re-downloading them.

.github/workflows/browser-ci.yml
rileyhgrant: Hmm, I think no, but that's not a strong opinion. Happy to change it if anyone else feels strongly.

I personally didn't want to do that in each CI file, because then the 4 different CI workflows would validate each other's steps, and I didn't like the idea of the `browser-ci.yml` (could rename to the `frontend` file for clarity) file checking the data pipeline's python formatting, etc.

I do definitely see how there could be some things that don't get kept in sync if we're not careful on the occasions that we update our CI/set-of-things-we-check-locally, though I think I land on the side that faster and more scoped CI runs on every pull request are worth the occasional potential for drift if we don't pay close enough attention.

As always, happy to be outvoted here if the team wants to do something different. There's a `validate-<DIR>` target for each of the main directories we could use in the CI if we wanted easier keeping local/llm/CI checks in sync, but worse CI UI output.

Also definitely lmk if I'm misunderstanding the question.

bpblanken: Nope!  I agree with this, was just curious.

bpblanken (approved):

rileyhgrant:
Going to wait and either rebase this off the `uv` branch post incoming changes, or `main` when the `uv` branch gets merged.

I opened this after the `uv` stuff was started, and I'll deal with the minor merge conflicts in this branch, rather than making you deal with it.

.github/workflows/data-pipeline-ci.yml
```diff
@@ -32,12 +34,10 @@ jobs:
...
-        run: ruff format --check data-pipeline/src/data_pipeline data-pipeline/tests data-pipeline/caids reads development/scripts
-      - name: Run Ruff
-        run: ruff check data-pipeline/src/data_pipeline
-      - name: Run Pyright
-        run: pyright --project data-pipeline
+        run: make check-format-data-pipeline
+      - name: Lint
+        run: make lint-data-pipeline
+      - name: Check types
+        run: make typecheck-data-pipeline
```
bpblanken: @rileyhgrant 

I just noticed we don't actually run the `pytest` tests here, whereas we do in `./check.sh`.  

I'm running into issues locally with them (nothing super interesting, just path resolution nonsense), and am wondering if it's worth just removing from `./check.sh` until they are thought through more thoroughly?

bpblanken: Nevermind!  I was able to get the tests passing reasonably (by changing pythgpath in the config) and added to CI. We can add to the `Makefile` up here too!

rileyhgrant: It looks like you got running tests sorted on your other branch. To be honest, I didn't really realize that `./check.sh` exists, and also didn't realize that running `pytest` locally was an issue. Both pretty big oopses from.

I guess that's just to say thanks for looking into this!

rileyhgrant:
@bpblanken 

I've rebased this off of `benb/add_uv_package_management`, and have this PR currently targeting that branch. I didn't want to make you have to deal with conflicts with pip/uv in the Makefile that was being used in the CI as part of this PR. The smol conflicts are all resolved and everything is looking good on this PR to my eyes.

Since this now is rebased off of that branch, is it possible for you to merge this PR while I'm away, after you merge the `benb/add_uv_package_management` branch?

Thanks!

PR #1904, opened 2026-06-22
perf: a few more lightweight, risk-free JS optimizations

bpblanken:
- Lifts two constant arrays to the global scope, preventing object allocations on the `variants` hot path.
- Moves two sequential `awaits` into ` Promise.all`.
- Mutates an existing `variant` object rather than creating a new one.

### DIFF
--- a/graphql-api/src/queries/coverage-queries.ts
+++ b/graphql-api/src/queries/coverage-queries.ts
@@ -171,23 +171,27 @@ export const _fetchCoverageForGene = async (esClient: any, datasetId: any, gene:
   // @ts-expect-error TS(7053) FIXME: Element implicitly has an 'any' type because expre... Remove this comment to see the full error message
   const genomeCoverageIndex = COVERAGE_INDICES[datasetId].genome
 
-  const exomeCoverage = exomeCoverageIndex
-    ? await fetchCoverage(esClient, {
+  const exomeCoveragePromise = exomeCoverageIndex
+    ? fetchCoverage(esClient, {
         index: exomeCoverageIndex,
         contig: gene.reference_genome === 'GRCh38' ? `chr${gene.chrom}` : gene.chrom,
         regions: mergedExons,
         bucketSize,
       })
     : []
-
-  const genomeCoverage = genomeCoverageIndex
-    ? await fetchCoverage(esClient, {
+  const genomeCoveragePromise = genomeCoverageIndex
+    ? fetchCoverage(esClient, {
         index: genomeCoverageIndex,
         contig: gene.reference_genome === 'GRCh38' ? `chr${gene.chrom}` : gene.chrom,
         regions: mergedExons,
         bucketSize,
       })
     : []
+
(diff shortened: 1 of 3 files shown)


graphql-api/src/queries/variant-datasets/gnomad-v4-variant-queries.ts
```diff
@@ -555,16 +554,14 @@ const fetchVariantsByRegion = async (esClient: any, region: any, subset: Subset)
     ])
   )
 
-  const variantsWithLofCurations = variants.map((variant: any) => ({
-    ...variant,
-    lof_curation: variant.transcript_consequence
+  for (const variant of variants) {
```
bpblanken, line 557: A few thoughts:
- This deviates from the existing functional style of the repo, which is potentially worth discussion.
- `map` allocates a new array on top of the object allocations, hence the for loop.

graphql-api/src/queries/coverage-queries.ts
```diff
@@ -243,6 +246,11 @@ const _fetchCoverageForTranscript = async (esClient: any, datasetId: any, transc
       })
     : []
 
+  const [exomeCoverage, genomeCoverage] = await Promise.all([
+    exomeCoveragePromise,
+    genomeCoveragePromise,
+  ])
```
rileyhgrant, line 252: Nice one.

rileyhgrant (approved):
LGTM!

mattsolo1 (approved):
> **Note:** AI-generated review from a code-review agent. Verified locally where noted; treat as input for human review.

Approving. The optimizations are safe:
- In-place mutation of `variant.lof_curation` is safe — variants are freshly `.map(shapeVariantSummary)`ed per request, not shared/cached.
- Lifting `IN_SILICO_PREDICTOR_IDS` / `REGIONAL_FLAGS` to module scope is safe (read-only).

**Verified locally:** applied to `main`, `flags.spec.ts` → 85/85 pass; typecheck clean.

Non-blocking nit: in `_fetchCoverageForTranscript`, `genomeCoveragePromise` still holds `? await fetchCoverage(...)` — an awaited value rather than a promise. It works (requests still overlap), but the gene path dropped both `await`s; dropping it here too would match the PR's intent.

mattsolo1 (approved):
> **Note:** AI-generated review from a code-review agent. Verified locally where noted; treat as input for human review.

Approving. The optimizations are safe:
- In-place mutation of `variant.lof_curation` is safe — variants are freshly `.map(shapeVariantSummary)`ed per request, not shared/cached.
- Lifting `IN_SILICO_PREDICTOR_IDS` / `REGIONAL_FLAGS` to module scope is safe (read-only).

**Verified locally:** applied to `main`, `flags.spec.ts` → 85/85 pass; typecheck clean.

Non-blocking nit: in `_fetchCoverageForTranscript`, `genomeCoveragePromise` still holds `? await fetchCoverage(...)` — an awaited value rather than a promise. It works (requests still overlap), but the gene path dropped both `await`s; dropping it here too would match the PR's intent.

PR #1909, opened 2026-06-23
chore: local api dev up and running

bpblanken:
Some small things I encountered when getting a local GraphQL instance running for profiling.

### DIFF
--- a/graphql-api/src/graphql/rate-limiting.ts
+++ b/graphql-api/src/graphql/rate-limiting.ts
@@ -6,21 +6,32 @@ import { isWhitelistedIP } from '../whitelist'
 
 let rateLimitDb: Redis
 
-if (config.REDIS_USE_SENTINEL) {
+let increaseRateLimitCounter = async (
+  _key: string,
+  _value: number
+): Promise<number> => 0
+
+
+if (config.REDIS_HOST && config.REDIS_USE_SENTINEL) {
   rateLimitDb = new Redis({
     sentinels: [{ host: config.REDIS_HOST, port: config.REDIS_PORT }],
     name: config.REDIS_GROUP_NAME,
     db: 2,
   })
-} else {
+} else if (config.REDIS_HOST) {
   rateLimitDb = new Redis({
     host: config.REDIS_HOST,
     port: config.REDIS_PORT,
     db: 2,
   })
+} else {
+  logger.warn('No redis configured for rate-limiting')
 }
 
-const increaseRateLimitCounter = (key: any, value: any) => {
+const increaseRateLimitCounter = (key: any, value: any): Promise<number> => {
+  if (!config.REDIS_HOST) {
+    return 0;
+  }
   return Promise.race([
     new Promise((resolve, reject) => {
       rateLimitDb
@@ -58,7 +69,6 @@ export const applyRateLimits = async (request: any) => {
       1
     )
 
-    // @ts-expect-error TS(2571) FIXME: Object is of type 'unknown'.
     if (totalRequestsInWindow > config.MAX_REQUESTS_PER_MINUTE) {
       throw new UserVisibleError('Query rate limit exceeded. Please try again in a few minutes.')
     }
@@ -68,7 +78,6 @@ export const applyRateLimits = async (request: any) => {
       request.graphqlQueryCost || 0
     )
 
-    // @ts-expect-error TS(2571) FIXME: Object is of type 'unknown'.
     if (totalCostInWindow > config.MAX_QUERY_COST_PER_MINUTE) {
       throw new UserVisibleError('Query rate limit exceeded. Please try again in a few minutes.')
     }
--- a/graphql-api/src/graphql/rate-limiting.ts
+++ b/graphql-api/src/graphql/rate-limiting.ts
@@ -26,9 +26,9 @@ const increaseRateLimitCounter = (key: any, value: any): Promise<number> => {
   if (!rateLimitDb) {
     return Promise.resolve(0)
   }
-  return Promise.race([
-    new Promise((resolve, reject) => {
-      rateLimitDb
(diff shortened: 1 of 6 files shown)


graphql-api/start.sh
```diff
@@ -2,22 +2,10 @@
 
 export NODE_ENV=development
 
-DEFAULT_PORT="8010"
```
bpblanken, line 5: I found these a tad confusing, as there are defaults supplied in the config.ts itself and they did not fully align (e.g. `CACHE_REDIS_URL` does not exist outside this file, the actual env var is `REDIS_HOST`).  I believe this change to be safe after once overing `gnomad-deployments`, but wouldn't mind other 👀s.

graphql-api/src/graphql/rate-limiting.ts
```diff
@@ -4,26 +4,31 @@ import { UserVisibleError } from '../errors'
 import logger from '../logger'
 import { isWhitelistedIP } from '../whitelist'
 
-let rateLimitDb: Redis
+let rateLimitDb: Redis | undefined;
 
-if (config.REDIS_USE_SENTINEL) {
+if (config.REDIS_HOST && config.REDIS_USE_SENTINEL) {
```
bpblanken, line 9: Gives this file a clean startup if Redis is not configured.

graphql-api/start.sh
bpblanken: Note that the api does not start without `ELASTICSEARCH_URL` defined, which I believe is correct behavior!

rileyhgrant: Hmm, that is indeed a bit confusing. 

I also did a once over the `gnomad-deployments`, and it also appeared safe to me do the explicitly setting of the required environment variables. I'll probably second your request for yet another pair of eyes (3 sets total?) here, just in case.

graphql-api/src/graphql/rate-limiting.ts
rileyhgrant: Nice! The elimination of periodic logs about redis in the running local API is really, really nice.

rileyhgrant (commented):
Looks great to me!

Leaving this as a comment, not an approval for a reason described below.

Main comment is that similar to you, I'd like yet another set of eyes to sanity check that the changing of the default value in `../config.ts` is safe in prod. I looked, and like you, believe that it is safe. I'm going to just comment on this, and leave whoever the third person to look at this and sanity check this to give the explicit approval required to merge this.

Then on a pedantic/housekeeping front, I have the same question as the `uv` PR. Are you planning on squashing or rebasing this? I didn't look at commit content/messages, assuming that it would be squashed, but wanted to check!

graphql-api/start.sh
bpblanken: @nadeaujoshua would you mind sanity checking the kubernetes env vars?

nadeaujoshua: Hey yeah, this looks right. Everywhere I looked we're using port 8000, but I guess it doesn't really matter what it is in the config file.

I do see some of those env vars are used in `development/api.docker-compose.yaml`, where they can probably be removed too.

bpblanken: What port does the browser run on in development?  There is this reference in the CONTRIBUTING.md


> - Start a local instance of the frontend, pointing to a local API (see below to start a local API)
> ```
> GNOMAD_API_URL=http://localhost:8010/api pnpm start:browser
> ```

rileyhgrant: In development I've always used `8010`, but I think that's just from the default. Maybe locally we didn't want a collision on `8000` so chose `8010`, but in production wanted `8000` by convention?

I think all the env vars set in this PR will keep things working like they have for the API, `8010` in local development, `8000` in production.

mattsolo1 (approved):
> **Note:** AI-generated review from a code-review agent. Verified locally where noted; treat as input for human review.

Approving. The one prod-relevant change is the default `PORT` fallback `8000` → `8010` in `config.ts`.

**PORT sanity check (passed):** every deployed/dev environment sets `PORT` explicitly, so the fallback change has no effect on them:
- `deploy/manifests/browser/base/api.deployment.yaml` sets `PORT: '8000'`, `containerPort: 8000`; `api.service.yaml` targets `8000`.
- `development/api.docker-compose.yaml` sets `PORT=8000` (maps `8010:8000`).

The new `8010` default only applies to a bare local `node` run with no env — the intended dev QoL improvement. `rate-limiting.ts` correctly makes `rateLimitDb` optional and no-ops without Redis.

Non-blocking nits: trailing semicolon on `let rateLimitDb: Redis | undefined;` (repo is no-semi); `api.docker-compose.yaml` still sets the now-dead `CACHE_REDIS_URL` / `RATE_LIMITER_REDIS_URL`.

mattsolo1 (approved):
> **Note:** AI-generated review from a code-review agent. Verified locally where noted; treat as input for human review.

Approving. The one prod-relevant change is the default `PORT` fallback `8000` → `8010` in `config.ts`.

**PORT sanity check (passed):** every deployed/dev environment sets `PORT` explicitly, so the fallback change has no effect on them:
- `deploy/manifests/browser/base/api.deployment.yaml` sets `PORT: '8000'`, `containerPort: 8000`; `api.service.yaml` targets `8000`.
- `development/api.docker-compose.yaml` sets `PORT=8000` (maps `8010:8000`).

The new `8010` default only applies to a bare local `node` run with no env — the intended dev QoL improvement. `rate-limiting.ts` correctly makes `rateLimitDb` optional and no-ops without Redis.

Non-blocking nits: trailing semicolon on `let rateLimitDb: Redis | undefined;` (repo is no-semi); `api.docker-compose.yaml` still sets the now-dead `CACHE_REDIS_URL` / `RATE_LIMITER_REDIS_URL`.

PR #1911, opened 2026-06-24
feat: graphQL api logging improvements

bpblanken:
Ok, I have this tested and working!

The gist:
- adds a request id & some per-request resource tracking baselines using the `AsyncLocalStorage` [utility](https://nodejs.org/api/async_context.html#introduction).  there's maybe way more we can do here, but it's scary!
- adds `json()` middleware so that the "POST" body of the graphQL query can be logged when a request starts.  The idea here is to help track down a request that blows the memory and triggers a restart.
- adds elasticsearch "start" and "end" logs, allowing traces by request id.
-  some elasticsearch "queue" latency logging, to see if the "Bottle" rate limiter is causing issues (my gut feeling is that it is not).


graphql-api/src/elasticsearch.ts
```diff
@@ -45,90 +46,128 @@ export const catchNotFound = (err: any) => {
...
+      requestId: ctx?.requestId,
+      event: 'esRequestQueueTimeout',
+      operation,
+      queueMs: performance.now() - queuedAt,
+      queued: await esLimiter.queued(),
+      running: await esLimiter.running(),
+    })
+  }, config.ELASTICSEARCH_QUEUE_TIMEOUT)
+
+  return esLimiter
```
bpblanken: One piece to note here is that I removed the `new Promise()` that was wrapping the esLimiter.  My understanding is that `esLimiter.schedule()` already returns a `Promise`, so we can avoid the additional wrapper.

The function passed to `.schedule()` also expects a returned `Promise()`, which we supply by `return fn().then()`.  Another strategy here might be:

```
  try {
    // log the request start
    const res = await fn();
    // log the request end
  } 
  catch {
    // log the error
  }

**Comment by bpblanken** [comment]
One testing note.  My harness is running a large-ish "variants-in-gene" query in one terminal and a single variant request in the other.  While the large query hangs in ES, the latency on the single variant blows up from 400ms -> 53s.  The log "under load"

```
{
  "requestId":"fd426ddc-d9f2-434b-9a43-8177f9e7fc4f",
  "event":"requestEnd",
  "latencyMs":53290.37275004387,
  "cpuUserMicros":2302968,
  "cpuSystemMicros":264641,
  "heapUsed":547004744,
  "heapDeltaBytes":127483488,
  "httpRequest":{"requestMethod":"POST","requestUrl":"http://localhost/api","userAgent":"python-requests/2.31.0","remoteIp":"::1","protocol":"HTTP/1.1","status":200,"responseSizeBytes":"85"},
  "graphqlRequest":{"graphqlQueryOperationName":null,"graphqlQueryString":"\nquery getVariant($variantId: String!) {\n  variant(variantId: $variantId, dataset: gnomad_r4) {\n    exome {\n      ac\n      an\n    }\n    genome {\n      ac\n      an\n    }\n  }\n}\n",
  "graphqlQueryVariables":{"variantId":"1-55051215-G-GA"},
  "graphqlQueryCost":1},
  "severity":"INFO"
}
```

File: graphql-api/src/app.ts [hunk 2]
```diff
@@ -23,64 +24,102 @@ process.on('unhandledRejection', (error) => {
...
+            }
+          }
+        : null,
+    })
+    next()
   })
+})
+
+app.use((req: any, res: any, next: any) => {
+  // NB: ALS propagation can occasionally surprise in third party contexts?
```
rileyhgrant: Little bit of a nit, and non-blocking:

I'm a bit confused as to what this comment wants me to note well. The question mark makes me think that it its not entirely sure as well

Is this comment pertaining to the needing to save the `ctx` to a local variable outside of the `onFinished` block?

If so, is it possible to change this to something like ~`NB: in the onFinished block, we lose access to the async context. Save it here to be able to call the ctx variable for logging`?

If not, what does this comment refer to? Is it just saying that sometimes, we might not have this context when we expect to?

rileyhgrant (approved):
This looks good to my eyes, with the caveat that I actually didn't try it myself locally, instead just looked over the diff.

Only minor, non blocking comment is about possibly rewording a comment that signaled itself as important (`NB:...`), though made me confused upon reading it.

graphql-api/src/app.ts
bpblanken: yep, this is a great call out of an incomplete and confusing comment ❤️!  

I flip/flopped this line quite a bit and spent some time exploring how exactly ALS works and was unable to really come to a clear conclusion.  Both putting the ` ctx = requestStore.getStore()` seem to work fine in local testing, but this was me running no more than 2 concurrent requests with a minimal testing harness.   

My best understanding of this is that the `onFinished` callback can run outside of the original request flow (including being trigger able by third party APIs), and that being defensive and relying on closure capturing will prevent `ctx` from potentially being `null` within the callback (which tbh would just return immediately).

I will update the comment to your suggestion with an additional "potentially" to indicate our own lack of thorough understanding.

mattsolo1 (changes requested):
Claude found a bug that drops the user-visible error after timeouts. Reproduced with a proposed fix in #1918.

bpblanken:
merged the add-on!

mattsolo1 (approved):

PR #1919, opened 2026-07-09
feat: incorporate handlers for sigterm and sigkill in app.ts teardown.

bpblanken:
Polishes some POC I wrote a few weeks ago to clean up the teardown, which was required to generate profiles.

### DIFF
--- a/graphql-api/src/app.ts
+++ b/graphql-api/src/app.ts
@@ -12,16 +12,6 @@ import { requestStore } from './request-context'
 import { closeCache } from './cache'
 import { loadWhitelist } from './whitelist'
 
-process.on('uncaughtException', (error) => {
-  logger.error(error)
-  process.exit(1)
-})
-
-process.on('unhandledRejection', (error) => {
-  logger.error(error)
-  process.exit(1)
-})
-
 const app = express()
 app.use(compression())
 app.use(cors())
@@ -123,11 +113,14 @@ app.use('/api/',
   })
 )
 
+// On shutdown (SIGTERM/SIGINT) or fatal error (uncaughtException/unhandledRejection), stop accepting
+// requests, drain connections, and close ES/cache clients before exiting. A 10-second timeout forces
+// exit if graceful teardown stalls.
 const server = app.listen(config.PORT, () => {
   logger.info({ event: 'serverStart', port: config.PORT })
 })
 
-const shutdown = (signal: string) => {
+const shutdown = (signal: string, exitCode = 0) => {
   logger.info({ event: 'shutdown', signal })
 
   const forceExit = setTimeout(() => {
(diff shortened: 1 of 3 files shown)


graphql-api/src/app.ts
```diff
@@ -122,4 +113,40 @@ app.use('/api/',
   })
 )
 
-app.listen(config.PORT)
+// On shutdown (SIGTERM/SIGINT) or fatal error (uncaughtException/unhandledRejection), stop accepting
+// requests, drain connections, and close ES/cache clients before exiting. A 10-second timeout forces
+// exit if graceful teardown stalls.
+const server = app.listen(config.PORT, () => {
+  logger.info({ event: 'serverStart', port: config.PORT })
```
bpblanken, line 120: I tested this with a couple scripts:

Fixed, with profiling: 
```
const http = require('http')

const PORT = process.env.PORT || 3002

function doWork() {
  let n = 0
  for (let i = 0; i < 5_000_000; i++) {
    n += Math.sqrt(i) * Math.sin(i)
  }
  return n
}

const app = http.createServer((req, res) => {
  const result = doWork()
  res.writeHead(200, { 'Content-Type': 'text/plain' })
  res.end(`result: ${result}\n`)
})

// New pattern: store the handle so shutdown can close the server gracefully.
const server = app.listen(PORT, () => {
  console.log(`[fixed]  listening on ${PORT} (pid ${process.pid})`)
})

const shutdown = (signal) => {
  console.log(`[fixed]  received ${signal}, closing server…`)

  const forceExit = setTimeout(() => {
    console.error('[fixed]  shutdown timed out, forcing exit')
    process.exit(1)
  }, 10_000)
  forceExit.unref()

  server.close(() => {
    clearTimeout(forceExit)
    console.log('[fixed]  server closed, calling process.exit(0)')
    // process.exit() triggers V8 at-exit hooks → --cpu-prof writes the file.
    process.exit(0)
  })
}

process.on('SIGTERM', () => shutdown('SIGTERM'))
process.on('SIGINT',  () => shutdown('SIGINT'))
```

old:
```
const http = require('http')

const PORT = process.env.PORT || 3001

function doWork() {
  let n = 0
  for (let i = 0; i < 5_000_000; i++) {
    n += Math.sqrt(i) * Math.sin(i)
  }
  return n
}

const app = http.createServer((req, res) => {
  const result = doWork()
  res.writeHead(200, { 'Content-Type': 'text/plain' })
  res.end(`result: ${result}\n`)
})

// Old pattern: result is not stored, so server.close() is unreachable later.
app.listen(PORT, () => {
  console.log(`[broken] listening on ${PORT} (pid ${process.pid})`)
})
```


The issue is that V8 exit hooks are not reached without the handle to call server.close() and process.exit().

nadeaujoshua (approved):
I can only speak to the Redis portion, but the closing of the cache looks good to me.

It might help with the `ioredis` timeout problem I'm seeing on certain restarts in the Cloud Run deployment, so thanks for putting this in and 🤞.

PR #1920, opened 2026-07-27
feat: update gnomAD TR data to the July 2026 release and fetch the reads db from GCS

bw2:
Two related changes for the July 2026 (`2026_07_20`) tandem repeat release. Changelog entry: https://gnomad.broadinstitute.org/news/2026-07-str-data-update/

## 1. Point the browser and pipeline at the new release

**Downloads page** (`browser/src/DataPage/GnomadV3Downloads.tsx`)
- Adds an "Update (July 2026)" section above the March 2025 one, linking the new genotypes TSV and distributions JSON.
- The README link intentionally points at the existing `README__2025_03_17.txt`: the column schema is unchanged from the March 2025 release (same 22 columns, same order), so no new README was produced. Its `loggingLabel` deliberately omits a July date so it does not imply a new document.
- The new entries carry distinct `loggingLabel`s so their download counts are separable from the 2025 and 2022 releases.

**TR overview page** (`browser/src/ShortTandemRepeatsPage/ShortTandemRepeatsPage.tsx`) — notes the data version (`2026-07-20`) in the intro paragraph.

**Data pipeline** — `pipelines/gnomad_v3_short_tandem_repeats.py`: default input is now the release's distributions JSON, overridable with `STR_DISTRIBUTIONS_JSON` (useful for local runs against a local file, which avoids GCS auth). Note this changes the input convention from a copy under `gs://gnomad-browser-data-pipeline/inputs/secondary-analyses/strs/` to the browser bucket's gzipped JSON — happy to move the file instead if the old convention is preferred.

## 2. Fetch the tandem repeat reads DB from GCS at startup

Previously the reads service opened `/readviz/datasets/gnomad_r4_short_tandem_repeats/str_reads.db` from a 6552 GB read-only persistent disk, so shipping a new ~170 MB DB meant snapshotting and cloning that disk. It now downloads the DB from a public GCS URL into the container's writable filesystem.

- **New module** `reads/src/shortTandemRepeatReadsDb.js` — download, verify, and hand out a memoized read-only handle.
  - Streamed with `pipeline()`, never buffered: the container has a 256Mi memory limit and the file is ~170 MB.
  - Downloads to a temp path and `rename()`s into place, so a partial transfer can never be opened.
  - The cache filename is derived from a hash of the URL, so pointing `STR_READS_DB_URL` at a new release is a cache miss by construction and can never serve the previous release's file.
  - Aborts on a 60s **stall** rather than capping total duration — a duration cap fails on any link slower than size/cap even when the transfer is healthy.
  - Verifies with `PRAGMA quick_check` plus a `SELECT 1 FROM reads` schema probe, since `quick_check` alone would pass an unrelated database.
  - Opens `OPEN_READONLY`. The default mode would silently create an empty DB if the file were missing, turning a startup failure into a confusing per-request `no such table: reads`.
- **`reads/src/server.js`** warms the DB before `listen()` so the common case is not a first request waiting on a download. A failure there is logged but not fatal: this service also serves variant reads off `/readviz`, which have no dependency on this DB, so exiting would cause a much larger outage than the one being reported.
- **`reads/src/resolveShortTandemRepeatReads.js`** uses one shared handle instead of opening and closing the DB on every request (it did so twice per page load). A query awaits the fetch, so it either waits on the in-flight download or retries a failed one.
- **`jest.config.ts`** gains a `reads` project. Without it, specs under `reads/` are silently never run. `.eslintrc.js` extends its jest-env override to `*spec.js` so those specs lint.

Also included: an `EP400` readviz folder override. EP400's images for the narrowed 548-611 locus definition live in `EP400__12-132062548-132062611/`, while `EP400/` still holds the wider-definition images (verified: genuinely different files, 63,718 vs 32,281 bytes). Without the override the Read Data panel would show images that disagree with the histograms. The entry should be deleted once `EP400/` is restaged.

## Data dependencies

- The reads DB is uploaded at `gs://gnomad-str-public/release_2026_07/reads_db/str_reads_2026_07_20.db` (public, identity-encoded, crc32c verified against the source file). Identity encoding matters: the size check is skipped for encoded responses.
- The download links 404 until the genotypes TSV and distributions JSON are published to `gs://gcp-public-data--gnomad/release/3.1.3/{tsv,json}/` **and** the AWS `gnomad-public-us-east-1` mirror, since the page renders a Google and an Amazon link for each file.
- The new Elasticsearch index already exists in production alongside the current one; the alias still points at `gnomad_v3_short_tandem_repeats-2025-03-17--19-04`, so nothing in this PR changes what the API serves.

## Checks

Run locally, matching the CI workflows:

- `pnpm eslint browser`, `pnpm eslint reads`, `pnpm stylelint`
- `pnpm ts-node ./browser/build/buildHelp.ts && pnpm typecheck`
- `pnpm jest` — 74 suites, 2059 tests, 1144 snapshots
- `ruff format --check data-pipeline` and `ruff check data-pipeline` with the version pinned in `uv.lock` (0.15.18)

Beyond the unit tests, the reads service was exercised end to end against the real 168 MB DB: a cold start downloads, verifies, and serves (`num_reads: 14035` for both EP400 and RAI1), the downloaded copy is byte-identical to the source, and with an unreachable DB URL the service still starts, still serves variant reads, and returns an error for tandem repeat queries.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -72,7 +72,7 @@ module.exports = {
   overrides: [
     {
       // Set environment for tests
-      files: ['**/*spec.ts', '**/*test.ts', 'tests/**/*.ts'],
+      files: ['**/*spec.ts', '**/*spec.js', '**/*test.ts', 'tests/**/*.ts'],
       env: {
         jest: true,
       },
--- a/browser/src/DataPage/GnomadV3Downloads.tsx
+++ b/browser/src/DataPage/GnomadV3Downloads.tsx
@@ -323,6 +323,13 @@ const GnomadV3Downloads = () => (
       <FileList>
         <ListItem>
           <DownloadLinks
+            label="README"
+            loggingLabel="v3.1.3 genotype TSV README (7-20-2026 update)"
+            path="/release/3.1.3/tsv/README__2025_03_17.txt"
+          />
+        </ListItem>
+        <ListItem>
+          <DownloadLinks
             label="Genotypes (TSV)"
             path="/release/3.1.3/tsv/gnomAD_STR_genotypes__2026_07_20.tsv.gz"
           />
--- a/browser/src/DataPage/GnomadV3Downloads.tsx
+++ b/browser/src/DataPage/GnomadV3Downloads.tsx
@@ -322,21 +322,25 @@ const GnomadV3Downloads = () => (
       <a href="https://gnomad.broadinstitute.org/news/2026-07-str-data-update/">changelog entry</a>]
       <FileList>
         <ListItem>
+          {/* The column schema did not change in this release, so this release reuses the
+              March 2025 README rather than publishing an identical new one. */}
           <DownloadLinks
             label="README"
-            loggingLabel="v3.1.3 genotype TSV README (7-20-2026 update)"
+            loggingLabel="v3.1.3 genotype TSV README"
             path="/release/3.1.3/tsv/README__2025_03_17.txt"
           />
         </ListItem>
         <ListItem>
           <DownloadLinks
             label="Genotypes (TSV)"
+            loggingLabel="v3.1.3 genotypes TSV (7-20-2026 update)"
             path="/release/3.1.3/tsv/gnomAD_STR_genotypes__2026_07_20.tsv.gz"
           />
         </ListItem>
         <ListItem>
           <DownloadLinks
             label="Distributions (JSON)"
+            loggingLabel="v3.1.3 distributions JSON (7-20-2026 update)"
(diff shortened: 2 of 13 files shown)


rileyhgrant:
What is happening with this open branch @bw2?

Based on the comment in the last paragraph of #1923, it seems like that one was merged over this PR?

> The browser and data-pipeline files here are byte-identical to the corresponding files in https://github.com/broadinstitute/gnomad-browser/pull/1920,

If that is the case, can we close this PR, and if not, what is the status of this?

bw2:
@rileyhgrant this PR was superseded by #1923 . I forgot to close this. Will close it now.

PR #1903, opened 2026-06-19
Add context files for LLMs/Agents

rileyhgrant:
This PR targets the `use-makefile-for-ci` branch, not `main`.

Does an inplace rename of the starting `CLAUDE.md` files, then adds a few basic skills to a directory in this repo and point the LLM instructions file (`AGENTS.md`) to make sure to use the skills when appropriate. 

The hope is this can help LLM assisted incoming PRs/changes to better adhere to existing standards, and to try and aid human reviewers by automating what we can, and removing some ~easily fixable noise (e.g. commit message names, content).

These skills are certainly not perfect, or maybe not even a decent starting point. All suggestions on better ways to go about this, and changes down the line, are expected.

I'm hoping we could use the existing methods team PRs, and the 2 remaining unopened ones to see what having these instructions, and skill files does in terms of the presentation and content of those LLM assisted PRs.

The required `Assisted-by: ____` tag in the commit message inclusion comes from how the Linux Kernel currently handles this: https://docs.kernel.org/process/coding-assistants.html. 

The Linux Kernel is certainly a project that needs to have more stringent processes and requirements than the gnomAD browser does. I do personally think their approach is pretty reasonable, but happy to go with whatever the team at large wants here. 

I also think that if we include some instruction in our agent instruction file(s) such as the inclusion of the `Assisted-by:` tag in commits, it becomes easier for us to see if any given incoming contribution followed our repos guidelines for LLMs. 

---

- Take in Ben's `Claude.md` files
- Rename them to `AGENTS.md` -- I haven't tested if this works yet, though I've read that Anthropic supports this pseudo standard. If not, we can create a symlink.
- Add a  `.agents/` dir with some initial `skills` (these I generated with an LLM and did some modifications to, will edit the respective commits with the `Assisted-by:___` tag if we want to do that

PR #1925, opened 2026-08-11
feat(frontend): add copyable anchor links to age distribution sections

### LINKED ISSUE #1921: Add link to age distribution section of variant pages
The ClinGen team is working through the updated draft guidance for variant classification. Part of this guidance discusses how to use population references like gnomAD as evidence. One feature that would be helpful for their team is a way to link directly to the Age Distribution section of variant pages (e.g., https://gnomad.broadinstitute.org/variant/1-55051215-G-GA?dataset=gnomad_r4):

Image

They'd like to add a link to this specific page location in their curation interface. Would it be possible to add a hyperlink to this section similar to the links we have on the [downloads](https://gnomad.broadinstitute.org/data#v4) page? Example link: 

Image

ch-kr:

resolves: #1921

### Merge strategy

### What does this PR do?

The ClinGen team is working through updated draft guidance for variant classification, which
covers using population references like gnomAD as evidence. Their curation interface needs to
link a reviewer directly to a variant's age distribution, rather than dropping them at the top
of the page to scroll and hunt for it.

This adds an `age-distribution` anchor to the Age Distribution heading, so
`/variant/1-55051215-G-GA?dataset=gnomad_r4#age-distribution` opens with that section in view.

### Architecture & Design

**Anchors reuse `withAnchor`.** Rather than inventing an affordance, this reuses the HOC already
behind the downloads and help page anchors, so the hover link icon looks and behaves the way it
does everywhere else on the site. A new `AnchoredSectionHeading` is exported from `AnchorLink.tsx`
so the four pages share one definition, mirroring how `DataPage/downloadsPageStyles.tsx` defines
its `SectionTitle` once.

**Copying is opt-in.** Clicking the icon copies the section URL, which suits a curation workflow
where the link gets pasted elsewhere. This is gated behind a new `copyUrlOnClick` prop, set only
by `AnchoredSectionHeading`, so the existing help, policies, and downloads anchors are unchanged.
It uses the same `navigator.clipboard` feature detection as the "Copy variant ID" button on the
variant page, and keeps `href` intact so right-click and middle-click still work.

**Why these pages need their own scrolling.** `App.tsx` already scrolls to the URL hash, but it
fires when a page's *code* finishes loading. These four pages render their content only after a
GraphQL query resolves, so at that moment the page still shows "Loading variant..." and the
section does not exist. The new `useScrollToHash` hook is therefore called from the component
that renders once data is available.

**Why the hook re-aligns rather than scrolling once.** A single scroll on mount lands in the wrong
place: the sections *below* the target have not rendered yet, so the document is too short for the
browser to bring the heading to the top and the scroll is clamped to the bottom of the page. This
was measured on a running browser — the heading sat at document offset 1809 while the page was
only 2558px tall, so `scrollY` clamped at 1658 and the heading came to rest 151px too low. The
amount varied per dataset, since each has a different quantity of late-arriving content. The hook
re-aligns on each animation frame for up to 2s while the page grows, stops as soon as the heading
is in place, and bails out immediately if the reader scrolls, touches, types, or clicks, so it
never fights someone who has started reading.

No new dependencies.

### Implementation

- Add `useScrollToHash`, which aligns the hash target while the page finishes rendering and yields
  to reader input
- Export `AnchoredSectionHeading` from `AnchorLink.tsx`, and add an opt-in `copyUrlOnClick` prop to
  `withAnchor` plus a `copyLinkToSection` helper that writes the absolute URL and shows the existing
  notification toast
- Use the anchored heading with `id="age-distribution"` and call `useScrollToHash` on the short
  variant, mitochondrial, structural variant, and short tandem repeat pages
- Regenerate the four page snapshots for the added anchor markup

### Tests

11 new tests, all passing:

- `useScrollToHash.spec.tsx` (7) — scrolls to the hash target; keeps re-aligning while the page
  grows; does not scroll when already aligned; stops once the reader scrolls; no-ops with no hash,
  with an unmatched hash, and with a hash that is not a valid CSS selector
- `AnchorLink.spec.tsx` (4) — renders the anchor and its id; leaves the clipboard alone for plain
  `withAnchor` consumers; copies the full section URL from `AnchoredSectionHeading`; does not throw
  where the Clipboard API is unavailable

68 snapshots across the four page suites were regenerated. The diff is limited to the added anchor
markup and the `onClick` handler.

**Coverage gaps, stated plainly:** jsdom has no layout engine, so `getBoundingClientRect` is
stubbed in the hook's tests. Those tests prove the retry loop runs and terminates correctly; they
do *not* prove it fixes real scroll clamping. That was verified by driving a real browser against
the dev server across `gnomad_r4`, `gnomad_r2_1`, `gnomad_r3`, and `exac` — all four land the
heading at 0px from the viewport top — but that verification is not committed and will not catch a
regression. A Playwright assertion in `tests/e2e/VariantPage.playwright.spec.ts` would be the place
to lock it in; happy to add one if reviewers want it, noting Playwright is not currently a CI gate
for browser changes.

[v4 short variant](https://gnomad.broadinstitute.org/variant/1-55051215-G-GA?dataset=gnomad_r4) now has link that, when clicked, copies a URL focused on the age distribution for the given variant:
image

This also applies to structural, mito, and TR variants:
[v4 SV](https://gnomad.broadinstitute.org/variant/DEL_CHR19_2C6DA7E7?dataset=gnomad_sv_r4) has same behavior as short variant (though reloaded page won't have the age distribution at the top due to shorter page length):
image

[v4 mito](https://gnomad.broadinstitute.org/variant/M-8602-T-C?dataset=gnomad_r4)
image

[TR](https://gnomad.broadinstitute.org/short-tandem-repeat/ATXN1?dataset=gnomad_r4)
image

1. **`MitochondrialVariantPage.tsx` looks far larger than it is.** Hosting a hook meant converting
   its concise arrow body (`=> (`) to a block body, which re-indented the entire JSX tree. Review
   it with `git diff -w`; the real change is 5 lines.

2. **The 2s settle window in `useScrollToHash` is a judgement call.** Measurements showed ~110ms
   between the clamped scroll and the corrected one, so 2s is comfortable, but content arriving
   after that on a very slow connection would leave the heading short. Worth a sanity check on the
   number.

3. **Short pages cannot always put the heading at the top, and this is not fixable by scrolling.**
   On e.g. `/variant/DEL_CHR19_413F7B28?dataset=gnomad_sv_r4#age-distribution`, the heading is
   478px from the end of a 1547px document, so reaching the top would require `scrollY=1069`
   against a ceiling of 647. The page is already scrolled to its end and the section is fully
   visible, just not at the top. The only lever would be padding the bottom of pages with empty
   space sitewide, which seems a poor trade for an edge case — flagging in case reviewers disagree.

4. **Mobile and breakpoint testing.** The deep link was checked across five viewports, straddling
   the 992px breakpoint where `ResponsiveSection` collapses to a single column. The heading lands
   0px from the top of the viewport in all of them:

   | Viewport | Deep link | Icon reachable |
   | --- | --- | --- |
   | iPhone 13 (390x844, touch) | 0px | tap |
   | Pixel 7 (412x915, touch) | 0px | tap |
   | iPad Mini (768x1024, touch) | 0px | tap |
   | narrow desktop (900x800, mouse) | 0px | hover |
   | wide desktop (1280x900, mouse) | 0px | hover |

   Below the breakpoint the document roughly doubles in height (3092px to 4866px on iPhone), which
   makes the clamping fix matter more, not less. Two caveats: this is Chromium device emulation
   rather than real hardware or WebKit, so iOS Safari's handling of `:hover` on tap is assumed
   rather than verified; and on touch the copy affordance needs two taps — one to reveal the icon,
   one to hit a 15x12px target. That awkwardness is inherited from the existing sitewide
   `withAnchor` design rather than introduced here, so this PR does not diverge from it, but
   reviewers may have a view.

5. **A changelog entry still needs writing** to announce this to users, once this PR is reviewed.

### Verification

- [x] I have the relevant `make` targets to run CI checks locally and verified that they succeed
      (note: `make validate-browser` arrives with #1903 and is not on `main` yet, so the equivalent
      steps from `browser-ci.yml` were run — eslint, stylelint, tsc, `pnpm build`, and the full
      Jest project: 1776 passing across 69 suites)
- [x] (If UI) I have tested this across desktop and mobile breakpoints.
- [x] (If a LLM/AI agent runner was used, e.g. OpenCode, Aider, or ClaudeCode) I have instructed the
      agent runner to load the `AGENTS.md` file into its context, and verified the the required
      `Assisted-by` trailer is on each commit.

### DIFF
--- a/browser/src/AnchorLink.spec.tsx
+++ b/browser/src/AnchorLink.spec.tsx
@@ -1,0 +1,66 @@
+import React from 'react'
+import { render } from '@testing-library/react'
+import userEvent from '@testing-library/user-event'
+import { afterEach, beforeEach, describe, expect, jest, test } from '@jest/globals'
+import styled from 'styled-components'
+
+import { AnchoredSectionHeading, withAnchor } from './AnchorLink'
+
+const PlainHeading = withAnchor(styled.h2``)
+
+const writeText = jest.fn(() => Promise.resolve())
+
+const originalClipboard = navigator.clipboard
+
+const setClipboard = (value: any) => {
+  Object.defineProperty(navigator, 'clipboard', { value, configurable: true, writable: true })
+}
+
+describe('withAnchor', () => {
+  beforeEach(() => {
+    writeText.mockClear()
+    setClipboard({ writeText })
+    window.history.pushState({}, '', '/variant/1-55051215-G-GA?dataset=gnomad_r4')
+  })
+
+  afterEach(() => {
+    setClipboard(originalClipboard)
+  })
+
+  test('renders an anchor carrying the section id', () => {
+    const { container } = render(
+      <PlainHeading id="age-distribution">Age Distribution</PlainHeading>
+    )
+    const anchor = container.querySelector('#age-distribution')
+    expect(anchor).not.toBeNull()
+    expect(anchor!.getAttribute('href')).toBe('#age-distribution')
+  })
+
+  test('does not touch the clipboard by default', async () => {
+    const { container } = render(
+      <PlainHeading id="age-distribution">Age Distribution</PlainHeading>
+    )
+    await userEvent.click(container.querySelector('#age-distribution')!)
+    expect(writeText).not.toHaveBeenCalled()
+  })
+
+  test('AnchoredSectionHeading copies the full section URL on click', async () => {
+    const { container } = render(
+      <AnchoredSectionHeading id="age-distribution">Age Distribution</AnchoredSectionHeading>
+    )
+    await userEvent.click(container.querySelector('#age-distribution')!)
+    expect(writeText).toHaveBeenCalledWith(
+      'http://localhost/variant/1-55051215-G-GA?dataset=gnomad_r4#age-distribution'
+    )
+  })
+
+  test('does nothing when the Clipboard API is unavailable', async () => {
+    setClipboard(undefined)
+    const { container } = render(
+      <AnchoredSectionHeading id="age-distribution">Age Distribution</AnchoredSectionHeading>
+    )
+    await expect(
+      userEvent.click(container.querySelector('#age-distribution')!)
+    ).resolves.not.toThrow()
+  })
+})
--- a/browser/src/AnchorLink.spec.tsx
+++ b/browser/src/AnchorLink.spec.tsx
@@ -1,119 +1,124 @@
 import React from 'react'
-import { render, screen } from '@testing-library/react'
+import { render, screen, within } from '@testing-library/react'
 import userEvent from '@testing-library/user-event'
 import { afterEach, beforeEach, describe, expect, jest, test } from '@jest/globals'
 import styled from 'styled-components'
 
-import { AgeDistributionHeading, withAnchor } from './AnchorLink'
+import { SectionHeading, withAnchor } from './AnchorLink'
 import Notifications from './Notifications'
 
 const PlainHeading = withAnchor(styled.h2``)
-
 const writeText = jest.fn(() => Promise.resolve())
-
 const originalClipboard = navigator.clipboard
 
 const setClipboard = (value: any) => {
   Object.defineProperty(navigator, 'clipboard', { value, configurable: true, writable: true })
 }
 
-const renderAgeDistributionHeading = () =>
-  render(
-    <>
-      <Notifications />
-      <AgeDistributionHeading>Age Distribution</AgeDistributionHeading>
-    </>
+beforeEach(() => {
+  writeText.mockClear()
+  setClipboard({ writeText })
+  window.history.pushState({}, '', '/variant/1-55051215-G-GA?dataset=gnomad_r4')
+})
+
+afterEach(() => {
+  setClipboard(originalClipboard)
+})
+
+test('withAnchor keeps legacy anchor targets unchanged and does not touch the clipboard', async () => {
+  const { container } = render(<PlainHeading id="help-section">Help section</PlainHeading>)
+  const anchor = container.querySelector('#help-section')
+
+  expect(anchor?.tagName).toBe('A')
+  expect(anchor?.getAttribute('href')).toBe('#help-section')
+  expect(anchor?.getAttribute('aria-hidden')).toBe('true')
+
+  await userEvent.click(anchor!)
+  expect(writeText).not.toHaveBeenCalled()
+})
+
+test('SectionHeading renders its title and adornments and forwards h2 attributes', () => {
+  const { container } = render(
+    <SectionHeading id="ancestry" title="Genetic Ancestry" className="custom" data-test="heading">
+      <button type="button">More information</button>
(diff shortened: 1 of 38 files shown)


bpblanken:
this is interesting!  from the description I'd have expected a much smaller diff, but looks like the LLM went and handled some additional edge cases.

ch-kr:
hopefully nothing excessive? 😬 it wrote what looked like a pretty simple plan initially
```
Understood — dropping the docs question. Announcing the change is a post-review changelog concern, not part of this work. I've noted that as a repo convention for future changes too.

The plan stands as written, scoped to code and tests:

useScrollToHash hook + spec
Shared withAnchor-wrapped Age Distribution heading with id="age-distribution"
Wired into all four variant-type pages
Snapshot regeneration, e2e matcher verified, make validate-browser
One commit with the Assisted-by: trailer and no Co-authored-by
Ready to implement whenever you want — just say go.
```
but tinkered around with the code more when I noted the age distribution in the new link wasn't at the top of the page (there was some buffer around it)

mattsolo1:
## Follow-up changes

This summarizes our additions in #1950/#1951 and the subsequent fixes.

- Made section links keyboard/touch accessible, with meaningful labels, visible focus states, and 44px targets.
- Added permalinks to 12 additional short-variant sections and reused the shared heading on Data, Help, and Stats pages, preserving existing IDs and conditional content.
- Updated Data's table of contents for the migrated headings and prevented repeated smooth-scroll restarts during initial alignment.
- Made copy success/failure feedback visible and accessible, including on vertically panned mobile viewports; viewport listeners run only while notifications exist.
- Tightened the existing hash-settling cleanup and fixed it overriding browser Back or a newer/cleared fragment.
- Rebased onto `main` (`778c066e6`) and resolved the Help snapshot conflict while retaining main's styling/content.

## Local validation

- [x] Manually reviewed the UI across desktop and narrow/mobile layouts, including Chrome device emulation.
- [x] Formatting, ESLint, Stylelint, TypeScript, and production build passed.
- [x] Full browser suite: **72 suites / 1,819 tests / 1,144 snapshots passed**.
- [x] **23 Chromium checks passed**, including mobile copy success/failure across all four variant families and Back/new-fragment navigation during settling.
- [x] Added regression tests alongside the fixes; negative controls reproduced the defects before correction, and independent browser checks passed afterward.

Jest was serialized with `--runInBand`; no assertions or timeouts were weakened. The intermittent parallel Data-test timeout and scaled-viewport horizontal clipping remain separate follow-ups. These checks do not claim physical-device, Safari/WebKit, or screen-reader coverage.

Reviewed candidate: `a3e37a1a6d52770401d22ea2e5ac18180be9d984`. The work was reviewed using `coding-standards` and `review-changes-before-committing`; this summary follows the concise drafting guidance from `draft-pull-request-description` without replacing the original PR description.

rileyhgrant (approved):
LGTM

PR #1927, opened 2026-08-19
feat(API): move compression to NGINX and add GCP trace ids for log correlation

nadeaujoshua:
There are 2 updates in this PR:
1. The compression of the API responses were moved from Node to NGINX. This change will help cut down response latency and CPU usage.
2. We are now extracting the GCP trace id from incoming request headers and attaching it to our logs to allow for trace correlation.

The first change is relatively small, while the second has a few moving parts.
- In NGINX, extract the trace id from an incoming request. Use it in the access log by setting it to the `logging.googleapis.com/trace` field
- Separately, use helpers in Node to try to use the trace id from any incoming request and pass it to the Node logger
- Inject the trace id in any subsequent logs in Node
- NGINX and Node logs that originate from the same request can now be associated with each other using the trace id

The logging update is optional, so deployments outside of GCP will not be negatively affected.

### DIFF
--- a/deploy/dockerfiles/browser/browser.dockerfile
+++ b/deploy/dockerfiles/browser/browser.dockerfile
@@ -42,5 +42,5 @@ COPY deploy/dockerfiles/browser/browser.proxy_cache.conf /etc/nginx/browser.prox
 COPY deploy/dockerfiles/browser/browser.nginx.conf /etc/nginx/browser.nginx.conf.template
 
 CMD REAL_IP_CONFIG=$([ -z "${PROXY_IPS:-}" ] || echo "$PROXY_IPS" | awk 'BEGIN { RS="," } { print "set_real_ip_from " $1 ";" }') \
-  envsubst "\$API_URL \$REAL_IP_CONFIG" < /etc/nginx/browser.nginx.conf.template > /etc/nginx/conf.d/default.conf && \
+  envsubst "\$API_URL \$REAL_IP_CONFIG \$GCP_PROJECT" < /etc/nginx/browser.nginx.conf.template > /etc/nginx/conf.d/default.conf && \
   nginx -g "daemon off;"
--- a/deploy/dockerfiles/browser/browser.nginx.conf
+++ b/deploy/dockerfiles/browser/browser.nginx.conf
@@ -1,23 +1,40 @@
-# Extract W3C trace ID when available.
+# Pass in the W3C traceparent header ($http_traceparent).
+# Try to extract a valid 32-character trace ID and set it as
+# $traceparent_trace_id.
+# Matches: <2 hex>-<32 hex>-<16 hex>-<2 hex>.
+# Capture group 1 (inside parentheses) is the trace ID.
+# If there is no valid match, the map value is empty.
 map $http_traceparent $traceparent_trace_id {
     default "";
     "~*^[0-9a-f]{2}-([0-9a-f]{32})-[0-9a-f]{16}-[0-9a-f]{2}$" $1;
 }
 
-# Fall back to Google's legacy trace context.
+# Pass in Google's legacy trace context header ($http_x_cloud_trace_context).
+# Try to extract a valid 32-character trace ID and set it as $cloud_trace_id.
+# Matches: <32 hex> followed by "/" or the end of the string.
+# Capture group 1 (inside parentheses) is the trace ID.
+# If there is no valid match, the map value is empty.
 map $http_x_cloud_trace_context $cloud_trace_id {
     default "";
     "~*^([0-9a-f]{32})(?:/|$)" $1;
 }
 
-# Prefer traceparent, then X-Cloud-Trace-Context.
+# Combine the trace IDs extracted above, preferring the W3C trace ID when both
+# formats are present.
+# The combined value has the form "<traceparent-trace-id>:<cloud-trace-id>".
+# Match either a W3C trace ID before the colon or, if absent, a legacy
+# Cloud Trace ID after the colon. Set the result as $gcp_trace_id.
 map "$traceparent_trace_id:$cloud_trace_id" $gcp_trace_id {
     default "";
     "~^([0-9a-f]{32}):" $1;
     "~^:([0-9a-f]{32})$" $1;
 }
 
-# Only construct a GCP trace resource when GCP_PROJECT is configured.
+# Combine GCP_PROJECT and the extracted trace ID.
+# Only construct a complete GCP trace resource when both the project 
+# and trace ID are available.
+# The resulting value is suitable for
+# logging.googleapis.com/trace: "projects/<project>/traces/<trace-id>".
 map "$GCP_PROJECT:$gcp_trace_id" $gcp_trace {
     default "";
     "~^([^:]+):([0-9a-f]{32})$" "projects/$1/traces/$2";
--- a/deploy/dockerfiles/browser/browser.nginx.conf
+++ b/deploy/dockerfiles/browser/browser.nginx.conf
@@ -1,7 +1,33 @@
+# Extract W3C trace ID when available.
+map $http_traceparent $traceparent_trace_id {
+    default "";
+    "~*^[0-9a-f]{2}-([0-9a-f]{32})-[0-9a-f]{16}-[0-9a-f]{2}$" $1;
+}
+
+# Fall back to Google's legacy trace context.
+map $http_x_cloud_trace_context $cloud_trace_id {
+    default "";
+    "~*^([0-9a-f]{32})(?:/|$)" $1;
+}
+
+# Prefer traceparent, then X-Cloud-Trace-Context.
+map "$traceparent_trace_id:$cloud_trace_id" $gcp_trace_id {
+    default "";
+    "~^([0-9a-f]{32}):" $1;
+    "~^:([0-9a-f]{32})$" $1;
+}
+
+# Only construct a GCP trace resource when GCP_PROJECT is configured.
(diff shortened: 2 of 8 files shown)


deploy/dockerfiles/browser/browser.dockerfile
```diff
@@ -42,5 +42,5 @@ COPY deploy/dockerfiles/browser/browser.proxy_cache.conf /etc/nginx/browser.prox
 COPY deploy/dockerfiles/browser/browser.nginx.conf /etc/nginx/browser.nginx.conf.template
 
 CMD REAL_IP_CONFIG=$([ -z "${PROXY_IPS:-}" ] || echo "$PROXY_IPS" | awk 'BEGIN { RS="," } { print "set_real_ip_from " $1 ";" }') \
-  envsubst "\$API_URL \$REAL_IP_CONFIG" < /etc/nginx/browser.nginx.conf.template > /etc/nginx/conf.d/default.conf && \
+  envsubst "\$API_URL \$REAL_IP_CONFIG \$GCP_PROJECT" < /etc/nginx/browser.nginx.conf.template > /etc/nginx/conf.d/default.conf && \
```
bpblanken, line 45: what happens if "GCP_PROJECT" is null here?

deploy/dockerfiles/browser/browser.nginx.conf
```diff
@@ -1,7 +1,33 @@
+# Extract W3C trace ID when available.
+map $http_traceparent $traceparent_trace_id {
+    default "";
+    "~*^[0-9a-f]{2}-([0-9a-f]{32})-[0-9a-f]{16}-[0-9a-f]{2}$" $1;
```
bpblanken: these regexes are challenging to interpret.  can we add a comment describing what they're trying to do?

graphql-api/src/app.ts
```diff
@@ -12,8 +11,36 @@ import { requestStore } from './request-context'
 import { closeCache } from './cache'
 import { loadWhitelist } from './whitelist'
 
+// Extract the trace ID from W3C or GCP trace context.
+const getGcpTraceId = (request: any) => {
+  // Prefer W3C Trace Context.
+  const traceParent = request.get('traceparent')
+
+  if (traceParent) {
+    const match = traceParent.match(
```
bpblanken: do we need a second regex check here if we're already running one in nginx?

graphql-api/src/config.ts
```diff
@@ -48,6 +48,9 @@ const config: Record<string, any> = {
   JSON_CACHE_LARGE_GENES: env.JSON_CACHE_LARGE_GENES === 'true' || false,
   JSON_CACHE_COMPRESSION: env.JSON_CACHE_COMPRESSION === 'true' || false,
   REDIS_CACHE_PREFIX: env.REDIS_CACHE_PREFIX || '',
+
+  // GCP
+  GCP_PROJECT: env.GCP_PROJECT,
```
bpblanken, line 53: is GCP_PROJECT injected into from the infrastructure somewhere?

graphql-api/src/app.ts
```diff
@@ -27,12 +54,20 @@ app.get('/health/ready', (_req: any, res: any) => {
 })
 
 app.use((req: any, res: any, next: any) => {
+  const traceId = config.GCP_PROJECT
+    ? getGcpTraceId(req)
```
bpblanken, line 55: can probably just do 
```
const trace = config.GCP_PROJECT ? `projects/${config.GCP_PROJECT}/traces/${getGcpTraceId(req)}`
```

deploy/dockerfiles/browser/browser.nginx.conf
```diff
@@ -48,18 +74,38 @@ server {
   ###############
 
   location /api/ {
+    gzip on;
```
bpblanken, line 94: One other potential thing to think about here, are the bodies from the express server going into the nginx cache compressed?  If they are... our cache will hold a lot fewer responses and we'll lose performance that way.

bpblanken (commented):
I think we should also look at dropping ` compression` and `@types/compression` from the `package.json`.

graphql-api/src/app.ts
```diff
@@ -12,8 +11,36 @@ import { requestStore } from './request-context'
...
+      /^[\da-f]{2}-([\da-f]{32})-[\da-f]{16}-[\da-f]{2}$/i
+    )
+
+    if (match) {
+      return match[1]
+    }
+  }
+
+  // Fall back to Google's legacy trace context header.
+  const cloudTrace = request.get('X-Cloud-Trace-Context')
```
bpblanken: can also maybe do 

```
const match = request.get('X-Cloud-Trace-Context')?.match(/^([\da-f]{32})(?:\/|$)/i)
return match?.[1]
``` 

here, the `?.` is nice sugar.

deploy/dockerfiles/browser/browser.dockerfile
nadeaujoshua: In Node, the config value will be `undefined`, and will result in `null` [here](https://github.com/broadinstitute/gnomad-browser/pull/1927/changes#diff-b2412f99024b1088cddaaa0c9f0d927b283124bcb164cf0c40199393d43010a4R57-R59).

In NGINX, the value will be an empty string by the `map` blocks, and the result will be `"logging.googleapis.com/trace":""` in the access log.

graphql-api/src/app.ts
nadeaujoshua: Yeah, because the NGINX and Node logs are independent given the same source request.

graphql-api/src/config.ts
nadeaujoshua: It is provided in our deployments repo. So for now, it will live in [this file](https://github.com/broadinstitute/gnomad-deployments/blob/main/gnomad-browser/base/browser.deployment.yaml).

Later, it will live in Terraform.

graphql-api/src/app.ts
nadeaujoshua: The function call can also return `null`, in which case it'd probably be better to set the entire value for the trace field to `null`.

deploy/dockerfiles/browser/browser.nginx.conf
nadeaujoshua: You're right that NGINX will now cache the uncompressed responses, meaning fewer responses will fit in the cache.

I think the trade-off is worth it, given a few things:
- The big problem for a while has been requests backing up in Node. Testing showed average cache-miss latency improved over 40%, with even larger improvements in the tail. So we’d be freeing up CPU in Node to handle more requests
- We currently have substantial cache capacity that we don’t use, so I’d rather use more of it to reduce CPU pressure and improve cache-miss latency
- It is cheap and easy to bump up NGINX cache disk allocation to enable larger response sizes, if necessary

Ultimately, reducing the latency tail and 500s is more valuable than maximizing the cache-hit rate, especially given the current cache utilization.

deploy/dockerfiles/browser/browser.nginx.conf
nadeaujoshua: Good point, will do.

deploy/dockerfiles/browser/browser.nginx.conf
bpblanken: Excellent response!  Thank you.

a few more questions:
-how are we measuring the cache size?
- do we need to bump the kubernetes cpu limit for the nginx container?

nadeaujoshua: No problem, for cache size, it's specified in the NGINX config as: `proxy_cache_path /var/cache/nginx levels=1:2 keys_zone=gnomadCache:30m max_size=20g inactive=30d;`

In the K8s deployment, we have an ephemeral disk allocation for 20g mounted to that path.

As for the cpu, I think we're ok keeping it for now. When I tested, which was in Cloud Run, I did so with the same specs, and peak utilization was 16%, with the average at about 10% under load. Currently, the K8s historical peak was 0.013 of the max 0.25, with it usually being under 0.01.

graphql-api/src/app.ts
nadeaujoshua: Sure, change the made in 34e150a.

deploy/dockerfiles/browser/browser.nginx.conf
nadeaujoshua: Added some comments in both NGINX and Node for the regexes here: cfcdb1e

graphql-api/src/app.ts
```diff
@@ -12,8 +11,34 @@ import { requestStore } from './request-context'
...
+
+    if (match) {
+      return match[1]
+    }
+  }
+
+  // Fall back to Google's legacy trace context header.
+  // Matches: <32 hex> followed by "/" or the end of the string.
+  // Capture group 1 (inside parentheses) is the 32-character trace ID.
+  const match = request.get('X-Cloud-Trace-Context')?.match(/^([\da-f]{32})(?:\/|$)/i)
```
bpblanken: You can make use of the same pattern in the block above too! (sorry to be pedantic!)

bpblanken (approved):
one trailing style nit but otherwise I think this is great!

graphql-api/src/app.ts
nadeaujoshua: Ah, regexes are different because the first pattern is `<2 hex>-<32 hex>-<16 hex>-<2 hex>`.
While the second is just `<32 hex>` with optional trailing `/`.

bpblanken: ah I meant the `?.` pattern.

PR #1929, opened 2026-08-25
feat(frontend): update markdown import code styling

rileyhgrant:
Update markdown styles to include have inline code and code blocks be fenced in a grey background, a la `this` and 

```
this
```

Makes the help text for hail tables more legible, imo.

Smol low stakes frontend only PR, testing out git worktrees (and trying to more regularly make time for this very smol scoped kind of thing)

---

Screenshot 2026-08-25 at 17 20 58

Screenshot 2026-08-25 at 17 21 06

### DIFF
--- a/browser/src/MarkdownContent.tsx
+++ b/browser/src/MarkdownContent.tsx
@@ -22,7 +22,7 @@ const MarkdownContentWrapper = styled.div`
   p {
     margin-top: 15px;
     margin-bottom: 15px;
-    line-height: 1.4;
+    line-height: 1.6;
   }
 
   a {
@@ -41,6 +41,34 @@ const MarkdownContentWrapper = styled.div`
     line-height: 1.4;
   }
 
+  code {
+    padding: 0.15em 0.35em;
+    border: 1px solid rgba(0, 0, 0, 0.1);
+    border-radius: 0.3em;
+    background: #eef0f3;
+    font-family: 'SFMono-Regular', Consolas, 'Liberation Mono', Menlo, monospace;
+    font-size: 0.9em;
+  }
+
+  pre {
+    overflow-x: auto;
+    max-width: 100%;
+    padding: 1em;
+    border: 1px solid rgba(0, 0, 0, 0.1);
+    margin: 1.25em 0;
+    border-radius: 0.4em;
+    background: #eef0f3;
+    line-height: 1.5;
+  }
+
+  pre code {
+    padding: 0;
+    border: 0;
+    border-radius: 0;
+    background: none;
+    font-size: 0.875em;
+  }
+
   ol,
   ul {
     padding-left: 20px;
@@ -49,6 +77,7 @@ const MarkdownContentWrapper = styled.div`
 
   li {
     margin-bottom: 0.5em;
+    line-height: 1.6;
   }
 
   table {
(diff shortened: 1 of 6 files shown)


rileyhgrant:
Tagging @ch-kr in for review also since this is a change in formatting from what's currently on the browser, so seeking a yes/no from you.

Affects several help text pages (any place where we have text formatted as code). This doesn't affect other pages, as this change is just on html is generated from markdown files (help text, faqs, about, etc).

This is super minor, small, and took like zero effort. If this is a no, then no big deal at all to just close this PR and not merge it.

ch-kr:
sorry I saw the notification for this earlier but was heads down on another task. I like this new change! one minor comment is that I'd like to remove some extra quotes in the text when this formatting change goes in. for example, from your screenshot above:
image
the quotes around `freq` now seem unnecessary

edit to also add something I should have included in my earlier response: thank you for working on this, and I'm excited at the prospect of having your attention regularly on some small gnomad tasks :)

browser/src/MarkdownContent.tsx
```diff
@@ -41,6 +41,34 @@ const MarkdownContentWrapper = styled.div`
...
+    border: 1px solid rgba(0, 0, 0, 0.1);
+    border-radius: 0.3em;
+    background: #eef0f3;
+    font-family: 'SFMono-Regular', Consolas, 'Liberation Mono', Menlo, monospace;
+    font-size: 0.9em;
+  }
+
+  pre {
+    overflow-x: auto;
+    max-width: 100%;
```
bpblanken: Out of curiosity, does `max-width` include padding?  Or is the `padding: 1em` added outside of the original "maximum width box"?

browser/src/MarkdownContent.tsx
```diff
@@ -41,6 +41,34 @@ const MarkdownContentWrapper = styled.div`
     line-height: 1.4;
   }
 
+  code {
+    padding: 0.15em 0.35em;
+    border: 1px solid rgba(0, 0, 0, 0.1);
+    border-radius: 0.3em;
+    background: #eef0f3;
```
bpblanken: we could maybe use a css variable for this?  Something like ```--code-background```?  Do we use those elsewhere?

browser/src/MarkdownContent.tsx
rileyhgrant: In this case since its a `pre` element (which is a box), the `padding` is inside the box, not added to the max width box.

With the padding:
Screenshot 2026-09-01 at 16 57 31

No padding:
Screenshot 2026-09-01 at 16 57 42

Edit:

Oh, huh. I guess thats really happening because of the default `width: auto` that's styled on the `pre` element, and `max-width` is actually never used because `width: auto` of the whole thing, - padding is always <100%.

Will remove that max-width. CSS drives me insane.

browser/src/MarkdownContent.tsx
rileyhgrant: Yep, totally could. Good call.

Not sure how many places we use them, probably worth it in a few places. I feel like gnomAD's background color being a very slightly not white color (`#fafafa`) causes a lot of tiny differences in color in components, icons on gnomAD pages, presentations with gnomAD screenshots, etc.

bpblanken (approved):

PR #1932, opened 2026-08-27
Improve ClinVar testing

rileyhgrant:
This PR adds additional snapshot and playwright tests surrounding ClinVar, and handles a full page crash on a rare but possible circumstance in production.

It could be reviewed as a single big diff, but I believe commit by commit using GitHub's functionality for that will be a far more clear way to review it, and didn't want to open 4 stacked PRs all around testing ClinVar. Each individual commit is quite small and scoped, could have been a seperate PR if we really want, but I personally find the github option to view a single commit at a time pretty useful, and think these commits are all quite related to each other. 

This is in prep for further changes to local development of the frontend/backend without the need for a tunnel to ES.

- Handle an existing crash where inability to get ClinVar variants (network failure, bad indice) would crash the existing gene page
- minor refactor to reduce duplication of ClinVar section rendering
- Add more snapshot tests to handle ClinVar with better synthetic data, in all of its new possible states (clinvar failed to load, clinvar loaded but no variants, clinvar loaded and has variants)
- Add playwright test to confirm behavior against chosen backend (live api or local backend), not synthetic mocked data, to help with adding confidence when we move clinvar to parquet

### DIFF
--- a/browser/src/ClinvarVariantsTrack/ClinvarVariantTrack.tsx
+++ b/browser/src/ClinvarVariantsTrack/ClinvarVariantTrack.tsx
@@ -295,7 +295,7 @@ const UnmemoizedClinvarVariantTrack = ({ referenceGenome, transcripts, variants 
 export const ClinvarVariantTrack = React.memo(UnmemoizedClinvarVariantTrack)
 
 type ClinvarVariantsProps = {
-  clinvarVariants: ClinvarVariant[]
+  clinvarVariants: ClinvarVariant[] | null | undefined
   clinvarReleaseDate: string
   referenceGenome: ReferenceGenome
   transcripts: Transcript[]
@@ -311,6 +311,27 @@ const ClinvarVariants = ({
   zoomRegion = null,
   pageType,
 }: ClinvarVariantsProps) => {
+  const heading = (
+    <TrackPageSection>
+      <h2>ClinVar variants</h2>
+    </TrackPageSection>
+  )
+
+  const clinvarUnavailableMessage = (
+    <TrackPageSection as="p">
+      ClinVar variants could not be loaded. Other data on this page is unaffected.
+    </TrackPageSection>
+  )
+
+  if (!clinvarVariants) {
+    return (
+      <>
+        {heading}
+        {clinvarUnavailableMessage}
+      </>
+    )
+  }
+
   const clinvarVariantTrack = (
     <>
       <ClinvarVariantTrack
@@ -330,9 +351,7 @@ const ClinvarVariants = ({
 
   return (
     <>
-      <TrackPageSection>
-        <h2>ClinVar variants</h2>
-      </TrackPageSection>
+      {heading}
       {clinvarVariants.length > 0 ? clinvarVariantTrack : noClinvarVariantsFoundMessage}
     </>
   )
--- a/browser/src/ClinvarVariantsTrack/ClinvarVariantTrack.tsx
+++ b/browser/src/ClinvarVariantsTrack/ClinvarVariantTrack.tsx
@@ -5,7 +5,8 @@ import { Track } from '@gnomad/region-viewer'
 import { Button, Checkbox, Modal, ExternalLink } from '@gnomad/ui'
 import CategoryFilterControl from '../CategoryFilterControl'
 import InfoButton from '../help/InfoButton'
-import { TrackPageSection } from '../TrackPage'
+import filterVariantsInZoomRegion from '../RegionViewer/filterVariantsInZoomRegion'
+import { PageType, TrackPageSection } from '../TrackPage'
 import {
   VEP_CONSEQUENCE_CATEGORIES,
   VEP_CONSEQUENCE_CATEGORY_LABELS,
@@ -21,8 +22,10 @@ import {
 import ClinvarAllVariantsPlot from './ClinvarAllVariantsPlot'
 import ClinvarBinnedVariantsPlot from './ClinvarBinnedVariantsPlot'
 import ClinvarVariantDetails from './ClinvarVariantDetails'
+import formatClinvarDate from './formatClinvarDate'
 import { ClinvarVariant } from '../VariantPage/VariantPage'
 import { Transcript } from '../TranscriptPage/TranscriptPage'
+import { ReferenceGenome } from '../../../dataset-metadata/metadata'
 
 const TopPanel = styled.div`
   display: flex;
@@ -98,7 +101,7 @@ type Props = {
   variants: ClinvarVariant[]
 }
 
-const ClinvarVariantTrack = ({ referenceGenome, transcripts, variants }: Props) => {
+const UnmemoizedClinvarVariantTrack = ({ referenceGenome, transcripts, variants }: Props) => {
   const [selectedVariant, setSelectedVariant] = useState(null)
 
   const [includedClinicalSignificanceCategories, setIncludedClinicalSignificanceCategories] =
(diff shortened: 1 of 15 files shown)


browser/src/ClinvarVariantsTrack/ClinvarVariantTrack.tsx
```diff
@@ -289,4 +292,50 @@ const ClinvarVariantTrack = ({ referenceGenome, transcripts, variants }: Props)
   )
 }
 
-export default React.memo(ClinvarVariantTrack)
+export const ClinvarVariantTrack = React.memo(UnmemoizedClinvarVariantTrack)
+
+type ClinvarVariantsProps = {
+  clinvarVariants: ClinvarVariant[]
+  clinvarReleaseDate: string
```
bpblanken, line 299: Would it be over-the-top to do something like:

```
type DateString = `${number}${number}${number}${number}-${number}${number}-${number}${number}`;
```

browser/src/ClinvarVariantsTrack/ClinvarVariantTrack.tsx
```diff
@@ -289,4 +292,50 @@ const ClinvarVariantTrack = ({ referenceGenome, transcripts, variants }: Props)
...
+  referenceGenome: ReferenceGenome
+  transcripts: Transcript[]
+  zoomRegion?: { start: number; stop: number } | null
```
bpblanken, line 302: Likely an out of scope question, but is there a reason this type doesn't include "chromosome"... sort of just behaving as a "region" type?

browser/src/GenePage/MitochondrialVariantsInGene.tsx
```diff
@@ -132,24 +130,14 @@ const MitochondrialVariantsInGene = ({ datasetId, gene, zoomRegion, ...rest }: P
...
-              <h2>ClinVar variants</h2>
-            </TrackPageSection>
-            {data.gene.clinvar_variants.length > 0 ? (
-              <>
-                <ClinvarVariantTrack
-                  referenceGenome={referenceGenome(datasetId)}
-                  transcripts={gene.transcripts}
-                  variants={filterVariantsInZoomRegion(data.gene.clinvar_variants, zoomRegion)}
-                />
-                <TrackPageSection as="p" style={{ margin: 0 }}>
```
bpblanken, line 145: Is the `style={{ margin: 0 }}` still necessary?  Can it be passed in as an additional property?

bpblanken (commented):
Just the first commit

browser/src/VariantList/annotateVariantsWithClinvar.ts
```diff
@@ -1,6 +1,20 @@
-const annotateVariantsWithClinvar = (variants: any, clinvarVariants: any) => {
-  const clinvarInfo = new Map()
-  clinvarVariants.forEach((clinvarVariant: any) => {
+import { ClinvarVariant } from '../VariantPage/VariantPage'
+
+type ClinvarAnnotation = {
+  clinical_significance?: string
```
bpblanken, line 4: are we stuck with snake case from the backend?

bpblanken:
The other three commits looked good to me though!

browser/src/ClinvarVariantsTrack/ClinvarVariantTrack.tsx
rileyhgrant: Hmm, I think that particular type would allow something like

`133747381247839210497-4378292478239-23748329` to be a valid `DateString`. 

But apart from me being pedantic (sorry), thank you for the catch and thinking about safety!

I think the way to get some safety here would be to add a guard clause in the `formatClinVarDate` helper that's responsible for parsing the string received, and have a fallback value there in case the incoming string is malformed. Will add the fallback value real quick

browser/src/ClinvarVariantsTrack/ClinvarVariantTrack.tsx
rileyhgrant: Yeah, good question. Pretty much the region viewer context provider (and all related components e.g. the zoomable region viewer these props get handed to) don't support rendering tracks across multiple chromosomes a la a manhattan plot. Those components are used to correctly scale the given data across the X axis of the plots allotted space on the page.

Not exactly sure of the original reasoning going into that, but at the gene/transcript level, notion of a chromosome is never needed for setting the x scale in rendering tracks/components. Perhaps it could come up in rendering arbitrary regions over abitrary chromosomes, but not sure when exactly it's helpful to actually view arbitrary regions across arbitrary chromosomes (there's probably some case where it is useful, idk). Anyhoo, that's just me thinking out loud, there are probably better answers.

browser/src/GenePage/MitochondrialVariantsInGene.tsx
rileyhgrant: This I'm pretty darn sure was just a mistake when the original clinvar track was copy pasta-d around. None of the rest of the places have it, and the margin here serves to make the 'clinvar release date XXXX' text very slightly closer to the clinvar track on this one page than it is on the other pages. Looking at the mito gene page, there is/was no clear reason for this, and its purely a small css styling difference.

I opted to just remove that difference as a judgement call and make it consistent across all occurences. Happy to put it back in if we feel strongly it should still be there.

rileyhgrant:
Okie, I think I addressed your comments, and I added two little fixup commits to address one comment. Ready for re-review, at your convenience.

browser/src/ClinvarVariantsTrack/formatClinvarDate.ts
```diff
@@ -1,10 +1,14 @@
...
-  const [year, month, day] = dateString.split('-').map(Number)
-  const date = new Date(year, month - 1, day)
-  return dateFormatter.format(date)
+const formatClinvarDate = (dateString: string): string => {
+  try {
+    // Dates in ClinVar date are formatted YYYY-MM-DD
+    const [year, month, day] = dateString.split('-').map(Number)
+    const date = new Date(year, month - 1, day)
+    return dateFormatter.format(date)
+  } catch (error) {
```
bpblanken: do we want to reject rollover here?

things like 
```
new Date(2023, 17, 0)
```

or
```
new Date(2023, 1, 37)
```
just work.

browser/src/ClinvarVariantsTrack/ClinvarVariantTrack.tsx
bpblanken: yeah that's reasonable.  A nice strong guard in the `formatClinVarDate` seems reasonable.

bpblanken:
(I clicked the wrong button 😮‍💨)

browser/src/ClinvarVariantsTrack/formatClinvarDate.ts
rileyhgrant: Woopsies, dang yeah good catch. I made assumptions about how `Date` worked, gah. Quick fixup incoming.

rileyhgrant:
Okay, added a lil commit that includes some tested datechecking. My gut feel is this is overkill, but also it was straightfoward for an LLM to extrude, is easy to read (imo), and has tests that give me confidence.

Happy to change anything, re-requested review.

browser/src/VariantList/annotateVariantsWithClinvar.ts
rileyhgrant: Yeah, unfortunately. We get that snake case from the API, and use that snake case in the table on the variant page, amongst other places.

Along with other initiatives I'm hoping to do as part of efforts to tidy up the browser, is handle this and parse the data when we get it right back from the API into strongly typed data that we use throughout the frontend a la 'Parse, don't validate'.

bpblanken (approved):

PR #1950, opened 2026-09-08
fix(frontend): make age distribution permalinks accessible

mattsolo1:
## Summary
Follow-up to #1925, **targeting `kc/age-distribution-link`, not `main`**. Contains only four fix commits on KC's current PR head; no unrelated main-branch commits.

- Make Age Distribution links keyboard accessible, directly discoverable from their gutter, and usable on narrow/touch screens.
- Keep copy feedback inside the viewport; use a persistent polite announcement region with separate error alerts.
- Preserve native fragment navigation and handle unavailable/denied clipboard access.
- Clean up hash-settling frames/listeners on reader input, timeout and unmount.
- Trim redundant comments and strengthen unit/browser regressions, retaining homepage navigation coverage.

The additional variant-section links are intentionally a separate PR, #1951, stacked on this branch. Keep this branch until that follow-up is retargeted.

## Validation
- Before restacking: 1,792 browser tests / 1,144 snapshots and 3 Chromium regressions passed; lint, Stylelint and production build passed.
- After cherry-picking onto KC's exact head, the combined fixes + stacked expansion pass 1,811 browser tests / 1,144 snapshots and all 5 Chromium regressions.
- Independent code review found no blockers.
- Typecheck on the implementation tree encountered the existing two missing `@google-cloud/storage` imports, not changed-file errors.

Human screen-reader speech-output and physical-device touch verification remain outstanding. A broader mitochondrial mobile-overflow/emulation issue was recorded separately; this PR does not claim to resolve it.

### DIFF
--- a/browser/src/AnchorLink.spec.tsx
+++ b/browser/src/AnchorLink.spec.tsx
@@ -1,10 +1,11 @@
 import React from 'react'
-import { render } from '@testing-library/react'
+import { render, screen } from '@testing-library/react'
 import userEvent from '@testing-library/user-event'
 import { afterEach, beforeEach, describe, expect, jest, test } from '@jest/globals'
 import styled from 'styled-components'
 
-import { AnchoredSectionHeading, withAnchor } from './AnchorLink'
+import { AgeDistributionHeading, withAnchor } from './AnchorLink'
+import Notifications from './Notifications'
 
 const PlainHeading = withAnchor(styled.h2``)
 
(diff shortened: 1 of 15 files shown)


rileyhgrant (approved):
LGTM!

I didn't in depth review the base branch this targets (kc's) as I wanted requested on that. Happy to do that too, if we want. In general this looks good to me.

---

Just to chime in support of getting shared context into the gnomAD browser repo if we want outside contributors to be able to submit something more to our standards, I'd encourage (if this was submitted with assistance from an agent) asking the agent what files in its context made the difference in terms of a few things:

- The PR description, I like this format of a concise high level summary, and a clear listing of the validation criteria that were checked
- Accessibility as a core part of the browser
- Amount and conciseness of comments. The ones here seemed reasonable to me
- The series of commits made

Those all stand out as a few things that were changed in this PR, are more in line with our standards, and that were different in the original PR. If we can get those into shared context files (and our PR submission template) and they make a difference, I think we should continue to try to do so.

PR #1953, opened 2026-09-09
bugfix(api): disable GraphiQL in production, preventing HTML responses

### LINKED ISSUE #1949: API response served GraphQLHTML interface
It was reported to the team that the BRCA1 gene page loaded with `Unable to load gene` in the main page, as seen here: https://the-tgg.slack.com/archives/C03P7FA3W3T/p1788879176116889

After checking, the API response status was 200, but the payload was the GraphQL HTML for the interactive interface.

Checking the networking diagnostics, we saw the the response was an NGINX cache hit, meaning this HTML was served from the API to NGINX, which was then cached for this given query.

To resolve the immediate issue for users, Riley cleared the NGINX cache, and reloaded NGINX, which resolved the issue.

Now, we need to figure out why the API responded this way, and/or why NGINX cached this HTML as opposed to the correct data.

bpblanken:

resolves #1949

### Merge strategy

### What does this PR do?

disables GraphiQL in production, preventing html responses from being generated (in addition to not cached!) by the API.

### Tests

It IS possible to add a test for this, but I elected not to.  I didn't think it had enormous value other than validating that we disabled the strange library behavior.

### Verification

- [x] I have the relevant `make` targets to run CI checks locally and verified that they succeed (e.g. `make validate-browser` if you made changes in the `browser/` directory, or `make validate-all` if you're unsure which checks to run)
- [x] (If UI) I have tested this across desktop and mobile breakpoints.
- [x] (If a LLM/AI agent runner was used, e.g. OpenCode, Aider, or ClaudeCode) I have instructed the agent runner to load the `AGENTS.md` file into its context, and verified the the required `Assisted-by` trailer is on each commit.

### DIFF
--- a/graphql-api/README.md
+++ b/graphql-api/README.md
@@ -9,13 +9,6 @@ Please note that this resource is under development and the query schema is subj
 
 The gnomAD API uses [GraphQL](https://graphql.org/learn/) for queries.
 
-To get started, open the interactive query editor at https://gnomad.broadinstitute.org/api.
-
-Click the "Docs" button in the top right-hand corner to open up the Documentation Explorer.
-GraphQL is self-documenting, so the fields and data described in this section are always
-up-to-date. Browsing through the Documentation Explorer is the best way to understand how
-to query data and learn which types of data are available to retrieve.
-
 ## Examples
 
 Examples of fetching gnomAD allele counts for a specific variant using different languages.
--- a/graphql-api/src/config.ts
+++ b/graphql-api/src/config.ts
@@ -20,6 +20,7 @@ const parseProxyConfig = (config: any) => {
 }
 
 const config: Record<string, any> = {
+  NODE_ENV: env.NODE_ENV || 'development',
   // Elasticsearch
   ELASTICSEARCH_URL: env.ELASTICSEARCH_URL,
   ELASTICSEARCH_USERNAME: env.ELASTICSEARCH_USERNAME,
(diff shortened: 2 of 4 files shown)


rileyhgrant (changes requested):
I'm on board with this, I agree that there's not much value to having the iGraphQL API in production, though it is a thing we have had, and some users might use (unfortunately I don't think we have numbers on that, but I'd confidently bet it's a low low amount of users)

Probably worth mentioning this in standup too, just to give a last chance to hear if there are any arguments to keep the iGraphQL UI around in prod.

Assuming we are all fine with removing this (which is my vote too), I think we should also remove the text on the data page directing users to the iGraphQL interactive UI:

https://github.com/broadinstitute/gnomad-browser/blob/05efc2ed166112e8076718efecde006925f32acc/browser/src/DataPage/GraphQLDocs.tsx#L38-L42

---

One additional thought, non-blocking.

- For consistency's sake, might be worth it to disable iGraphQL in the seperate reads api too?
https://github.com/broadinstitute/gnomad-browser/blob/05efc2ed166112e8076718efecde006925f32acc/reads/src/server.js#L45

bpblanken:
yeah, I didn't realize there was documentation for this 🤔 .  I've been on that page several times and did not notice it.

rileyhgrant (approved):
LGTM, with the caveat of I probably want another opinion on removing this. Though it's quite easy to just add back if someone does use and want it and makes a strong case, so a third opinion from someone else is not blocking.

bpblanken:
I made a ticket for a potential [doc drift check](https://github.com/broadinstitute/gnomad-browser/issues/1955) inspired by the review here.

nadeaujoshua (approved):

PR #1954, opened 2026-09-09
feat(browser): make variant table search hotkeys discoverable

### LINKED ISSUE #1679: Update ctrl+f functionality in gnomAD browser

bpblanken:

closes: #1679

### Merge strategy

### What does this PR do?

Several respondents to the 2024 user survey asked that Ctrl+F/Cmd+F search all variants in a gene page's table, not just the rows currently rendered — gnomAD's variant tables are virtualized/paginated, so the browser's native find can't see off-screen rows. This PR makes the existing "/" search hotkey discoverable via a badge + tooltip next to each variant table's search box, wires that hotkey up consistently across all four variant table types, and adds Ctrl+F/Cmd+F as an alternate trigger that falls back to the browser's native find on a second press.

### Architecture & Design

Reused the existing `/` hotkey mechanism (`mousetrap` via `@gnomad/ui`'s `KeyboardShortcut`) rather than trying to fully override native find. Binding `mod+f` alongside `/` works because Mousetrap's default `stopCallback` skips a bound handler whenever the keydown target is already an input/textarea/select. So the first Ctrl+F/Cmd+F (focus elsewhere on the page) is intercepted and focuses the search box, and a second press (once that input already has focus) is not intercepted and falls through to the browser's native find — no press-counting or timing logic needed. No new dependencies; `mousetrap`/`KeyboardShortcut` were already in use for `/`.

There are a few other consideration:
- Adding invisible table rows with just the text, and using Chrome's `beforematch` event combined with the hidden="until-found" attribute (shipped in Chromium ~2022, alongside content-visibility: auto).  This didn't work in Safari though....
- We likely want to paginate these in the future... so wiring up to the backend better would render an invisible rows solution to immediately be tech debt.

### Implementation

- Added a `Badge` ("/") with a tooltip ("Press / or Ctrl+F (⌘F on Mac) to search the variant table") next to each of the four variant tables' search inputs
- Wired up the "/" hotkey (previously only present on the SNV/indel and mitochondrial variant tables) to the structural and copy number variant tables as well, for consistency
- Added a `mod+f` (Ctrl+F / Cmd+F) `KeyboardShortcut` binding alongside `/` on all four tables
- Extracted the shared ref/handler/bindings into `browser/src/useSearchHotkeys.tsx`

### Tests

No new tests were added. This is a presentational and keybinding change; the existing search behavior (`onChange`/`searchText` plumbing) is unmodified, and the existing test suite passed unchanged (`make validate-browser`).

Screenshot 2026-09-09 at 9 13 31 AM

Browsers generally allow a page to intercept Ctrl+F/Cmd+F this way (Gmail, GitHub, and VS Code Web all do the same), but it isn't a formal spec guarantee — worth a manual check in Safari specifically, since it's historically been the pickiest about letting pages override this shortcut.

### Verification

- [x] I have the relevant `make` targets to run CI checks locally and verified that they succeed (`make validate-browser`)
- [ ] (If UI) I have tested this across desktop and mobile breakpoints.
- [x] (If a LLM/AI agent runner was used, e.g. OpenCode, Aider, or ClaudeCode) I have instructed the agent runner to load the `AGENTS.md` file into its context, and verified the the required `Assisted-by` trailer is on each commit.

### DIFF
--- a/browser/src/CopyNumberVariantList/CopyNumberVariantFilterControls.tsx
+++ b/browser/src/CopyNumberVariantList/CopyNumberVariantFilterControls.tsx
@@ -1,9 +1,10 @@
-import React, { useRef } from 'react'
+import React from 'react'
 import styled from 'styled-components'
 
-import { Badge, Checkbox, KeyboardShortcut, SearchInput } from '@gnomad/ui'
+import { Badge, Checkbox, SearchInput } from '@gnomad/ui'
 
 import CategoryFilterControl from '../CategoryFilterControl'
+import useSearchHotkeys from '../useSearchHotkeys'
 
 import { cnvTypes, cnvTypeColors } from './copyNumberVariantTypes'
 
@@ -56,7 +57,7 @@ type Props = {
 }
 
 const CopyNumberVariantFilterControls = ({ onChange, colorKey, value }: Props) => {
-  const searchInput = useRef(null)
+  const { searchInputRef, searchHotkeys } = useSearchHotkeys()
 
   return (
     <SettingsWrapper>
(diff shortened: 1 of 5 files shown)


rileyhgrant (approved):
lgtm!

Perused the code, and opened a local frontend to try it out myself.

Seems like a very reasonable way to go about this. I like the second `mod f` falling back to native browser find.

PR #1958, opened 2026-09-10
fix(api): relax parameters for the API readiness probe

nadeaujoshua:

N/A

### Merge strategy

### What does this PR do?

This PR relaxes the readiness probe for the API deployment.

Currently, our API deployments constantly fail readiness probes (hundreds of times a day), potentially causing performance problems. I believe that the readiness probe parameters are a contributing factor that can be adjusted to improve the situation.

Hypothetical sequence:

1. API pod A (of two) gets a surge of requests, including large ones, that can cause event-loop contention/delay. It's normal for some requests to take over 1 second
2. Probe sends a request to the same pod that lands in the "backlog" and it fails as the default timeout is 1 second
3. Probe waits 10 seconds. After 3 consecutive failures, K8s marks the pod unready

So heavier traffic can cause a pod to be temporarily taken out of service, potentially because the readiness probe didn't receive a response within such a tight window, not necessarily because there was something wrong with the API.

The potential implication is that if we're experiencing a surge of traffic that is evenly distributed across pods A and B, but this readiness probe failure takes pod A out of service, it temporarily doubles the workload for pod B. Overloads can worsen performance and even crash the pod, which seems to be supported by high restart counts in our API deployments.

### Architecture & Design

N/A

### Implementation

We currently only declare the initial delay at 3 seconds. The rest default to:
- periodSeconds = 10
- timeoutSeconds = 1  (very aggressive)
- failureThreshold = 3
- successThreshold = 1

The changes in this PR relax the timeout duration and timeout count threshold, but we're only increasing the entire probe lifecycle from roughly 30 seconds to roughly 50 seconds. This preserves relatively quick detection of an actual unhealthy pod while allowing more room for a temporary load increase.

### Tests

N/A

N/A

N/A

### Verification

N/A

### DIFF
--- a/deploy/manifests/browser/base/api.deployment.yaml
+++ b/deploy/manifests/browser/base/api.deployment.yaml
@@ -77,7 +77,9 @@ spec:
             httpGet:
               path: /health/ready
               port: http
-            initialDelaySeconds: 3
+            initialDelaySeconds: 5
             periodSeconds: 10
+            timeoutSeconds: 3
+            failureThreshold: 5
       nodeSelector:
         cloud.google.com/gke-nodepool: 'main-pool'
--- a/graphql-api/src/app.ts
+++ b/graphql-api/src/app.ts
@@ -45,6 +45,13 @@ app.set('trust proxy', config.TRUST_PROXY)
 // GCE load balancers require a 200 response from the health check endpoint, so this must be
 // registered before the HTTP=>HTTPS redirect middleware, which would return a 30x response.
 app.get('/health/ready', (_req: any, res: any) => {
+  const startAt = performance.now()
+  onFinished(res, () => {
+    logger.info({
+      event: 'healthCheck',
+      latencyMs: performance.now() - startAt
+    })
+  })
   res.send('ok')
 })
 
--- a/graphql-api/src/app.ts
+++ b/graphql-api/src/app.ts
@@ -49,7 +49,7 @@ app.get('/health/ready', (_req: any, res: any) => {
   onFinished(res, () => {
     logger.info({
       event: 'healthCheck',
-      latencyMs: performance.now() - startAt
+      latencyMs: performance.now() - startAt,
     })
   })
   res.send('ok')


deploy/manifests/browser/base/api.deployment.yaml
```diff
@@ -77,7 +77,9 @@ spec:
             httpGet:
               path: /health/ready
               port: http
-            initialDelaySeconds: 3
+            initialDelaySeconds: 5
             periodSeconds: 10
+            timeoutSeconds: 3
```
bpblanken, line 82: I think adding a `livenessProbe` to trigger a pod restart is also warranted here:

```
          livenessProbe:
            httpGet:
              path: /health/ready
              port: http
            initialDelaySeconds: 5
            periodSeconds: 10
            timeoutSeconds: 3
            failureThreshold: 10
```

bpblanken, line 82: Also curious if we need increases to both the timeout and the failure threshold... it's possible only single one is necessary.

rileyhgrant (approved):
I agree relaxing this seems like a good direction, approving because these numbers look reasonable to my eyes, fwiw!

More for my knowledge's sake than anything else, what stats, numbers, etc, will we be watching for once this hits production to see if its having the intended effect?

bpblanken:
One other suggestion:

We don't currently log how long health-checks are taking so our increase in the timeout is a guess here (there are many requests that take longer than 1s, but also many that take longer than 10s).  The only logs for these are:

```
"Readiness probe failed: Get "http://10.164.1.7:8000/health/ready": context deadline exceeded (Client.Timeout exceeded while awaiting headers)"
```

Maybe adding something like:
```
 app.get('/health/ready', (_req: any, res: any) => {
+  const startAt = performance.now()
+  onFinished(res, () => {
+    logger.info({ event: 'healthCheck', latencyMs: performance.now() - startAt })
+  })
   res.send('ok')
 })
```

bpblanken:
@rileyhgrant 

The number of matching logs for this should decrease: 

Screenshot 2026-09-10 at 12 39 19 PM

deploy/manifests/browser/base/api.deployment.yaml
nadeaujoshua: The liveness probe can be useful and we can probably add it elsewhere at a later time. Doing so in this update, to the same endpoint, would cause restarts when transient load is causing the readiness probe to flap.

nadeaujoshua: Both are useful for different things:
- Increasing the timeout increases the tolerance of transient event-loop contention
- Increasing the failure threshold prevents a small number of transient failures caused by backed-up requests from taking the pod out of service

bpblanken: Adding a liveness probe with larger thresholds wouldn't necessarily overlap though right?

bpblanken (approved):

bpblanken:
One other thing (not blocking but I think we might run into an issue):

The snippet I proposed above maybe doesn't work -  the `onFinished` is tied to the local healthcheck `res.send` and not the rest of the response.  I saw:

```
  SERVER LOG > {"event":"requestStart","url":"/block?ms=2500"}
  SERVER LOG > {"event":"requestEnd","url":"/block?ms=2500","latencyMs":2500.981}
  SERVER LOG > {"event":"healthCheck","latencyMs":0.29}
  CLIENT observed >  /health/ready took 2210.2 ms end-to-end
```
when testing locally.  Not totally sure of the solution but might be worth a second look.

deploy/manifests/browser/base/api.deployment.yaml
nadeaujoshua: Oh woops, yeah I didn't catch that. We can add that in the next iteration after confirming whether this update fixes the original issue.

PR #1961, opened 2026-09-17
fix(backend): restore API request and query-cost limits

mattsolo1:

No GitHub issue; discovered during the bot-protection audit.

### Merge strategy

### What does this PR do?

Restore the API's per-IP request and query-cost quotas. The current implementation resolves an ioredis `[error, result]` tuple rather than its numeric counter, so successful Redis replies silently bypass both numeric limit comparisons.

This restores protection for GraphQL executions that pass edge controls; it does not govern static assets or requests served from cache, and is separate from the ShapBot-specific Cloud Armor rule.

### Architecture & Design

Keep the existing Redis transaction and policy, but unpack the INCRBY result after checking both command errors and validating the counter. Whitelist handling, thresholds, window/expiry, timeout, and the existing warn-and-fail-open behavior for Redis failures remain unchanged; no dependencies are added.

### Implementation

- Use ioredis's inferred callback types and resolve the numeric counter instead of `replies[1]`.
- Surface SET/INCRBY errors and incomplete transaction replies through the existing warning path.
- Reject invalid counter values rather than silently comparing them with limits.

### Tests

Added `graphql-api/src/graphql/rate-limiting.spec.ts` with realistic ioredis EXEC replies, mocked Redis/logger/whitelist, and fake timers. Coverage includes exact thresholds, request/cost overages, SET NX returning null for an existing key, whitelist bypass, command/transaction errors, incomplete replies, invalid counters, and timeouts; no Redis server or GCP access is needed.

**Local before/after proof** (test limits: 30 requests/minute and 300 cost units/minute):

| Successful Redis replies | Before | After |
| --- | --- | --- |
| Request counter `[[null, 'OK'], [null, 31]]` | Allowed | Rejects with `UserVisibleError` |
| Request count 1, then cost counter `[[null, 'OK'], [null, 301]]` | Allowed | Rejects with `UserVisibleError` |
| Request count 30 and cost count 300 | Allowed | Allowed |

The bug is JavaScript coercion: `[null, 31] > 30` is false, while `31 > 30` is true. The new suite produced **16 failures / 5 passes before the fix**, including both over-limit cases resolving instead of rejecting; **all 21 tests pass with the fix**.

Also verified the before/after source modules against **real ioredis and a fresh local Redis instance**, isolated on a Unix socket with TCP and persistence disabled. With test configuration/clock/whitelist/logger supplied by the harness, real Redis replies produced the same three outcomes above: both overages allowed before and rejected after, exact limits allowed in both versions, and no warnings. No production service or GCP access was involved.

Run the regression suite:

```bash
pnpm jest --selectProjects graphql-api --runInBand --runTestsByPath graphql-api/src/graphql/rate-limiting.spec.ts
```

None.

This proves the bug in the tested source, not its presence in a particular deployed image. Existing GCP logs do not expose these Redis replies/counter values; HTTP 200 alone cannot prove enforcement failure because caching, whitelisting, and GraphQL error responses complicate interpretation. Verify the deployed revision separately.

Restoring previously ineffective quotas can increase legitimate users' rate-limit errors. Monitor those errors and whitelist needs during rollout. Redis failures still fail open intentionally; changing that availability policy is outside this PR.

### Verification

- [x] `make validate-graphql-api` passed: formatting, ESLint, workspace typecheck, and all 282 API tests across six suites.
- UI desktop/mobile verification: not applicable.
- [x] Repository instructions followed and the required `Assisted-by` trailer is present on the commit.


rileyhgrant (approved):
Yikes. Good catch, thanks for opening this PR, looks like a swap to ioredis from redis + this `any` resulted in this being incorrect for a long time now.

LGTM. Curious to see how this affects our numbers.

PR #1962, opened 2026-09-17
Use the new helper function abbreviateLongReadTableLabel to limit variant ID and SR match column id size

bw2:
A variant ID whose REF or ALT is a long sequence was printed in full, running off the row. Two display paths still did this after `abbreviateLongReadTableLabel` landed, because that helper is only applied to `compactLabel`:

- `formatLongReadVariantId()` returned a bare `withoutChr(sourceId)`. This is the label for the variant page title fallback, breadcrumbs, search suggestions, the Overlapping Variant Calls links, the Enveloping TR link, and the multiple-matches list in `VariantPageRouter`.
- The SR Match column in `HaplotypeVariantTable` printed `v.short_read_match_id` raw. That one is a gnomAD v4 short-read ID, so it never passed through the long-read formatter at all.

Both now call the same helper, so there is one abbreviation convention rather than a competing one:

```
link : 15-90871538-G-GCGGCGGGCGGAC…GGAGGTGGCGCC
link : 15-90871538-GCGGCGGGCGGACGA…AGGTGGCGCC-G
link : 4-3113748-3113861-CAGCAGCAG…CAGCAGCAGCAG
link : 15-90871538-G-A
table: 1-55039879-A-GCGGCGGGCGGACG…GGAGCCGGCGGT   (unchanged)
```

Nothing that is used as an identity changes: link targets, `href`s and search values keep the full ID, and the SR Match cell gains `title={v.short_read_match_id}` so the exact ID is still reachable on hover.

Deliberately left alone: the "Variant ID" column described as "Exact ID from the source VCF record", and the `Canonical browser ID` / `Source VCF record ID` `<code>` rows on the variant page. Those exist to be copied exactly, and abbreviating them would make a text selection silently yield a wrong ID.

Tests: 4 new cases (abbreviated ALT, abbreviated TR motif, the `(Allele N)` marker surviving abbreviation, and the SR Match cell's displayed text plus its full `title` and `href`). Across VariantPage, Haplotypes, VariantList, search and LongReadVariantPage: 894 passed, 26 failed, all pre-existing on the base branch.

Based on `lr-plot-controls-and-catalog-context-20260917`, so it should merge after that branch.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01ARmR9K6tuoD7SHbqKg73so


bw2:
This fixes a very long id we noticed with Xuefang earlier today when looking at the Furin gene page.

mattsolo1 (approved):

